PR #9, opened 2022-10-14
Support clients using CMake

friendlyanon:
This commit introduces proper CMake usage that allows users to simply use "find_package(glaze REQUIRED)" to begin using the library. There are also further changes that make it possible to use the library by vendoring (add_subdirectory, FetchContent, etc).

The Conan package manager was integrated to transparently acquire dependencies.

I did not touch tests beyond what was necessary to make things work using a package manager.

The workflow with these changes quite a bit.  
To begin developing, you might want to document/do something like this: https://github.com/friendlyanon/cmake-init-header-only/blob/master/HACKING.md#developer-mode  
For the user facing documentation, you might want to do something like this: https://github.com/friendlyanon/cmake-init-header-only/blob/master/BUILDING.md

To manage Conan, you have a variety of options and I'm impartial to either of them. You can check the Conan docs or the conan channel on the C++ Slack for more info.

With these changes, it will be trivial to consume the project via CMake:

```cmake
# Installed
find_package(glaze REQUIRED)
# or vendored
add_subdirectory(third-party/glaze)

# Using it is the exact same, no clobbering of names
target_link_libraries(project_target PRIVATE glaze::glaze)
```

Oh, also this doesn't build with MSVC. Something about calling declval.


stephenberry:
Thanks for this, I'll make sure to look through it before pulling, but it seems great.

It should build with MSVC. If you're able to replicate the build issue with the latest version, let me know.

Ahajha:
Regarding consumption of the package, the interface from CMake (find_package/glaze::glaze) will be identical whether consuming from CMake (fetchcontent or CPM) or from Conan, which is nice.

friendlyanon:
No, you need to setup the package manager first for you system:

```sh
# Conan can be installed using pip
pip install conan
# Conan needs a profile setup
conan profile new default --detect
# On Linux, you need to set ABI to be C++11 if the detected compiler is GCC
conan profile update settings.compiler.libcxx=libstdc++11 default
# C++ standard is not detected, set to C++20 (NanoRange requires this)
conan profile update settings.compiler.cppstd=20 default
# Set the build type to debug for developing
conan profile update settings.build_type=Debug default
```

This is a one-time setup and allows you to manage projects' dependencies via Conan. To use it with glaze, do the following:

```sh
# Install debug type dependencies, building those from source for which there are no pre-built binaries in Conan Center Index
conan install . -b missing
# Configure the project with developer mode, using Conan generated preset
cmake --preset=debug -Dglaze_DEVELOPER_MODE=1
# Build in debug mode
cmake --build build
```

The above assumes that you are using Linux, where Conan and CMake default to a single config generator. On Windows, the default is a multi config generator, where the CMake commands are different:

```sh
cmake --preset=default -Dglaze_DEVELOPER_MODE=1
cmake --build build --config Debug
```

friendlyanon:
I have also updated the CI workflow. It seems GCC 10 hits an ICE issue, so I upgraded the workflow to Ubuntu 22.04 to use GCC 11, which works fine. Only the test hits some weird issue.

mwalcott3:
Overall I support this change. The CMake changes are a definite improvement.

I do have mixed feelings about requiring conan for pulling in dependencies if they are not already on the system. However, conan does appear to have considerable and growing userbase and is quickly becoming the standard for dependency management in C++. CPM was originally used since it only depended on CMake. We will need to see how easy it is to handle offline environments since that is a required use case but I don't imagine that will be that painful.

friendlyanon:
> requiring conan

You misunderstand. Conan is *not a requirement*. It integrates with CMake *completely transparently*. `find_package()` and following upstream install interfaces of your dependencies does not lock you into any dependency manager. You can substitue Conan with anything you want to, like vcpkg, the new dependency providers, injecting CPM is some way, or your own creative solution.

PR #25, opened 2022-10-24
CMake find package

stephenberry:
Modified Support clients using CMake #9 to use CPM.


stephenberry:
find_package should now work, originally from Support clients using CMake #9

friendlyanon:
I'd highly recommend restoring the state of the CMake files back to how they were in #9

If you wish to use CPM, then just use the new dependency provider feature in CMake 3.24 to not pollute project code with CPM. Just create this file somewhere with the details filled in:

```cmake
include(some/path/on/your/system/not/tracked/by/git/to/CPM.cmake)

macro(redirect_find_package_to_cpm method package)
  if("${package}" STREQUAL "fmt")
    CPMAddPackage(...)
  elseif("${package}" STREQUAL "FastFloat")
    ...
  elseif(...)
    ...
  endif()
endmacro()

cmake_language(
    SET_DEPENDENCY_PROVIDER redirect_find_package_to_cpm
    SUPPORTED_METHODS FIND_PACKAGE
)
```

and pass the file's location to CMake >=3.24 when configuring: `cmake ... -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=use-cpm.cmake`

This way you can continue to use CPM and the CMake files stay clean.

stephenberry:
CPM should only kick in if find_package fails. The following option is set `option(CPM_USE_LOCAL_PACKAGES "Try `find_package` before downloading dependencies" ON)`.

What is the danger of including CPM? How does it pollute?

Thanks for your feedback.

friendlyanon:
> What is the danger of including CPM? How does it pollute?

It's unnecessary, because the most universal mechanism to import dependencies is `find_package()`. Especially with CMake 3.24, you can trivially substitute it with anything else you want.

Someone who knows how `find_package()` works will immediately know what the project needs to build. Not so with CPM. Personally I don't know what its options are and don't really care either, since I use package managers like Conan and vcpkg.

stephenberry:
@friendlyanon 

I followed your helpful recommendation and am using CMake's dependency provider feature (#26).

Reading the original issue was really helpful for me (https://gitlab.kitware.com/cmake/cmake/-/issues/22619).

My only concern is with other libraries using CPM. Forcing them to use or edit our `use_cpm.cmake` file and not being able to just run their CMake script directly makes their life more difficult. I say this because I have customers who use CPM.

Can you think of any way around requiring this option to be an additional argument when calling CMake? Even if it is something they add to their CMake file?

And, what happens when multiple libraries define their own dependency providers? Can't there only be one? Wouldn't the user have to merge these scripts?

friendlyanon:
> My only concern is with other libraries using CPM.

No need to be concerned. The library is FetchContent-ready because of #9  
For transitive dependencies, they may use their own dependency providers as well, but ultimately, vendoring this way simply does not scale, so you really shouldn't focus on this. The focus should be the normal `find_package()` workflow.

> Can you think of any way around requiring this option to be an additional argument when calling CMake?

Dependency providers very intentionally cannot be defined outside of the script executed by `CMAKE_PROJECT_TOP_LEVEL_INCLUDES`. The intention is for projects to use only `find_package()` and this way allows users to conveniently overwrite the behavior without intrusive modifications to the CMake code, such as using `CPMAddPackage` instead of `find_package`.

> what happens when multiple libraries define their own dependency providers?

That's impossible, since libraries by definition cannot define dependency providers, only the user configuring the project can.

stephenberry:
Thanks for your feedback!

stephenberry:
So, we decided to go back to using CPM, but using CPMFindPackage. It basically just tries to find_package and then call a fetch content if find_package fails. So, this plays well with anyone using find_package.

This also makes it easier for users to just download the package and build it with CMake, no other steps or additional CMake includes are necessary.

For users of CPM it also means they don't have to develop a dependency provider that covers all the dependencies in their project. It is nice to just have dependencies automatically pull in their own dependencies. And this will all work with find_package by default.

Issue #33, opened 2022-10-27
Hash performance on glaze_objects with small number of keys.


mwalcott3:
This is based on conversations I have had outside of github with @stephenberry and Im adding it here to be tracked. We have seen that a naive space inefficient constexpr perfect hash table can get better lookup perf (~2x) than frozen unordered_map when there are a small number of compile-time known keys. Performance disparity should exist for greater numbers of keys but either space or compiletime requirements grow too rapidly for it to be feasible. This should affect object json read perf as Ive seen the map lookup time take anywhere from 5-50% of the runtime depending on the complexity of the object. 

There is another potential optimization but it can cause problems in corner cases. We can get further performance on long keys by changing the key validity test. We already have the correct table key hashes compiletime and the input key hash is calculated each lookup so a hash test is very cheap on large keys and while not perfect is probably good enough. This is opposed to the sting equality test used by default in frozen. Correct key hashes cannot collide due to perfect hashing but invalid keys have a small chance (~1/(2^64)) to collide but are generally unexpected so I think the hash test is good enough but we should probably only support this as a compile time option as it is incorrect. 

I suggest we implement a small naive perfect hash map that follows a subset of the frozen api that can be used when we know key sizes are low (<10 or so) and then switch to frozen (using "if constexpr") when key sizes in glaze objects are large. This should not affect end users since all the constxepr map stuff is internal and part of the detail impl.

PR #44, opened 2022-11-02
Hash map

mwalcott3:


mwalcott3:
This addresses #33. Funny enough the major perf improvement in initial testing I was seeing is still mainly due to the string comparison. frozen_string implements its own equality operator for frozen_string which is notably slower than memcmp (What string comparisons tend to use) on some platforms where that operation is vectorized. Hash checks are way faster if you are okay with the collision chance for invalid keys. Valid keys can not collide due to perfect hashing. I dont expect significant perf improvement on windows but gcc and clang should be faster from my testing.

PR #65, opened 2022-12-01
cmake adjustment to factor in clang-cl MSVC compiler

calr0:
Just added a small CMAKE tweak to avoid adding /Zc:preprocessor when building with the MSVC clang-cl compiler.


stephenberry:
Thanks!

calr0:
sure thing, thanks for the nice json lib!

PR #83, opened 2022-12-15
shrink_to_fit memory optimization

stephenberry:
compile time option to call shrink_to_fit after resize calls


stephenberry:
Addresses #83

Issue #55, opened 2022-11-11
Proper escaped string support


mwalcott3:
Right now only quotes and slashes are handled properly. Other control chars and \u encoded unicode is not handled. 

Refer to spec for more info on what has to be handled https://www.json.org/json-en.html

mwalcott3:
This is a known issue but I'm adding it here to be tracked. There are quite a number of things I want to change before this. So despite it being a major conformance issue its not actively being worked on at the moment.

mwalcott3:
Shouldn't we be unescaping unicode and control chars? #71 appears to just be storing the escaped versions of those chars.

PR #86, opened 2022-12-19
Unescape control and unicode chars when reading. And escape control chars when writing.

mwalcott3:
This deals with issue #55.

Note that the handling of escaped Unicode could be improved quite a bit but I don't expect them that often. Instead, Id expect utf8 encoded strings.


mwalcott3:
Note that non contiguous buffers have no tests and some bugs. However, support for non contiguous buffers is getting removed anyways.

Issue #92, opened 2022-12-26
Any plan to support NDJSON?


toge:
Hello.
I think glaze is a great library.
I have started to use glaze when I write new programs.
Next phase, I would like to replace other libraries with glaze in existing programs where speed is important, but NDJSON(Newline Delimited JSON) support is an issue.
http://ndjson.org/

As far as I know, glaze does not support NDJSON.
Are there any plans to support NDJSON?

stephenberry:
NDJSON is a reasonable and great suggestion. We'll try to add support for this soon, probably as a compile time option.

stephenberry:
@toge

NDJSON support has been added for array types like `std::array`, `std::vector`, `std::tuple`, and `glz::array`

For runtime deduced variant types see this [Variant Handling](https://github.com/stephenberry/glaze/wiki/Variant-Handling) documentation on the Wiki. This will let you deduce the type from the NDJSON.

Here are some tests that demonstrate the behavior and calling syntax:
```c++
      std::vector<std::string> x = { "Hello", "World", "Ice", "Cream" };
      std::string s = glz::write_ndjson(x);
      
      expect(s ==
R"("Hello"
"World"
"Ice"
"Cream")");
      
      x.clear();
      
      glz::read_ndjson(x, s);
      expect(x[0] == "Hello");
      expect(x[1] == "World");
      expect(x[2] == "Ice");
      expect(x[3] == "Cream");
```

Another example:
```c++
      std::tuple<my_struct, sub_thing> x{};
      
      std::string s = glz::write_ndjson(x);
      
      expect(s ==
R"({"i":287,"d":3.14,"hello":"Hello World","arr":[1,2,3]}
{"a":3.14,"b":"stuff"})");
      
      auto& first = std::get<0>(x);
      auto& second = std::get<1>(x);
      
      first.hello.clear();
      first.arr[0] = 0;
      second.a = 0.0;
      second.b.clear();
      
      glz::read_ndjson(x, s);
      
      expect(first.hello == "Hello World");
      expect(first.arr[0] = 1);
      expect(second.a == 3.14);
      expect(second.b == "stuff");
```

toge:
@stephenberry 
Thanks a lot!
It is wonderful.
I will give it a try!

PR #94, opened 2023-01-03
ndjson support

stephenberry:


stephenberry:
#92 - Adding initial NDJSON support.

PR #102, opened 2023-01-07
Deleted obsolete command in install-rules.cmake

klemenpeganc:
The file referenced was removed in https://github.com/stephenberry/glaze/commit/ab561f11ae4c17a5b9d9597a611ffd04ff97a42c.

### DIFF
--- a/cmake/install-rules.cmake
+++ b/cmake/install-rules.cmake
@@ -37,13 +37,6 @@ set(
 mark_as_advanced(glaze_INSTALL_CMAKEDIR)
 
 install(
-    FILES cmake/install-config.cmake
-    DESTINATION "${glaze_INSTALL_CMAKEDIR}"
-    RENAME "${package}Config.cmake"
-    COMPONENT glaze_Development
-)
-
-install(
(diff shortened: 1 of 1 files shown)


klemenpeganc:
Something is broken here. Without this command, `glazeConfig.cmake` file will not be generated, which means projects won't be able to call `find_package(glaze REQUIRED)`. However, having this install command will lead to a File doesn't exist error. I have temporarily fixed it on my fork by adding the file `install-config.cmake` and only having `include("${CMAKE_CURRENT_LIST_DIR}/glazeTargets.cmake")` in it. Please take a look.

mwalcott3:
> Something is broken here. Without this command, `glazeConfig.cmake` file will not be generated, which means projects won't be able to call `find_package(glaze REQUIRED)`. However, having this install command will lead to a File doesn't exist error. I have temporarily fixed it on my fork by adding the file `install-config.cmake` and only having `include("${CMAKE_CURRENT_LIST_DIR}/glazeTargets.cmake")` in it. Please take a look.

@klemenpeganc Thanks for bringing this up. Fixed in #103.

PR #113, opened 2023-01-17
checking if buffer is null terminated

stephenberry:


stephenberry:
Decided that it is better to just require null termination than try to handle both cases, as the null termination is not hard to achieve and check for contiguous buffers and it offers better performance.

PR #156, opened 2023-02-06
No except

stephenberry:


mwalcott3:
@stephenberry shrink_to_fit does nothing on gcc if exceptions are disabled. This breaks the shrink_to_fit test.

```c++
#if __cplusplus >= 201103L
  template<typename _Tp, bool
    = __or_<is_copy_constructible<typename _Tp::value_type>,
            is_nothrow_move_constructible<typename _Tp::value_type>>::value>
    struct __shrink_to_fit_aux
    { static bool _S_do_it(_Tp&) noexcept { return false; } };

  template<typename _Tp>
    struct __shrink_to_fit_aux<_Tp, true>
    {
      _GLIBCXX20_CONSTEXPR
      static bool
      _S_do_it(_Tp& __c) noexcept
      {
#if __cpp_exceptions
	try
	  {
	    _Tp(__make_move_if_noexcept_iterator(__c.begin()),
		__make_move_if_noexcept_iterator(__c.end()),
		__c.get_allocator()).swap(__c);
	    return true;
	  }
	catch(...)
	  { return false; }
#else
	return false;
#endif
      }
    };
#endif
```
https://github.com/gcc-mirror/gcc/blob/4341106354c6a463ce3628a4ef9c1a1d37193b59/libstdc%2B%2B-v3/include/bits/alloc_traits.h#LL888-L917C7

shrink_to_fit makes no guarantees. So we cant really test it.
> It is a non-binding request to reduce [capacity()](https://en.cppreference.com/w/cpp/container/vector/capacity) to [size()](https://en.cppreference.com/w/cpp/container/vector/size).

stephenberry:
> @stephenberry shrink_to_fit does nothing on gcc if exceptions are disabled. This breaks the shrink_to_fit test.
> 
> shrink_to_fit makes no guarantees. So we cant really test it. Should we just remove the test or is there another way to make sure its working?
> 
> [shrink_to_fit reference ](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit):
> 
> > It is a non-binding request to reduce [capacity()](https://en.cppreference.com/w/cpp/container/vector/capacity) to [size()](https://en.cppreference.com/w/cpp/container/vector/size).
> 
> [gcc implementation](https://github.com/gcc-mirror/gcc/blob/4341106354c6a463ce3628a4ef9c1a1d37193b59/libstdc%2B%2B-v3/include/bits/alloc_traits.h#LL888-L917C7):
> 
> ```c++
> #if __cplusplus >= 201103L
>   template<typename _Tp, bool
>     = __or_<is_copy_constructible<typename _Tp::value_type>,
>             is_nothrow_move_constructible<typename _Tp::value_type>>::value>
>     struct __shrink_to_fit_aux
>     { static bool _S_do_it(_Tp&) noexcept { return false; } };
> 
>   template<typename _Tp>
>     struct __shrink_to_fit_aux<_Tp, true>
>     {
>       _GLIBCXX20_CONSTEXPR
>       static bool
>       _S_do_it(_Tp& __c) noexcept
>       {
> #if __cpp_exceptions
> 	try
> 	  {
> 	    _Tp(__make_move_if_noexcept_iterator(__c.begin()),
> 		__make_move_if_noexcept_iterator(__c.end()),
> 		__c.get_allocator()).swap(__c);
> 	    return true;
> 	  }
> 	catch(...)
> 	  { return false; }
> #else
> 	return false;
> #endif
>       }
>     };
> #endif
> ```

We could probably just keep the test and only run it if __cpp_exceptions is enabled

Issue #165, opened 2023-02-14
Nested objects in glz::json_t result in 'std::length_error` exception


klemenpeganc:
Hey!
So I am trying to parse a schema into the `glz::json_t` object. This will be better explained in code:
```
static const glz::json_t messageSchema = {
    {"type","struct"},
    {"fields", {
        {{"field","branch"}, {"type","string"}},
        }
    }
};
```
This results in:
```
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_S_create
```
If I try to do it like this (without extra brackets around `{"field","branch"}, {"type","string"}`) it works, but not what I want:
```
static const glz::json_t messageSchema = {
    {"type","struct"},
    {"fields", {
        {"field","branch"}, {"type","string"},
        }
    }
};
```

Does the `glz::json_t` have issues with nesting and if you have any suggestions on how to fix it? That would be much appreciated!

EDIT:
```
    glz::json_t messageSchema;
    glz::json_t field1 = {{"field", "branch"}, {"type", "string"}};
    messageSchema["type"]  = "struct";
    messageSchema["fields"] = {field1};
```
This does work though!

mwalcott3:
Added a nested initialization test for json_t based on your code and fixed the issue in #166.

> The problem is that the initializer list for constructing the object type was std::initializer_list<std::pair<const std::string, json_t>>
> 
> And std::string would match pairs of string literals like {"literal", "other_literal"} and treat them as the start and end of the string.
> 
> The workaround was std::initializer_list<std::pair<const char *, json_t>> since {"literal", "other_literal"} is not a valid initializer for const char * but unfortunately, this requires data to be copied over since you are not supposed to move elements out of an initializer list.

PR #166, opened 2023-02-14
Nested initialization test and fix

mwalcott3:
Addresses and adds a test for the behavior noted in #165


mwalcott3:
The problem is that the initializer list for constructing the object type was std::initializer_list<std::pair<const std::string, json_t>>

And std::string would match pairs of string literals like {"literal", "other_literal"} and treat them as the start and end of the string.

The workaround was std::initializer_list<std::pair<const char *, json_t>> since {"literal", "other_literal"} is not a valid constructor for const char * but unfortunately, this requires data to be copied over since you are not supposed to move elements out of an initializer list.

I might talk with @stephenberry about this later since Id like to avoid unnecessary copies if possible to see if there is another way to fix the initialization problem.

PR #167, opened 2023-02-15
Improve json_t perf and remove lib specific support

mwalcott3:
Improve json_t performance by removing unnecessary unwrapping and using an unordered map. Remove json_t specific code in the reader so it could be implemented by a user and doesn't require direct support in the library.

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -11,6 +11,7 @@ on:
     branches: [main]
     paths-ignore:
     - '**.md'
+  workflow_dispatch:
 
 env:
   BUILD_TYPE: Release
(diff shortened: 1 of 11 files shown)


mwalcott3:
Cannot use unordered_map due to the inability to do recursion without additional indirection like unique_ptr due to it not allowing incomplete types. As of gcc 12 the 3 major compilers allow you to do such but it is a bit iffy for anyone not using those compilers and gcc 12 is too recent to reliably have on most linux distros.

The main perf problem with json_t right now is not the map anyways its the allocator due to the heavy heap allocations. Supporting pmr would make more sense as an optimization.

PR #180, opened 2023-03-02
Fix includes for vendoring projects

friendlyanon:
Vendoring projects would previously install glaze directly to `/usr/local/include` if `/usr/local` is their install prefix and wish to install glaze themselves.

Example where this could go wrong:

    set(CMAKE_SKIP_INSTALL_RULES OFF)
    add_subdirectory(third-party/glaze)

And when installing:

    $ cmake -B build -DCMAKE_INSTALL_PREFIX=/foo
    ...
    $ cmake --build build -t install
    ...
    -- Installing: /foo/include/glaze.hpp
    ...

This breaks `#include <glaze/glaze.hpp>` includes!

### DIFF
--- a/cmake/install-rules.cmake
+++ b/cmake/install-rules.cmake
@@ -12,7 +12,7 @@ include(GNUInstallDirs)
 set(package glaze)
 
 install(
-    DIRECTORY include/glaze/
+    DIRECTORY include/
     DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
     COMPONENT glaze_Development
 )


friendlyanon:
Accidentally wrote `CMAKE_INSTALL_PATH` instead of `CMAKE_INSTALL_PREFIX` in the commit. Could you please fix that when squash merging? I don't want to rerun CI just for that :)

friendlyanon:
Oh no! The commit wasn't squashed! Now my shameful lazyness will forever be part of the project's git history 😔

mwalcott3:
My bad I was a little to trigger happy merging that.

friendlyanon:
Hopefully someone looking at that commit in the future will find their way back here and discover that the example code is known to use a made up variable by mistake.

Issue #177, opened 2023-03-02
Silence the compiler for unused variables


sweihub:
Howdy

Glaze for serializing/deserializing JSON is really awsome, would you please silence the compiler by removing the unused variables, thanks.

enable the compiler options `-Wunused-variable -Wunused-but-set-variable`

```
/home/swei/quant/trade/core/include/glaze/util/strod.hpp:404:22: warning: variable 'sig_end' set but not used [-Wunused-but-set-variable]
      const uint8_t *sig_end = nullptr;  /* significant part ending position */
...
/home/swei/quant/trade/core/include/glaze/util/dtoa.hpp:981:12: warning: unused variable 'start' [-Wunused-variable]
      auto start = buffer;
           ^
...

```

mwalcott3:
@stephenberry and I were just talking about this. I thought he removed those today already.

Beyond this we should probably use -Wall -Wextra and -Werror in the unit tests so the ci builds will fail to make sure we are avoiding the common warnings.

mwalcott3:
Addressed in #182

Glaze should no longer emit warnings with
-Wall -Wextra -pedantic
This is now enforced in the github workflow with -Werror for the unit tests so hopefully we shouldnt pollute other peoples builds with common warnings in the future.

MSVC is using /W4. There is still one warning left but that is an actual bug that needs to be fixed. Ill add /WX to the tests when that gets addressed.

PR #182, opened 2023-03-02
Add warnings flags to test targets and enforce in github workflow with Werror

mwalcott3:
Addresses #177

Try to handle a bunch of warnings and start enforcing lack of warnings in the tests

Adds the following flags to the test targets
 -Wall -Wextra -pedantic $<$<CONFIG:Debug>:-Werror> on gcc/clang
/W4 on windows

### DIFF
--- a/.github/workflows/clang.yml
+++ b/.github/workflows/clang.yml
@@ -14,7 +14,7 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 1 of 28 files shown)


mwalcott3:
I am not adding Werror to the tests till I get rid of all the gcc/clang warnings

PR #190, opened 2023-03-06
Make it able to compile with gcc-10

fregate:
I used gcc-10 for embedded compilation and saw a few errors.

Using with enums not supported for this version - this is easy part.

But in `include/glaze/json/json_ptr.hpp` gcc has crashed in these line. So I just split it to 2 constexpr and now it compiles.


mwalcott3 (approved):
Looks good, thank you for sharing these changes. Ill be sure to add in gcc-10 testing to the workflows.

PR #201, opened 2023-03-07
fix: string overflow in get_source_info

pwqbot:
when error position point to '\n', `pnl == nnl`, then
```cpp
 std::string context{
               std::begin(buffer) +
                  (pnl == std::rend(buffer) ? 0 : index - dist + 1),
               nnl};
```
the first arg will be `nnl + 1`, the constructor will throw `basic_string::_M_create`.

### DIFF
--- a/include/glaze/util/validate.hpp
+++ b/include/glaze/util/validate.hpp
@@ -3,6 +3,7 @@
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


mwalcott3:
Thanks for fixing this and adding a test case as well.

Issue #208, opened 2023-03-12
Have these codes been formatted correctly? 


pwqbot:
Have these codes been formatted correctly? Some of the code exceeds the column limit of 120, and when I save the code using clang-format, it will be reformatted.

mwalcott3:
Addressed in #209. clang-format was run on the entire project and @pwqbot added a github action to enforce that future additions are formatted correctly to help prevent this issue from popping up in the future.

PR #209, opened 2023-03-12
add clang-format test action

pwqbot:
issue #208


mwalcott3:
Thanks, I think checking the formatting in the action is generally a good idea. It can be harder to review changes when a bunch of code changes are just the formatter changing lines of code.

PR #210, opened 2023-03-13
add string_view support

SlaaneshChosen:


SlaaneshChosen:
`string` and `string_view` are two different things when parsing a json. For a `string`, you copy it from the orginal json string and handle escaped. Sometimes you may just want to have a `view` of a field and will never change it, and `string_view` is a better choince.

include/glaze/json/read.hpp
```diff
@@ -394,6 +394,92 @@ namespace glz
...
+               switch (*it) {
+               case '"':
+               case '\\':
+               case '/':
+               case 'b':
+               case 'f':
+               case 'n':
+               case 'r':
+               case 't':
+               case 'u': {
```
mwalcott3, line 426: Should probably validate \u is followed by 4 hex digits when force_conformance == true. But I can make that change later.

tests/json_test/json_test.cpp
```diff
@@ -64,6 +64,35 @@ suite starter = [] {
    };
 };
 
+//struct my_strings
```
mwalcott3: Why is this test commented out?

mwalcott3 (approved):
Looks good. I should be talking to @stephenberry soon so Ill probably wait for his input before we merge this.

PR #217, opened 2023-03-16
fix quoted compile error caused by error_on_missing_keys

pwqbot:
``` cpp
/Users/benben/dev/github/glaze/include/glaze/core/common.hpp:779:107: error: reference member of type 'double &' uninitialized
               std::decay_t<member_t<T, std::tuple_element_t<1, std::tuple_element_t<I, meta_t<T>>>>>...>{};
                                                                                                          ^
/Users/benben/dev/github/glaze/include/glaze/core/common.hpp:789:32: note: in instantiation of function template specialization 'glz::detail::members_from_meta_impl<A, 0UL>' requested he
re
         using type = decltype(members_from_meta_impl<T, I...>());
                               ^
/Users/benben/dev/github/glaze/include/glaze/core/common.hpp:793:7: note: in instantiation of template class 'glz::detail::members_from_meta<A, std::integer_sequence<unsigned long, 0>>'
requested here
      using member_tuple_t = typename members_from_meta<T>::type;
      ^
/Users/benben/dev/github/glaze/include/glaze/core/common.hpp:819:77: note: in instantiation of template type alias 'member_tuple_t' requested here
                           !nullable_t<std::decay_t<std::tuple_element_t<I, member_tuple_t<T>>>>;
                                                                            ^
/Users/benben/dev/github/glaze/include/glaze/json/read.hpp:1150:50: note: in instantiation of function template specialization 'glz::detail::required_fields<A, {10, false, true, true, fa
lse, false, 32, 3, false, true, true, false, true, false, false}>' requested here
                     constexpr auto req_fields = required_fields<T, Opts>();
                                                 ^
```
quoted holds reference in it and cannot be holds by tuple.

### DIFF
--- a/include/glaze/core/common.hpp
+++ b/include/glaze/core/common.hpp
@@ -815,8 +815,9 @@ namespace glz
          bit_array<n> fields{};
          if constexpr (Opts.error_on_missing_keys) {
             for_each<n>([&](auto I) constexpr {
-               fields[I] = Opts.skip_null_members == false ||
-                           !nullable_t<std::decay_t<std::tuple_element_t<I, member_tuple_t<T>>>>;
+               fields[I] =
+                  !static_cast<bool>(Opts.skip_null_members) ||
+                  !nullable_t<std::decay_t<member_t<T, std::tuple_element_t<1, std::tuple_element_t<I, meta_t<T>>>>>>;
             });
          }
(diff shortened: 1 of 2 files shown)


mwalcott3:
Thanks, member_t is cleaner as well.

PR #220, opened 2023-03-18
jsonrpc change client callback mechanism

jbbjarnason:
On second thought, the jsonrpc client API with long living callback is not great.
Instead I changed so the user can provide one callback per request then that callback gets removed after it is called.
This will make it easier to integrate with async loops.

I know this is breaking changes but it should not impact much giving how new this is.

There is a diff of the server call method, I changed one lambda from being brace initialized to normal assignment, since clang-format version 15 is incompetent in formatting such cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1082,6 +1082,7 @@ auto main(int, char**) -> int {
                 rpc::client_method_t<"bar", method_bar_params, method_bar_result>>
        client;
    
+    // One long living callback per method for the server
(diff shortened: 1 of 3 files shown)


mwalcott3:
Handling a single callback per request does seem easier to work with.

Better to make breaking changes earlier than later. None of this stuff is on a tagged version right now so is it even a breaking change?

jbbjarnason:
Only breaking from master so yes not really breaking

Issue #216, opened 2023-03-15
Enums should not require metas unless we want to override behavior


mwalcott3:
There is a way to do static reflection with enums that someone found https://blog.rink.nu/2023/02/12/behind-the-magic-of-magic_enum. It is nonstandard but supported on the major compilers. I don't like relying on nonstd stuff but propper reflection is probably 6 years away so we should probably take what we can get.

https://github.com/Neargye/magic_enum

I don't yet know the limitations or how suitable it is but it looks promising. Ill look into this over the weekend.

mwalcott3:
On a somewhat related/unrelated note magic_get style reflection could be used in binary serialization if use_cx_tags = false or we broke the format when we get full static reflection #213

__builtin_dump_struct could technically be used on clang to get rid of the meta for objects but that seems like a very bad idea. https://www.reddit.com/r/cpp/comments/107lrvw/c20clang15experiment_static_reflection_via/

mwalcott3:
Wow, this is likely way more expensive compile-time than I would have thought, each possible value has to be tested so large enum types are not really feasible. This implementation is about as exhaustive as you can reasonably get https://github.com/willwray/enum_traits

This may be a very bad idea. I'll have to do some build-time benchmarking to see how reasonable this is.

mwalcott3:
I have some experimental code enum reflections I'm playing around with and doing some build time profiling. I agree with the limitations that magic_tuple imposes as this really needs to be limited to not tank build times. the approach used by[willwray/enum_traits](https://github.com/willwray/enum_traits) is faster builtime wise but checks way too many values. I'm thinking we can adopt an approach based on what's used in willwray/enum_traits with the following limitations.

- Will only work on modern versions of Clang/gcc/MSVC.
- Only Enums with small values will work.
   - Im thinking of expanding magic_enums limits but not testing 2^16 values like enum_traits
   - I'm thinking of testing 1024 values (0 to1024 for unsigned) (-512 to 511 for signed)
   - May want to limit this further initaly as we can always raise it but lowering it is more likely to break code
- Aliases wont work
- Enums in templates will likely cause problems

I don't think supporting this in such a limited fashion is bad because people can always just generate the meta themselves.

mwalcott3:
Build time is somewhat reasonable. I'm somewhat happy with this implementation https://godbolt.org/z/a1944Gsq4.

mwalcott3:
Clang is utterly broken if the enums are inside templates. But you can use the following workaround
[...]

PR #221, opened 2023-03-18
magic_enum style enum reflection and automatic meta for enums

mwalcott3:
I want to add some nonstandard magic_enum style enum reflection. Results are consistent across the major compilers, and other people can just explicitly define metas. Some other info is on the related issue #216

This uses a similar batching approach as [willwray/enum_traits](https://github.com/willwray/enum_traits) to improve build time costs over [magic_enum](https://github.com/Neargye/magic_enum).

Limitations:
   * Nonstandard and likely only works on tested compilers
      * Tested on gcc10-12, clang13-15, and msvc2019-2022
      * Might work on some others depending on what their std::source_location::current().function_name() looks like
   * Only tests 2048 values in the enum
      * Can be expanded for a build time cost, but will never be fully exhaustive. 2^16 is possible on clang and gcc but can add 1s per enum per translation unit to the build times which is rather heavy.
   * May be problematic with enums in templates
      * On clang If an enum is in a template one of the values must be used before the reflection code is run `static_assert(Color::Red == Color::Red)`
      * Only gcc11 and up handles this
 
This is in response to local enums not being possible, as pointed out by #214. This makes it virtually impossible to write a meta for certain enums since you cannot partially specialize on nested templates (You can still use concepts though). However, I doubt anyone is actually doing that in the real world, so it is probably a non issue.

### DIFF
--- a/include/glaze/reflect/enum.hpp
+++ b/include/glaze/reflect/enum.hpp
@@ -216,11 +216,14 @@ namespace glz::detail
    }
 }
 
-template <class E>
-requires std::is_enum_v<E>
-struct glz::meta<E>
+namespace glz
 {
-   static constexpr std::string_view name =
-      glz::detail::type_name<E>();  // Should be consistant across gcc/MSVC/clang unless enum is in a template
-   static constexpr auto value = glz::detail::Enum{glz::detail::auto_enum<E>()};
-};
+   template <class E>
+      requires std::is_enum_v<E>
+   struct meta<E>
+   {
+      static constexpr std::string_view name =
+         glz::detail::type_name<E>();  // Should be consistant across gcc/MSVC/clang unless enum is in a template
+      static constexpr auto value = glz::detail::Enum{glz::detail::auto_enum<E>()};
+   };
+}
--- a/include/glaze/reflect/enum.hpp
+++ b/include/glaze/reflect/enum.hpp
@@ -214,14 +214,13 @@ namespace glz::detail
       constexpr auto vals = enum_array<T>();
       return auto_enum<T, vals>(std::make_index_sequence<vals.size()>{});
    }
-   template <class T>
-   concept is_enum = std::is_enum_v<T>;
 }
 
-template <glz::detail::is_enum E>
+template <class E>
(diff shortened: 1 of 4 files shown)


mwalcott3:
Bug on msvc when used with enums with no values (Not sure why anyone would do this) but ill fix that later and add a test case. Can get rid of it by brace initializing res in concat_arrays.

include/glaze/reflect/enum.hpp
```diff
@@ -0,0 +1,229 @@
...
+            if (!valid) it += 5;
+#endif
+            it += offset;
+            // bool valid =*it > '9' || *it < '0';
+            if (valid) {
+               valid_enums[count++] = T(i + Start);
+            }
+         };
+         validate_and_consume_prefix(0);
+         for (size_t i = 1; i < n; ++i) {
```
pwqbot, line 145: Here we can return early, because most enumeration values are concentrated near the initial value and reduce build time.

mwalcott3:
> I think it is easy to cause bugs if only testing 2048 values without compile errors if exceed. The speed is also a problem but won't be a bug, maybe better tell users to use int8_t if they want speed in doc and use std::numeric_limits as loop range.
This is the strongest argument for not merging this in or restricting it to underlying types that are two bytes or less. 

I agree that this could easily cause unintended errors. There isn't really a good way of testing if limits are exceeded without doing an exhaustive search because we can't even assume continuous enums and only have a mechanism to check if a number is associated with a name in the enum. This is simply impossible with the default int-based enums since you will 100% hit stepsize limits and most likely other limits (We can't possibly check billions of values). Only 8 and 16 bit underlying_tyeps can reasonably be exhaustively checked. It's sad because most enums are likely 32 bit enums with a limited range but there is no way of verifying that.

While hacks like this are cool/interesting it might be best to wait for proper static reflection in C++ even if that is 6+ years away.

mwalcott3:
I have talked this over with @stephenberry and we have decided against merging this in. Too easy to shoot yourself in the foot with enums using the default underlying type and its not worth having if it only applies to uint8_t and int8_t based enums.

PR #222, opened 2023-03-20
fix error handling

pwqbot:
Theoretically should check error after every `op` call, but maybe a performance impact.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -1186,6 +1186,9 @@ namespace glz
                            read<json>::op<ws_handled<Opts>()>(get_member(value, member_ptr), ctx, it, end);
                         },
                         member_it->second);
+                     if (static_cast<bool>(ctx.error)) [[unlikely]] {
+                        return;
+                     }
                   }
                   else [[unlikely]] {
                      if constexpr (Opts.error_on_unknown_keys) {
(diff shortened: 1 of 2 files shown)


mwalcott3:
Thanks, merged.

My theory is that doing all the error checks immediately could actually speed things up. I think Stephen is skeptical. The ideal assembly is to immediately jump to the error handling on error and to never do the error checks that simply propagate it up upward. Provided most calls get inlined the compiler can get rid of the intermediate error checks. And I have verified this with toy problems on godbolt (gcc and clang got rid of the extra error checks when functions where inclined and MSVC produced garbage with error code based error handling). Of course if stuff is actually using function calls the intermediate checks have to be performed and we cant inline the whole json parser.

The thing is the compiler has to do what your code tells it to do so if we don't simply propagate the errors upwards immediately it might end up doing do the intermediate checks.

Ill probably end up testing checking after every op call this Thusday or Friday and as long as id doesn't cause a significant performance loss Ill do it because not having errors reported in the right place or having the wrong errors reported is annoying.

mwalcott3:
> I'm worry about error check impact on right input, because in most situation if you correctly write field mapping, it's rare to get error, where exception may have a performance advantage

Yes, exceptions are indeed faster with correct input. There was close to a 25% performance reduction moving away from exceptions in glaze that was mostly mitigated by getting the compiler to inline more stuff (Of course this leads to increased binary size). I am way more of a fan of exceptions than error codes in most circumstances but there was a hard requirement that it work without exceptions.

mwalcott3:
Here is an example of how skipping some error checks when everything is being inlined can actually result in worse performance:
https://quick-bench.com/q/xC0IFCkpgEQ1avxNUTQQ0HUHnIc

If everything is being inlined intermediate checks should not even be present in the assembly and skipping checks potentially results in more labels, branching, and worse code generation.

If you remove [[gnu::always_inline]] you will note better performance in the version that skips some error checks.

PR #226, opened 2023-03-25
Add intermediate error checks to read for better error messages.

mwalcott3:
This should prevent the iterator advancing and the error code from changing.
- Should get optimized away if inlined. https://godbolt.org/z/GdToboW9a
- This based off of the discussion in #222
- I just did it for the read_json stuff
- I see a ~8% read performance improvement on the benchmark using clang on my machine but that is likely just more stuff getting inlined. 
   - gcc performance appears to be unchanged. I expect MSVC to be slower with these changes based on how it handles intermediate checks in godbolt and lack of inlineing.
- I might have missed some error checks since we dont return an error code using [[nodiscard]]
   - That might make it harder to shoot yourself in the foot with error checking
   - Right now its pretty much checking after everything that takes the context as an argument
   - Any lamdas that capture the context should be checked as well
- I got rid of some of the error checks at the top since they should no longer be needed, that might break some code if people are writing custom parsers but all those methods are in the detail namespace so I think its probably fine

I miss exceptions.

### DIFF
--- a/include/glaze/json/json_ptr.hpp
+++ b/include/glaze/json/json_ptr.hpp
@@ -641,8 +641,13 @@ namespace glz
                while (true) {
                   skip_ws<Opts>(ctx, it, end);
                   const auto k = parse_key(ctx, it, end);
-                  if (k) {
-                     if (cx_string_cmp<key>(*k)) {
+                  if (bool(ctx.error)) [[unlikely]] {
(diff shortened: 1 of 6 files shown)


mwalcott3:
Should probably check that this doesnt cause performance problems in gcc. I expect MSVC to be slower.

PR #234, opened 2023-03-27
Test auto lambda in api

mwalcott3:
Not sure why this is broken on msvc2022


mwalcott3:
Seeing as its only broken in MSVC 2022 and the error isnt really supper helpful I dont think I'm going to look into this until it actually causes a problem.

PR #237, opened 2023-03-29
wrap unexpected to prevent compile errors with clang 16 & libc++ 16

jbbjarnason:
template deduction is not supported for alias from the cpp standard

the error: 

```bash
error: alias template 'unexpected' requires template arguments; argument deduction only allowed for class templates
                  return unexpected(error_code::unknown_key);
                         ^
```

### DIFF
--- a/include/glaze/util/expected.hpp
+++ b/include/glaze/util/expected.hpp
@@ -59,7 +59,11 @@ namespace glz
    using expected = std::expected<expected_t, unexpected_t>;
 
    template <class unexpected_t>
-   using unexpected = std::unexpected<unexpected_t>;
+   struct unexpected : public std::unexpected<unexpected_t> {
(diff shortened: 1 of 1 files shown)


mwalcott3:
Class template argument deduction for alias templates is part of the C++20 standard [P1814R0](https://wg21.link/P1814R0) but is only implemented in MSVC 19.27, GCC 10, eccp 6.5, nvcc 12.0 
 and up https://en.cppreference.com/w/cpp/compiler_support.

jbbjarnason:
damn it clang, always a bit behind.

jbbjarnason:
this is supported in clang 19

stephenberry:
I refactored this code recently for aliasing std::expected and it works for GCC, just check out the top of Glaze's expected header.

PR #247, opened 2023-04-08
Replace frozen map

mwalcott3:
Get rid of a bunch of code we don't need. Re-enable unsafe buffer optimizations and hash checks to look at perf impact. 

Those two re-enabled appear to be around a 10% boost to read speed in our internal benchmark. But as they are both problematic I decided to re-disable them.


mwalcott3:
Should probably make all reductions a power of 2 since that becomes a single bitwise and instead of the following
% 6
```
        mov     edx, 2863311531
        mov     rcx, rax
        imul    rax, rdx
        shr     rax, 34
        lea     edx, [rax+rax*2]
        mov     eax, ecx
        add     edx, edx
        sub     eax, edx
```
% 8
```
and     eax, 7
```

mwalcott3:
Ill be surprised if someone doesn't have a problem with this but I got the go ahead to add the unsafe opts back in.

mwalcott3:
We should probably improve the quality of the bitmix function. Could result in projects failing to compile or increased build times.

Issue #258, opened 2023-04-19
Cannot support the same key


hulingyue:
{
    "arg": {
        "channel": "tickers",
        "instId": "LTC-USD-200327"
    },
    "data": [{
        "instType": "SWAP",
        "instId": "LTC-USD-200327",
        "last": "9999.99",
        "lastSz": "0.1",
        "askPx": "9999.99",
        "askSz": "11",
        "bidPx": "8888.88",
        "bidSz": "5",
        "open24h": "9000",
        "high24h": "10000",
        "low24h": "8888.88",
        "volCcy24h": "2222",
        "vol24h": "2222",
        "sodUtc0": "2222",
        "sodUtc8": "2222",
        "ts": "1597026383085"
    }]
}

It's a response from okex. I failed to parsing the json string , due to 'arg' and 'data' both has a key 'instId'

stephenberry:
This shouldn't be a problem, because the key is used in different contexts. Could you share some of your glaze code? It is most likely a registration problem.

hulingyue:
> This shouldn't be a problem, because the key is used in different contexts. Could you share some of your glaze code? It is most likely a registration problem.

Of cause. 
I found the error: note: non-constexpr function 'glaze_error' cannot be used in a constant expression
            glaze_error("Unable to find perfect hash");

so i think maybe i can expand the bucket. So i change the function: `template <size_t N>
      constexpr auto naive_bucket_size() noexcept
      {
         return N < 8 ? 3 * N : 6 * N;
      }` and then compilation passed


Those are some of my code, hope those can help you recurrence issues

hulingyue:
GLZ_META(OkexMarketJsonStruct::OkexMarketCommonRequestParameter, channel, instId);
GLZ_META(OkexMarketJsonStruct::OkexMarketEventSuccessArg, channel, instId);
GLZ_META(OkexMarketJsonStruct::OkexMarketDepthResponseDataParameter, asks, bids, ts, checksum);
GLZ_META(OkexMarketJsonStruct::OkexMarketBboResponseDataParameter, instType, instId, last, lastSz, askPx, askSz, bidPx, bidSz, open24h, high24h, low24h, volCcy24h, vol24h, sodUtc0, sodUtc8, ts);

template <typename A>
struct glz::meta<OkexMarketJsonStruct::OkexMarketCommand <A>> {
    using T = OkexMarketJsonStruct::OkexMarketCommand <A>;
    static constexpr auto value = object(
[...]

PR #259, opened 2023-04-20
Just use the Murmur bitmix function.

mwalcott3:
We have no idea what the statistical properties are of the current bitmixer so the chance of failing to find a perfect hash function may be way worse than anticipated. Partially in response to #258 but they likely dont have a problem on the latest version.

### DIFF
--- a/include/glaze/util/hash_map.hpp
+++ b/include/glaze/util/hash_map.hpp
@@ -100,14 +100,14 @@ namespace glz::detail
    // This is one such terible hashing alg
    struct naive_hash
    {
-      // TODO Use better constants calculated by something like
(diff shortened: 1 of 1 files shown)


stephenberry:
I think the problem was the bucket size. Wouldn't this hash lose a tiny fraction of performance?

stephenberry:
Ah, just read your comments on the issue. I think ideally we would use the stupid hash and then fall back on the murmor hash if the stupid one can't find a solution. But, the performance cost of just using murmor is probably minuscule.

stephenberry:
After thinking more, I like this change and so I'm merging it.

PR #285, opened 2023-05-04
CMake Upgrades

justend29:
While writing a vcpkg port for glaze (which has since been added by others), I noticed a few areas where glaze's CMake build-system could be enhanced.

## Changes

1. The option `glaze_DEVELOPER_MODE` was introduced by not used.
    Toggling developer mode was always based upon the `PROJECT_IS_TOP_LEVEL` variable. This caused the *glaze_ide* binary to be built in any situation where the project was built independently, for example, when building the vcpkg port. The vcpkg port introduced by others still builds this unnecessary target.
    
2. The explicitly introduced cache variable `CMAKE_INSTALL_INCLUDEDIR` doubly nested glaze's installed headers.
    When installing glaze, the headers would be placed in `.../include/glaze/glaze/` instead of `.../include/glaze/`. I'm not sure if this nesting was intentional, but assuming it was not, CMake's *GNUInstallDirs* module already introduces this cache variable.
    
 3. Tests for installing & finding glaze were registered with CTest.
     Now, glaze's installation and the finding & usage of *that* installed glaze instance are tested as part of glaze's automated tests.

## Future Work

Should these changes be incorporated into glaze, I would like to edit the upstream vcpkg port & add basic documentation into glaze for using it through vcpkg.

## Questions

Is my understanding correct that Eigen is not treated as a dependency, even though it's part of `ext/eigen.hpp`?
Would the introduction of the CMake component *ext*, to provide the target `glaze::ext` with an interface dependency on *Eigen3* be of interest? This would involve adding a call to `find_dependency(Eigen3)` in `install-config.cmake` that's conditional upon a user's call to `find_package(glaze REQUIRED COMPONENTS ext)`.

### DIFF
--- a/.github/workflows/gcc.yml
+++ b/.github/workflows/gcc.yml
@@ -43,10 +43,3 @@ jobs:
     - name: Test
       working-directory: build
       run: ctest -j 2 --output-on-failure
-      
-    - name: Test Install/find_package
-      working-directory: build
-      run: |
-        sudo cmake --build . --target install
-        cmake -S ${{github.workspace}}/tests/find_package -B ${{github.workspace}}/tests/find_package/build -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
-        cmake --build ${{github.workspace}}/tests/find_package/build -j 2
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *build/
 *bin/
+.idea/
 .vscode/
 .cache/
 conan/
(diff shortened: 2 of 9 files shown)


mwalcott3:
I have not got the faintest clue when it comes to cmake besides the very basics and I don't think @stephenberry does either so I appreciate you making these improvements.

> Is my understanding correct that Eigen is not treated as a dependency, even though it's part of ext/eigen.hpp?
Would the introduction of the CMake component ext, to provide the target glaze::ext with an interface dependency on Eigen3 be of interest? 

Yeah, it was just exposing some eigen helpers that people could use if they were already using eigen. It's not useful without eigen so It was just kind of assumed if you were using it you already pulled in eigen yourself. However, adding it as a component like you stated sounds like a smarter approach.

PR #289, opened 2023-05-08
add support for modern C++ modules

katazina0:
This pull request fixes compilation on clang++ using `-stdlib=libc++` and `-fmodules` flags. The main `glaze/glaze.hpp` can now be used in a `.modulemap` file and it's also possible for it to be compiled into a PCM.

The changes are tiny, however, I'm not sure if `static` should be removed completely, or just replaced with an `extern` specifier.

I have not tested libstdc++, but it does seem to be working perfectly using libc++ with CMake.

### DIFF
--- a/include/glaze/api/name.hpp
+++ b/include/glaze/api/name.hpp
@@ -58,7 +58,7 @@ namespace glz
 #else
       template <const std::string_view&... Strs>
 #endif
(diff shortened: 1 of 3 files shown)


stephenberry:
Thanks for these changes and for testing module support.

PR #300, opened 2023-05-24
support parsing empty sets

justend29:
Previously, glaze would emit a syntax error when parsing an empty JSON array (`[]`) to a set.

This PR adds a check against the parsed string being empty, like the parser for `std::vector`, and adds assertions in the unit tests to validate this for both `std::unordered_set` and `std::set`.


mwalcott3:
Thank you for the bugfix and tests are always appreciated.

PR #301, opened 2023-05-26
Fix prettify of tagged variants

justend29:
Also changed order of read & write tests since a read function is now used in the write test.


stephenberry:
Thanks for this contribution!

PR #329, opened 2023-06-29
adding glz::invoke wrapper

stephenberry:


jbbjarnason:
This is a nice feature, I made a similar one, the difference being the value is owned by a wrapper which has an invocable function for changes and is called when json is read. If someone is interested, https://github.com/Skaginn3x/framework/blob/main/libs/confman/inc/public/tfc/confman/observable.hpp

Issue #335, opened 2023-07-03
Unset char as key cut off the json output


maor-da:
The "pretty" JSON is cut off when using `char`.

```c++
struct test{
  int a;
  char c;
  std::string str;
}
```
```c++
test obj{};
std::string buffer{};
glz::write<glz::opts{.prettify = true}>(obj, buffer);

printf("%s", buffer.c_str());
```

result:
```json
{
   "a": 0,
   "c": "
```

when expecting:
```json
{
   "a": 0,
   "c": "", // since json has only string
   "str": ""
}
```

Another issue is with `char*`, it causes to dereference null in the code.
I can PR if you want.

stephenberry:
Thanks for bringing this up. If you're able to submit a PR with a fix that would be great!

stephenberry:
This should all be working now.

PR #337, opened 2023-07-05
Handle string_view dereference nullptr

maor-da:
https://github.com/stephenberry/glaze/issues/335
The second issue is when `string_view` dereference `nullptr` of unset `char*` and similar.

### DIFF
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -194,10 +194,10 @@ namespace glz
                   dump<'"'>(b, ix);
                }
                else {
-                  const sv str = [&]() -> const std::decay_t<T>& {
+                  const sv str = [&]() -> sv {
                      if constexpr (std::is_pointer_v<std::decay_t<T>>) {
                         if (value == nullptr) {
-                           return const_cast<std::decay_t<T>>("");
+                           return "";
                         }
                      }
 
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -196,7 +196,7 @@ namespace glz
                else {
                   const sv str = [&]() -> sv {
                      if constexpr (std::is_pointer_v<std::decay_t<T>>) {
-                        if (value == nullptr) {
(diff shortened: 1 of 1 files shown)


stephenberry:
You shouldn't need to return a const reference. Instead, just return by decltype(auto) or just by value and the optimizer should handle it just fine.

maor-da:
But if I return by value the `string_view` will point to R-value and will dispose the `std::string` right after assignment.

maor-da:
Since the `value` is valid outside the lambda it is ok to return reference to r-value. I think we need to suppress this error in gcc.

stephenberry:
Don't return a std::string from the lambda. Instead do this:

```c++

const sv str = [&]() -> sv {
                     if constexpr (std::is_pointer_v<std::decay_t<T>>) {
                        if (value == nullptr) {
                           return "");
                        }
                     }

                     return value;
                  }();
````

maor-da:
About the code, it gets to the `constexpr` when the type is a pointer.
I had to bypass the compiler assumption that reference can't be null `error: the compiler can assume that the address of ‘value’ will never be NULL`.

stephenberry:
Thanks! I'll try to look over this fix soon.

stephenberry:
@maor-da Can you provide a test case for what you're trying to do? You can't write to a char* buffer without memory already being allocated. We probably should guard against this with Glaze, but Glaze lets you do some dangerous things for those who handle their own memory.

stephenberry:
@maor-da Can you provide a test case for what you are trying to accomplish?

stephenberry:
Thanks for your changes, I think I'm going to alter the constexpr logic to only check for nullptr when writing a char* and not a char[]. I can update this pull request.

stephenberry:
Okay, I took some of your changes here and used them in a separate pull request #340. I changed around some of the concepts and type handling so that I could compile time switch on char array types and avoid the gcc warning as well as reduce runtime overhead of the check. Let me know if this fixes your active issue.

PR #338, opened 2023-07-06
Increased json schema support

jbbjarnason:
# Edit:
Implement a json schema metadata object to assign to parameters of glz::object.

Concerns: 
- Default value of objects or arrays, it would be nice if glz::write_json could be used but I think it is too much of an effort and or not possible. So users will need to write the json as string_view for those variables.

Nice to haves:
- With the new glz::merge it would be nice to implement the `glz::meta<glz::schema>` as object for its variables and within schema_ref do something like `glz::object("$ref", &T::ref, glz::merge<glz::schema>(&T::attributes));` But I am not sure this is any better than the current implementation

# Original description:

This pull request is not complete and I would like to get feedback from you before continuing. I will begin with describing the problem:
We are using glaze to publish json schema to frontend such as https://github.com/ui-schema/ui-schema/ to make configuration UI based on json for the backend. By doing so we need to be able to specify more metadata than the `comment`/`description` string_view is.

This pull request suggests a way to inject a metadata struct into the glz::Object by declaring a constexpr value within the struct to get passed the filter of make_group construction.

Please share your thoughts on this.

Reference: https://www.learnjsonschema.com/


stephenberry:
I really like this and I think this is the proper place to pass in schema parameters, replacing the comment.

I do think glz::schema::property_attributes class should be named more succinctly, because it is user facing.
I would prefer to just remove the schema namespace and turn glz::schema::property_attributes into `glz::schema`.

This would change: `glz::schema::property_attributes{.minimum = 2}` to `glz::schema{.minimum = 2}`

Also, I would rather not have member names that begin with an underscore, as these are best reserved for user defined literals. Instead change `_const` to something like `const_` or `constant`.

stephenberry:
I do want to make functions like glz::write_json constexpr someday. However, we need better support for constexpr std::string and std::vector before it really makes sense. And, static constexpr in functions via C++23 will also be immensely helpful for cleaner code. So, for now we need to wait until these features are more broadly implemented.

stephenberry:
I'm planning to look into this in more detail soon. I'll probably improve merge first to see if it is useful here.

stephenberry:
@jbbjarnason 
I solved the merge problem by just making `schema` hold `ref`.
I pushed some other minor cleanup.
Let me know if you have any concerns. I'll probably merge soon.

jbbjarnason:
Looks good 👍

stephenberry:
I've enabled using constexpr std::vector if supported by the compiler

PR #339, opened 2023-07-07
Fix std::min and std::max issue when compiling with MSVC

alfedotov:
Fix https://github.com/stephenberry/glaze/pull/172 one more time 😄


stephenberry:
Haha, thanks for this!

PR #340, opened 2023-07-08
Fixed char* null issue and better string type concepts

stephenberry:
fixed char* nullptr issue pointed out by maor-da


stephenberry:
Addressed the null pointer issue from #335

PR #343, opened 2023-07-10
JSON RPC cleanup

stephenberry:


stephenberry:
@jbbjarnason,
Can you look at this pull request and let me know if it looks good and works for you?

The big breaking change is that both the `rpc::server` and `rpc::client` take a new `rpc::method`, which cleans up the code from a user perspective.

But, there are also other cleaning changes and using std::string_view rather than std::string where I think it makes sense for performance.

jbbjarnason (approved):
This looks nice, I tested it a bit and it seems to be okay.

PR #344, opened 2023-07-12
support string_view as map key; eliminate unnecessary copy

SlaaneshChosen:


stephenberry:
Thanks for this pull request, however your changes actually introduce an extra copy instead of getting rid of one. Notice that the old code used `std::string& key = string_buffer();`, which gets a `static thread_local` reference to a `std::string` and is not a copy. It is reused memory on a given thread.

However, your new code has `value[k_t(key)]`, which will cause a new allocation and copy when `k_t` is a `std::string`.

If you agree, just revert this specific change on your pull request. Or, we can discuss further.

stephenberry:
@SlaaneshChosen 
Thanks for your further explanation. However, if the map key is a `std::string`, then passing the `std::string` as the key argument will not cause an allocation if the key exists. Whereas, in your code the allocation must always happen. `value[key]` does not allocate if the key exists, which is often the case.

In the current code there are no new allocations if the key exits in the map. Your code introduces allocations even if the key exists when the key is a `std::string`.

You are correct that we can avoid copying characters when the key is a string_view type. And, that string_view should point to the buffer being passed in and not the reused `string_buffer()`.

I think we just need to have a constexpr switch on the key type. I'll make the changes to this pull request and you can look over them. Thanks for your help!

stephenberry:
@SlaaneshChosen 
This is associated with another error I need to fix, so I'm merging this. Let me know if you have any more thoughts or concerns.

SlaaneshChosen:
@stephenberry 
I used `static std::string` seldomly and thought it behaves like a normal `std::string` in my last comment, and today I learnt it. 

Your changes are systematic and generalized! Thank you for your time to review and update the codes.

Still, there may be a small issue in line 1378 in `include/glaze/json/read.hpp`: `str_t` is a concept to check whether the object is `std::string` or `std::string_view`.  I think the constexpr is aimed to make sure that the key type is `std::string`;  `std::is_same_v<std::remove_cv_ref<T::first_type>, std::string>` might be better?
I just have a glance over the code changes on the github webpage, and cannot verify this in debugger/IDE for now.
For your information.

PR #348, opened 2023-07-14
properly serialize maps skipping nulls

justend29:
Previously, serializing map-like objects would incorrectly write commas when entries were skipped. This was because although the entry itself was skipped, the separator was not.

This PR changes that behaviour, and amalgamates some basic functionality used in the map serialization and many other serialization functions: writing entry separators and skipping members. Additionally, skipping nulls on pairs (single maps) has also been completed. With tests, of course.

### DIFF
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -320,6 +320,16 @@ namespace glz
          }
       };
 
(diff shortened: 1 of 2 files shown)


stephenberry:
Sweet, thanks for this fix!

Issue #355, opened 2023-07-19
Incorrectly parsed sets. Why is default value static?


justend29:
While parsing to a `std::set` of objects with a `std::set` field, I noticed that the field of subsequent objects contained an accumulation of the field entries from all previous object entries.

A reproduction of what I mean:
```c++
struct Thing { std::set<int> field; };
std::set<Thing> things;
const auto input_string = R"([
  {"field": [1] },
  {"field": [2,3]},
  {"field": [4,5]}
])";
glz::read_json(things, input_string);
// parsed incorrectly
assert(glz::write_json(things)  == R"([
  {"field": [1] },
  {"field": [1,2,3]},
  {"field": [1,2,3,4,5]}
])";
```

Tracking this down came to the use of `static` on the parsing for sets [here](https://github.com/stephenberry/glaze/blob/main/include/glaze/json/read.hpp#L813):
```c++
               if constexpr (sizeof(V) > 8) {
                  static thread_local V v;  // <---- static used here
                  read<json>::op<Opts>(v, ctx, it, end);
                  if (bool(ctx.error)) [[unlikely]]
                     return;
                  value.emplace(v);
               }
               else {
                  V v;
                  read<json>::op<Opts>(v, ctx, it, end);
                  if (bool(ctx.error)) [[unlikely]]
                     return;
                  value.emplace(std::move(v));
               }
```

This use of 'static' causes subsequent entries in a set not to be default constructed - instead, the previously parsed value is used as the "default". Consequently, operations that amend the "default" value, like parsing a `std::set` field, as opposed to replacing the "default" value, like parsing an `int`, contain remnants of previously parsed entries.

Why is `static` used here? It seems to be for a performance reason. Can it be removed or is there an alternative solution to this issue?

stephenberry:
@justend29 
Thanks for pointing this out. In this case because we are always emplacing the item, we don't need to use `static thread_local`.

I've merged in more std::set tests (#356), including what I thought matches your issue, along with removing `static thread_local`. However, even before removing `static thread_local`, my test case that was designed to match yours worked. So, you might want to check my test case called `"std::set<set_object>"` to see how your use case differs from this one.

justend29:
You're right with my little example, it doesn't illustrate the problem.
I was trying to come up with something simple, instead of my actually use-case, and was a little too confident in it.

I'll close the issue but will soon try to correct my example to reproduce the issue properly.

PR #359, opened 2023-07-20
legitimize test for sets containing amended entries accumulating values

justend29:
The tests added as part of issue https://github.com/stephenberry/glaze/issues/355, in PR https://github.com/stephenberry/glaze/pull/356, were based on an example I gave which didn't properly reproduce the issue.

This PR updates the tests to reflect the now reproducing example in the issue.


stephenberry:
Thanks!

PR #394, opened 2023-08-10
Thread pool improvements

stephenberry:


stephenberry:
This should be good to go, but I'm going to be testing this code more before merging.

PR #407, opened 2023-08-21
fix name_v for long long and unsigned long long

jbbjarnason:
This is not true: `static_assert(std::is_same_v<int64_t, long long>);` resulting in long long being glz::unknown.

Added specialization for long long and unsigned long long.


stephenberry:
Thanks for this!

PR #408, opened 2023-08-21
Fix number of array elements

huangminghuang:
This PR fixes the counting of json array elements when the element type is object or array.


stephenberry:
Thanks for this fix!

PR #437, opened 2023-09-21
fix write of long long and unsigned long long

jbbjarnason:
compare integer size in bytes, not bits


stephenberry:
Thanks for this fix! That was a dumb error :)

jbbjarnason:
@stephenberry could you make 1.4.3 release so I can update vcpkg and use this fix?

stephenberry:
@jbbjarnason I added the 1.4.3 release

PR #442, opened 2023-09-25
Named null

jbbjarnason:
Use case 

```cpp
struct named_always_null : std::monostate {};
template<>
glz::meta<named_always_null> { static constexpr std::string_view name{ "my name" };

std::variant<named_always_null, int, std::string> my_variant{};
```
`glz::write_json_schema` ... would return title as the described name in `named_always_null`

Alternative would be defining `glz::to_json<>` `glz::from_json<>`

What are your thought on this?

### DIFF
--- a/include/glaze/core/common.hpp
+++ b/include/glaze/core/common.hpp
@@ -483,7 +483,7 @@ namespace glz
 
       template <class T>
       concept always_null_t =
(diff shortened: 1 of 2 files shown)


stephenberry:
I think I understand your question, but could you provide a JSON/C++ example of a simple use case?

stephenberry:
I just mean in a comment here.

stephenberry:
After looking over your changes and understanding your use case I think this is fantastic. Thanks!

PR #443, opened 2023-09-25
Fix: Arbitrary inputs accepted as integers

justend29:
An example of what was passing:
```c++
#include <cstdint>

#include <glaze/json/read.hpp>

template<typename T>
void should_be_error() {
  const auto parsed = glz::read_json<int>("asdf;lkajsdf");
  assert(parsed.has_value());
  assert(parsed.value() == 0);
}

int main() {
  should_be_error<std::int8_t>();
  should_be_error<std::int16_t>();
  should_be_error<std::int32_t>();
  should_be_error<std::int64_t>();
  should_be_error<std::uint8_t>();
  should_be_error<std::uint16_t>();
  should_be_error<std::uint32_t>();
  should_be_error<std::uint64_t>();
}
```

The following changes primarily test if the first character is a digit in `stoui64()` instead of assuming it is. Tests and a little refactor of stoui64 follow.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -1569,21 +1569,39 @@ namespace glz
                            ctx.error = error_code::unknown_key;
                            return;
                         }
+                        else {
+                           // We duplicate this code to avoid generating unreachable code
+                           skip_ws<Opts>(ctx, it, end);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                           match<':'>(ctx, it, end);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                           skip_ws<Opts>(ctx, it, end);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+
+                           skip_value<Opts>(ctx, it, end);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                        }
                      }
+                     else {
+                        // We duplicate this code to avoid generating unreachable code
+                        skip_ws<Opts>(ctx, it, end);
+                        if (bool(ctx.error)) [[unlikely]]
+                           return;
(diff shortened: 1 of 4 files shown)


stephenberry:
Thanks for this fix! And, I threw this into QuickBench to just ensure that the refactoring didn't have any surprise performance changes, and your new code is actually 10% faster than the old code! https://quick-bench.com/q/2lkOH9X16AUaXnigATNcEsRV9Wc

justend29:
@stephenberry  That's exciting :) Gotta love iterators. Any idea why MSVC tests are taking forever in CI?

stephenberry:
@justend29, from my experience MSVC takes significantly longer than clang and gcc to build glaze. And, this extra build time doesn't end up with better performance, e.g. we avoid force inlining on MSVC because it then makes compilation absolutely terrible, so performance is often much worse on MSVC. Also, the VM is windows, so there's overhead there as well.

stephenberry:
@justend29 On gcc 13.2 your new code is 40% faster! That's crazy

justend29:
> @justend29, from my experience MSVC takes significantly longer than clang and gcc to build glaze. And, this extra build time doesn't end up with better performance, e.g. we avoid force inlining on MSVC because it then makes compilation absolutely terrible, so performance is often much worse on MSVC. Also, the VM is windows, so there's overhead there as well.

That all makes sense, but the test job in the MSVC CI pipeline is taking so long it's [timing out](https://github.com/stephenberry/glaze/actions/runs/6303195789/job/17111955873). The build and all take slightly longer, but the tests seem to be doing nothing. The VM and slow MSVC couldn't be the culprit there, eh? Are the runners ok?

stephenberry:
Yeah, I'm seeing that problem with MSVC, the test is taking forever. I'm looking into it now.

PR #445, opened 2023-09-26
clang + libstdc++ workaround

huangminghuang:
This is a workaround for the compilation issue with clang++ with libstdc++ as described in https://github.com/llvm/llvm-project/issues/50248.

https://godbolt.org/z/bndzMjMsM demonstrates this problem.


stephenberry:
Thanks for this fix!

PR #452, opened 2023-10-01
Parse empty meta object.

justend29:
Support parsing empty objects registered through glaze's `meta` registration system.


stephenberry:
Thanks for adding handling for empty objects and the additional cleanup.

PR #460, opened 2023-10-17
Using std::span for schema required and examples

stephenberry:


include/glaze/json/schema.hpp
```diff
@@ -38,9 +36,7 @@ namespace glz
       std::optional<std::uint64_t> min_properties{};
       std::optional<std::uint64_t> max_properties{};
 //      std::optional<std::map<std::string_view, std::vector<std::string_view>>> dependent_required{};
-#ifdef __cpp_lib_constexpr_vector
-      std::optional<std::vector<std::string_view>> required{};
-#endif
+      std::optional<std::span<const std::string_view>> required{};
```
jbbjarnason, line 39: Nice!

stephenberry: I experimented with std::span<const char*>, but it didn't play nice, and I prefer using std::string_view, even though it requires naming the number of arguments in the std::array.

But, I've added support for:
```c++
template <>
struct glz::meta<Thing>
{
   using T = Thing;
   static constexpr std::string_view name = "Thing";
   static constexpr std::array<std::string_view, 2> required = { "thing", "i" };
...
```

Where the `required` field here will be applied to the `Thing` object. Or, you can add `required` field to schema elements within the metadata.

stephenberry:
Supports the following syntax for adding examples and required keys to JSON schema:

```c++
template <>
struct glz::meta<Thing>
{
   static constexpr std::array<std::string_view, 2> required = { "thing", "i" };
   static constexpr std::array<std::string_view, 1> examples = { R"({"thing":{},"i":42})" };
```

PR #462, opened 2023-10-21
Feature/devcontainer one click

Jason5480:
The goal of this pull request is to make it possible to for any developer to easily set up the C++ dev environment with all the tooling (clangd/format/tidy etc.) and start contributing right away.

This can be done using codespaces, VSCode or any other IDE that supports devcontainers and CMakePresets!
I wrote the Dockerfile so that
* The ubuntu version (jammy) is well defined in the devcontainers.json file and will get updated on-demand
* Added gcc and explicit version flag for both clang (17 latest) and gcc (13 latest)
* Option to use clang (default USE_CLANG) ( This can also be controlled through the CC and CXX env)
* Install the latest CMake from kitware. FIX problem with default cmake version being too old and ninja generator not installed
* Create non-root user and login by default to avoid potential security issues
* Made the clangd to work right away since the export of the compile_commands.json file was broken
* Fix an issue with line endings when the host was on windows by adding the .gitattributes file
* Added some files to .gitignore

Now with only one click and "open with VSCode" you will have the environment set up and running if the host machine has Docker installed. Or even better if you click the codespaces button you can write directly into the browser with full clangd support and formatting.

Issues to discuss:
A) I usually commit and push from inside the devcontainer and I use two more extensions with ids "eamodio.gitlens", "mhutchie.git-graph". When I find them to be very convenient. I should not install them if this is not useful for most of the delopers (e.g. if most of them use the git cli directly). So shall I add them or not.

B) Editorconfig plugin with id "editorconfig.editorconfig" might not agree with clang-format. Is it wise to have two linters? Every modern C++ IDE should support clang-format nowadays. Shall I remove the existing plugin?

C) I also tried make the CMake installed version to be explicit, but I failed to find a proper way to do it. The kitware apt repository doesn't provide a version option. Maybe there is an other way out there to do it and I missed it, or using the latest CMake version is what is expected nowadays.


stephenberry:
Fantastic! Thanks!

stephenberry:
@Jason5480 

Concerning your discussion points:

A) I typically use the built in VSCode git features in codespaces, so as long as it doesn't affect these, I'm good.

B) If there's a conflict I'd prefer to use clang-format. So you can remove Editorconfig plugin.

C) I'd be happy just using the latest CMake version.

Jason5480:
> @Jason5480
> 
> Concerning your discussion points:
> 
> A) I typically use the built in VSCode git features in codespaces, so as long as it doesn't affect these, I'm good.
> 
> B) If there's a conflict I'd prefer to use clang-format. So you can remove Editorconfig plugin.
> 
> C) I'd be happy just using the latest CMake version.

Cool the git extensions do not get in the way of the VSC build in git functionality. I pushed a new commit to add those and remove the editorconfig extension. I didn't remove .editorconfig file because I don't know how this would affect other IDEs. Since the discussion points I was not sure about were resolved this PR is ready for review. Thanks for that awesome library BTW!

stephenberry:
This looks excellent, thanks!

PR #465, opened 2023-10-24
Using GLZ_T rather than T for templated classes

stephenberry:


stephenberry:
This solves issues when using GLZ_LOCAL_META on classes that are templated with `<class T>`, where T would collide with the old macro code.

PR #466, opened 2023-10-27
Deducing return type mismatch in seek_impl

justend29:
Literally just this change of return type:  https://github.com/stephenberry/glaze/commit/6f3a59f914101620815be0418d6f27f761523cb3#diff-a90b69deeada8429e785b7ccfb47d1f30a32626da32dea5a90c6f1f41a6b655fL580-R581. When `N!=0` this code isn't compiled, so the mismatched return type snuck through.

But, there were also plenty of clang-format changes applied by the auto-formatter. We should make it such that the source files are always formatted before getting merged such that PRs don't have huge changes caused by format changes. I could turn off my format on save, but that seems like the wrong solution.


stephenberry:
Thanks for this fix.
I'm slightly confused by your comments about clang-format. Is your auto-formatter not using the clang-format file?
I'm also okay with the formatting getting fixed after merging, which is set up in the actions.

justend29:
@stephenberry All I was mentioning is there's consistently merged code that _hasn't_ been auto-formatted, meaning pull-requests are cluttered with unrelated formatting changes that inevitably come from the ubiquitous use of auto-formatters. Maybe it's just me, and I can do something better (let me know), but reviewing 20 files changed for a 2-line fix isn't ideal.

Am I interpreting the situation right? Or, am I just locally mangling all the formatting that's nice upstream?

justend29:
Oh, I see the action that commits clang-format changes, now. It uses [clang-format 15](https://github.com/stephenberry/glaze/blob/main/.github/workflows/clang-format.yml#L22), while the dev container uses version 17. Maybe what's actually happening is the action pushes formatting changes w/ version 15 that will inevitably be changed by version 17.

stephenberry:
Oh, that's probably what is happening. We should update the action to use the latest version.

PR #470, opened 2023-10-29
fix map with quoted_num

huangminghuang:
This PR fixes the problem to use quoted_num with map.


stephenberry:
Excellent, thanks for this fix!

Issue #471, opened 2023-10-30
Bad control character generated, potential buffer-overflow


sweihub:
Hi

I have a lots of error in my Rust side logs, which were generated by cpp side glaze json, after tracing, I can confirm that the corrupted JSON string was sent from my CPP appplication, which Rust side detects  `257`-th character was a bad `control character (\u0000-\u001F)`,  you can find it after `side` keyword below. What bothers me much is I cannot reproduce with the same data, but it happens frequently in production environment. Can you identify this issue, the `257` is near to `256`, `2*128`, so I guess is a potential buffer-overflow issue?

`
[2023-10-30 08:55:09.833] [ERROR] server: error: control character (\u0000-\u001F) found while parsing a string at line 1 column 257
Raw data from ws 16551:
{"id":706,"method":"save_quote","params":{"time":1698627291351456360,"action":"send","quote":"kill","account":"603302","uid":11,"session_id":1,"request_id":41,"state":0,"order_id":"2023103000180021","exchange":"CZCE","type":"","tif":"","offset":"","side":"^@^@"symbol":"SPD RM401&RM403","price":0,"quantity":0,"traded":0}}
`

1. Sent from cpp, a `control character` was there, the `side` ought to be empty
![cpp_send](https://github.com/stephenberry/glaze/assets/5334841/ebee8eef-4c8b-4669-a047-d9ddd90f3d73)
2, Rust side, detected lots of corrupted data (`control character`)
![rust_recv](https://github.com/stephenberry/glaze/assets/5334841/662ae452-64bf-470e-9b0a-d1bf3ceae668)

jbbjarnason:
This sounds serious, @sweihub could you provide an example of code reproducing this? Maybe just dump to file of given conversion?

sweihub:
Yes, it's serious, the following is another error JSON string, but same location `257`. I will add detecting code to dump the all fields once it happens again. Will get back to you guys.

![image](https://github.com/stephenberry/glaze/assets/5334841/45df5b11-88e1-4299-abba-c3409fb68630)

sweihub:
Good News, I reproduce the crash!!

**code**
```cpp

#include "core/json.hpp"
#include "glaze/core/macros.hpp"
#include <iostream>
#include <optional>

namespace core {
    template <class T, class Buffer>
    inline Buffer &write(T &value, Buffer &buffer)
    {
        glz::write<glz::opts{}>(value, buffer);
        return buffer;
    }
}

[...]

PR #473, opened 2023-10-30
Fixed empty string writing allocation bug

stephenberry:
When writing out empty strings not enough memory was allocated in the buffer.


stephenberry:
Fixes #471

PR #479, opened 2023-10-30
Rectify circular include preventing compilation

justend29:
The aggregate header `glaze/json.hpp` contains `glaze/json/quoted.hpp`, which was transitively included by its use of `glaze/json.hpp`.

This caused compilation errors in my code, so it may prevent others from using Glaze, too. Might be worth a fix version bump


stephenberry:
Thanks for this.

PR #482, opened 2023-11-01
io_state for get_member and read/write lambdas

stephenberry:


stephenberry:
The issue with this approach is that we hide the bifurcation logic of the read/write after a get_member call. When doing operations like `seek`, we do not know whether a `read` or `write` will take place, or perhaps both. This creates terrible limitations to seeking and similar operations. It is better to make types like the current `custom` that explicitly define the read/write behavior.

PR #483, opened 2023-11-01
Much cleaner and more flexible glz::custom

stephenberry:


stephenberry:
## glz::manage breaking changes!

`glz::manage` has a new API: `glz::manage<&T::x, &T::read_x, &T::write_x>` 

The read and write functions can also be lambda functions. And, the functions registered no longer take in a `glz::manage_state`.

These changes make `glz::manage` far more flexible.

PR #485, opened 2023-11-02
Fix negation bug in json_exceptions.hpp

### LINKED ISSUE #484: Bug in json exception wrappers causes (all) valid json reads to fail.
There is a bug in the json exception wrappers. A minimal failing example is the following.
```c++
#include <iostream>
#include <glaze/glaze_exceptions.hpp>
#include "glaze/core/macros.hpp"

struct Foo {
	int x;
};

GLZ_META(Foo, x);

int main([[maybe_unused]] const int argC, [[maybe_unused]] char* argV[]) {
	std::string buffer = R"({"x":287})";
	auto res = glz::ex::read_json<Foo>(buffer);
	std::cout << res.x;
	return 0;
}
```

In my understanding the reason is a mishandling of an variable of type "glz::expected" as an error code. See accompanying pull request.

Haatschii:
A variable of type "expected" converts to true if there was no error and false if there was one. This commit fixes a bug where an expected variable was treated as an error code (which behaves the other way round).

fixes #484

### DIFF
--- a/include/glaze/exceptions/json_exceptions.hpp
+++ b/include/glaze/exceptions/json_exceptions.hpp
@@ -31,7 +31,7 @@ namespace glz::ex
    [[nodiscard]] T read_json(Buffer&& buffer)
    {
       const auto ex = glz::read_json<T>(std::forward<Buffer>(buffer));
-      if (ex) {
+      if (!ex) {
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for this fix!

PR #491, opened 2023-11-03
Add exception wrapper functions for the ndjson read and write functions

### LINKED ISSUE #488: Missing exception wrappers for newline delimited JSON
As far as I can see there are no exception wrapper functions in `glz::ex` for the NDJSON functions. Is there a specific reason for this, or was there just no demand? In the latter case I could write a pull request adding these functions if you like?

Haatschii:
Functions are adapted from the existing functions for normal json.

Added `const` in some existing functions.

resolves #488


stephenberry:
Sweet, this looks good!

PR #502, opened 2023-11-04
Support writing unsized, uncommon ranges.

justend29:
write functions for arrays and objects expected const input values, which prevents the writing of lazy ranges that mutate internal state. The qualifiers were changed and tests were added.

It would be really nice to support writing any range as a string, but that's something to tackle for a future day.


justend29:
Note to self: squash future merges

PR #509, opened 2023-11-07
Support parsing any map key type from string

### LINKED ISSUE #507: std::map with custom class as keys
The following minimal example fails to compile:

```c++
struct Foo {
	int64_t value;
};

template<>
struct glz::meta<Foo> {
	static constexpr auto value{ &Foo::value };
};


struct Bar {
	std::map<Foo, int> values;
	 
	GLZ_LOCAL_META(Bar, values);
};


int main() {
	auto x = glz::read_json<Bar>("{\"values\":[[3,7], [7,3], [73,37]]}");
}
```

Following the error message it seems that the reason is a failing `static_assert`, which, as far as I understand, enforces that only string-like and arithmetic types can be used when reading a `std::map` (I might be missing something, though). 

Is this indeed the case, that only string-like and arithmetic types are allowed as keys?

justend29:
Closes #507

Additionally, although `writiable_map`s could accept key types other than strings, no key type was tested that *contained* a string. An issue existed where these weren't properly escaped to be contained within the string keys of all JSON objects, thereby writing out invalid JSON.
As an example, a key value of `"["first","second"]"` would be written instead of `"[\"first\",\"second\"]"`.

Note that the implementation for both writing and reading object keys of custom types uses `glz::quoted_t`, and therefore serializes/deserializes to/from the thread local string internal to `quoted_t`. However, its possible that the `quoted_t` implementation could be changed by wrapping the output iterators in some smart output iterators that escape all assigned escape sequences. 

Lastly,  `glaze/json/read.hpp` and `glaze/json/write.hpp` are both included in `glaze/json/quoted.hpp`, but also use `quoted_t`, thereby needing its definition. This likely alludes to organizational changes, but for the immediate time it means that to avoid circular includes, users who wish to read/write custom keys must also include `glaze/json/quoted.hpp`. Thoughts on how to correct that?

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -11,6 +11,7 @@
 #include <locale>
 #include <ranges>
 #include <sstream>
+#include <type_traits>
 
 #include "glaze/core/common.hpp"
 #include "glaze/core/format.hpp"
@@ -25,6 +26,11 @@
 
 namespace glz
 {
+
+   // forward declare from json/quoted.hpp to avoid circular include
+   template <class T>
+   struct quoted_t;
+
    namespace detail
    {
       // Unless we can mutate the input buffer we need somewhere to store escaped strings for key lookup and such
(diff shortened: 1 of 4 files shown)


stephenberry:
I like this approach and the default syntax. I do think we can move away from using `glz::quoted` in the future for better performance, but that's not a priority right now. Thanks for your work on this!

stephenberry:
Feel free to merge this.

PR #511, opened 2023-11-09
Add missing inline specifier for mulhi64 function

TheLuckyCoder:
This fixes multiple definitions error when compiling for 32-bit Architectures.

This issue was observed when compiling with GCC using LTO for Armv7


justend29:
Thank you for finding and fixing this!

PR #512, opened 2023-11-10
supporting constexpr values/types in variant

jbbjarnason:
Added support for handling reads of constexpr struct values in variants.

Introduced type handling of const types for glaze value, and split the const types from the non const types during the parse of variant types.

### DIFF
--- a/include/glaze/core/common.hpp
+++ b/include/glaze/core/common.hpp
@@ -535,6 +535,9 @@ namespace glz
       concept glaze_value_t =
          glaze_t<T> && !(glaze_array_t<T> || glaze_object_t<T> || glaze_enum_t<T> || glaze_flags_t<T>);
 
+      template <class T>
+      concept glaze_const_value_t = glaze_value_t<T> && std::is_pointer_v<glz::meta_wrapper_t<T>> && std::is_const_v<std::remove_pointer_t<glz::meta_wrapper_t<T>>>;
+
       template <class From, class To>
       concept non_narrowing_convertable = requires(From from, To to) {
 #if __GNUC__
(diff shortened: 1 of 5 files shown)


stephenberry (changes requested):
@jbbjarnason I think this is looking good. My one thought is that you might want to use `glz::tuplet::tuple_cat` and `glz::tuplet` rather than `std::tuple`. We try to use `glz::tuplet` internally because it significantly improves compile times compared to `std::tuple`.

jbbjarnason:
very unofficial compilation time benchmarking

before
```bash
jonb@jonb-desktop ~/P/glaze (support-direct-constexpr-read)> time cmake --build build --target json_test
[0/2] Re-checking globbed directories...
[2/2] Linking CXX executable tests/json_test/json_test

________________________________________________________
Executed in   43.86 secs    fish           external
   usr time   40.95 secs  246.00 micros   40.95 secs
   sys time    2.76 secs  351.00 micros    2.76 secs
```

after
```bash
jonb@jonb-desktop ~/P/glaze (support-direct-constexpr-read) [1]> time cmake --build build --target json_test
[0/2] Re-checking globbed directories...
[2/2] Linking CXX executable tests/json_test/json_test

________________________________________________________
Executed in   42.71 secs    fish           external
   usr time   40.01 secs  233.00 micros   40.01 secs
   sys time    2.58 secs  345.00 micros    2.58 secs
```

Does not show any major difference but it's a bit better

stephenberry:
Thanks for the benchmark and making those changes. I guess there's not a significant difference in this context, but I've seen in make a big difference elsewhere.

stephenberry:
I think this can be merged, if you think it's ready.

PR #513, opened 2023-11-11
Fixed issue with read_file_binary always outputting true for bools

dabader:
Added test case to binary_test. All tests passing.


stephenberry:
Thanks for this fix!

PR #519, opened 2023-11-14
add emplacer style reader for glz::custom

SlaaneshChosen:
`glz::custom` accepts a function with a signature like `void read_x(const std::string& s)` as its `read function` for a member. In its implement, a std::string will be built on the stack and modified when parsing the json. Then the variable will be called as the argument in `read_x`. The overhead is very small for those data structures that have an efficient `std::move`, like `std::vector`. For those with a less efficient move function, the copy is inevitable. And for a `std::span`, the temporary variable is empty, which makes it not work.

I added some codes to support another kind of function, with a signature like `void read_x(const std::function<void(std::string&)>& x_builder)`. This will help to reduce runtime overhead for some cases, and support those data structures do not **own** the memory resource.

In theory, the read function should support any callable with correct sigature as its argument. However, when registering `glz::object`, a concrete type must be provided. Thus I constrain the input must be a `std::function` with unary augument.

### DIFF
--- a/include/glaze/json/custom.hpp
+++ b/include/glaze/json/custom.hpp
@@ -48,11 +48,22 @@ namespace glz
                         (value.val.*(value.from))();
                      }
                      else if constexpr (std::tuple_size_v<Tuple> == 1) {
-                        std::decay_t<std::tuple_element_t<0, Tuple>> input{};
-                        read<json>::op<Opts>(input, ctx, it, end);
-                        if (bool(ctx.error)) [[unlikely]]
-                           return;
-                        (value.val.*(value.from))(input);
+                        using Input = std::decay_t<std::tuple_element_t<0, Tuple>>;
+                        const auto dummy_emplacer = [](auto& i) {};
+                        if (std::invocable<std::remove_cvref_t<decltype(value.val.*(value.from))>,
+                                           decltype(dummy_emplacer)>) {
+                           const auto emplacer = [&](auto& input) { read<json>::op<Opts>(input, ctx, it, end); };
+                           (value.val.*(value.from))(emplacer);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                        }
+                        else {
+                           Input input{};
+                           read<json>::op<Opts>(input, ctx, it, end);
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                           (value.val.*(value.from))(input);
+                        }
                      }
                      else {
                         static_assert(false_v<T>, "function cannot have more than one input");
--- a/include/glaze/json/custom.hpp
+++ b/include/glaze/json/custom.hpp
@@ -49,11 +49,8 @@ namespace glz
                      }
                      else if constexpr (std::tuple_size_v<Tuple> == 1) {
                         using Input = std::decay_t<std::tuple_element_t<0, Tuple>>;
-                        const auto dummy_emplacer = [](auto& i) {};
-                        if (std::invocable<std::remove_cvref_t<decltype(value.val.*(value.from))>,
-                                           decltype(dummy_emplacer)>) {
-                           const auto emplacer = [&](auto& input) { read<json>::op<Opts>(input, ctx, it, end); };
-                           (value.val.*(value.from))(emplacer);
+                        if constexpr (std::is_base_of_v<std::_Function_base, Input>) {
+                           (value.val.*(value.from))([&](auto& input) { read<json>::op<Opts>(input, ctx, it, end); });
                            if (bool(ctx.error)) [[unlikely]]
                               return;
                         }
--- a/tests/json_test/json_test.cpp
+++ b/tests/json_test/json_test.cpp
@@ -5278,6 +5278,10 @@ struct custom_encoding
    std::string y{};
    std::array<uint32_t, 3> z{};
 
+   //   std::span<uint32_t> w{buf.data(), 16};
+   //   std::array<uint32_t, 32> buf{};
+   int32_t w{};
+
    void read_x(const std::string& s) { x = std::stoi(s); }
 
    uint64_t write_x() { return x; }
@@ -5289,21 +5293,30 @@ struct custom_encoding
       z[0] = 5;
       return z;
    }
+
+   void read_w(std::function<void(int32_t&)> w_builder)
+   {
+      w_builder(w);
+      static_assert(std::invocable<decltype(w_builder), int32_t&>);
+   }
 };
 
 template <>
 struct glz::meta<custom_encoding>
 {
    using T = custom_encoding;
-   static constexpr auto value = object("x", custom<&T::read_x, &T::write_x>, //
-                                        "y", custom<&T::read_y, &T::y>, //
-                                        "z", custom<&T::z, &T::write_z>);
+   static constexpr auto value = object(
+       "x", custom<&T::read_x, &T::write_x>, //
+                                         "y", custom<&T::read_y, &T::y>, //
+                                        "z", custom<&T::z, &T::write_z>,
+      "w", custom<&T::read_w, &T::w>
+   );
 };
 
 suite custom_encoding_test = [] {
    "custom_reading"_test = [] {
       custom_encoding obj{};
-      std::string s = R"({"x":"3","y":"world","z":[1,2,3]})";
+      std::string s = R"({"x":"3","y":"world","z":[1,2,3], "w":12 })";
       expect(!glz::read_json(obj, s));
       expect(obj.x == 3);
       expect(obj.y == "helloworld");
--- a/tests/json_test/json_test.cpp
+++ b/tests/json_test/json_test.cpp
@@ -5278,9 +5278,8 @@ struct custom_encoding
    std::string y{};
    std::array<uint32_t, 3> z{};
 
-   //   std::span<uint32_t> w{buf.data(), 16};
-   //   std::array<uint32_t, 32> buf{};
-   int32_t w{};
+   std::span<int32_t> w{buf.data(), 3};
+   std::array<int32_t, 3> buf{};
 
    void read_x(const std::string& s) { x = std::stoi(s); }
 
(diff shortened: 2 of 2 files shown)


sam20908:
> In theory, the read function should support any callable with correct sigature as its argument. However, when registering glz::object, a concrete type must be provided. Thus I constrain the input must be a std::function with unary argument.

RIP

tests/json_test/json_test.cpp
```diff
@@ -5297,26 +5302,39 @@ struct glz::meta<custom_encoding>
...
                                         "y", custom<&T::read_y, &T::y>, //
-                                        "z", custom<&T::z, &T::write_z>);
+                                        "z", custom<&T::z, &T::write_z>, //
+                                        "w",
+                                        custom<
+#if defined(__GNUC__) && (!defined __clang__)
+                                           &T::read_w
+#else
+                                           &T::w
+#endif
```
justend29, line 5312: What's the specific feature clang doesn't offer that prevents `read_w` as a non-type template parameter?

include/glaze/json/custom.hpp
```diff
@@ -48,11 +48,22 @@ namespace glz
...
-                        (value.val.*(value.from))(input);
+                        using Input = std::decay_t<std::tuple_element_t<0, Tuple>>;
+#if defined(__GNUC__) && (!defined __clang__)
+                        if constexpr (std::is_base_of_v<std::_Function_base, Input>) {
+                           (value.val.*(value.from))([&](auto& input) { read<json>::op<Opts>(input, ctx, it, end); });
+                           if (bool(ctx.error)) [[unlikely]]
+                              return;
+                        }
+                        else
+#endif
```
justend29, line 59: Same here. What's the specific feature clang is missing?

include/glaze/json/custom.hpp
```diff
@@ -48,11 +48,22 @@ namespace glz
...
                      }
                      else if constexpr (std::tuple_size_v<Tuple> == 1) {
-                        std::decay_t<std::tuple_element_t<0, Tuple>> input{};
-                        read<json>::op<Opts>(input, ctx, it, end);
-                        if (bool(ctx.error)) [[unlikely]]
-                           return;
-                        (value.val.*(value.from))(input);
+                        using Input = std::decay_t<std::tuple_element_t<0, Tuple>>;
+#if defined(__GNUC__) && (!defined __clang__)
+                        if constexpr (std::is_base_of_v<std::_Function_base, Input>) {
```
justend29, line 53: Is the goal here just to verify that the provided argument is a std::function?

If so, I believe the use of std::_Function_base should be avoided. I don't think that's standardized, and just so happens to be the name of a type in the internal code of libstdc++. Furthermore, libstdc++ can be used as the standard library with clang, so the pre-processor check doesn't target this "feature", or non feature in this case.

Please, correct me if I'm wrong

justend29:
If I'm understanding you right, the two issues you've found are:
1. the custom read callbacks provided to `glz::custom` cannot accept a `std::span` as its argument, so an owning container, like a `std::vector` or `std::array`, must be used.
2. in such case, this variable then undergoes the cost of moving and/or copying the container to your destination

On point one, as shown below, parsing directly to `std::span` will work outside of `glz::custom`. Maybe we can find a solution to that issue directly instead of adding the nested callbacks. 
```c++
  std::array<int, 3> buf{};
  std::span<int, 3> view{buf};
  const auto read_error = glz::read_json(view, "[1,2,3]");

  assert(!read_error);
  assert(std::ranges::equal(view, std::array{1,2,3}));
  assert(std::ranges::equal(buf, std::array{1,2,3}));

```
On point two, I obviously haven't run any benchmarks, but I would be very surprised if the cost of copying/moving those types is more expensive than allocating the body of a `std::function` and allocating and copying to a `std::string`.

Additionally, with the current version of glaze, custom parsing with a non-owning buffer argument is possible through the use of `glz::raw_json_view`. Although non-owning, to construct it, glaze must internally iterate a portion of the input buffer to form the `raw_json_view` over the JSON value. Although this is quicker than properly parsing that value, it does incur costs that parsing directly to an owning buffer does not, and would be slower than a simple memcpy.

The following example illustrates the use of `glz::raw_json_view` with the use of `glz::custom` in a similar fashion to the test you've added in this PR. Please share a specific use case that this does not fulfill. I'll look further into why `span` cannot be used as the argument to the custom read callback.

```c++
#include <span>
#include <algorithm>
#include <string_view>
#include <cassert>
#include <array>
#include <span>

#include <glaze/core/common.hpp> // raw_json_view
#include <glaze/core/meta.hpp>
#include <glaze/json/custom.hpp>
#include <glaze/json/read.hpp>

struct My_type {
  // implement special member functions to avoid dangling view2
  My_type() = default;
  My_type(const My_type& other)
    : buf1{other.buf1}, buf2{other.buf2}, view2{buf2}, buf3{other.buf3}
  {}
  My_type operator=(const My_type& other)
  {
    buf1 = other.buf1;
    buf2 = other.buf2;
    view2 = buf2;
    buf3 = other.buf3;
    return *this;
  }
  My_type(My_type&& other) noexcept 
    : buf1{other.buf1}, buf2{other.buf2}, view2{buf2}, buf3{other.buf3}
  {}

  My_type& operator=(My_type&& other) noexcept {
    buf1 = other.buf1;
    buf2 = other.buf2;
    view2 = buf2;
    buf3 = other.buf3;
    return *this;
  }

  std::array<int, 3> buf1{};

  std::array<int, 3> buf2{};
  std::span<int, 3> view2{buf2};

  std::array<int, 2> buf3{};

  void custom_read1(const glz::raw_json_view value_json) {
    // accept non-owning argument, parse directly to buffer
    assert(value_json.str == "[1,2,3]");
    const auto read_error = glz::read_json(buf1, value_json.str);
  }

  void custom_read2(const glz::raw_json_view value_json) {
    // accept non-owning argument, parse to span on buffer
    assert(value_json.str == "[4,5,6]");
    const auto read_error = glz::read_json(view2, value_json.str);
  }

  void custom_read3(std::array<int, 2> buf) {
    // accept owning argument of parsed data, copy to buffer (very fast)
    assert(std::ranges::equal(buf, std::array{7, 8}));
    buf3 = buf;
  }
};

namespace glz {

template<>
struct meta<My_type> {
  using T = My_type;
      "member1", custom<&T::custom_read1, &T::buf1>,
      "member2", custom<&T::custom_read2, &T::view2>,
      "member3", custom<&T::custom_read3, &T::buf1>
    );
};

}

int main() {
  constexpr std::string_view serialized = 
    R"( {"member1":[1,2,3], "member2":[4,5,6], "member3":[7,8]} )";

  const auto my_value = glz::read_json<My_type>(serialized);
  assert(my_value.has_value());
  assert(std::ranges::equal(my_value->buf1, std::array{1,2,3}));
  assert(std::ranges::equal(my_value->buf2, std::array{4,5,6}));
  assert(std::ranges::equal(my_value->view2, std::array{4,5,6}));
  assert(std::ranges::equal(my_value->buf3, std::array{7, 8}));
}

```

SlaaneshChosen:
I must admit that the usage of `glz::raw_json_view` in read funtion is a much elegent and handy way to handle this issue, thus a nested callback is no longer needed.
Still, `std::span` with dynamic length cannot be used as the argument of a read function directly, .i.e, `void custom_read2(std::span<int32_t> value_json)`, because in the parsing process a brandly new `std::span` will be constructed on the stack and its default length is 0; parsed value will be skipped and not stored in the `std::span`. The behavior is somehow weird. Maybe this should be forbidden in the compile time?

sam20908:
Don't mind me if I don't know the design of the code, but why can't it accept any invocable function and it has to be std::function?

handgunsoldier:
I must admit that your 

> Don't mind me if I don't know the design of the code, but why can't it accept any invocable function and it has to be std::function?

When registering glz::object, a concrete type must be provided. 
If a function takes `lambda`, `auto` as its argument, it is actually a template function.

justend29:
@SlaaneshChosen I'll keep thinking about parsing to a span or span-like thing. Maybe a function factory can be accepted, maybe an initializer, maybe double callbacks, maybe the raw parsing iterators. I'm not sure right now. Seeing as the current need is fulfilled, the thought will rest on the back-burner of my mind.

To best understand the issue, could someone detail specifically what we're trying to accomplish with the span? I don't know if the test case serves as an adequate example of why `span` is needed since it's just referring to another buffer with the same interface, meaning that buffer type could just be accepted as the argument (`custom_read3`). Is it the copy of the array that we're trying to avoid, (`buf3 = buf`)?

SlaaneshChosen:
@justend29 
Thank you for your snippet, from which I learnt a lot. Indeed, this is enough for most cases. However, the performance issues still needs to be noticed.

For the sake of focusing the main topic, let us leave `std::span` aside and discuss `std:array` only. If there is **a quite large array** of `int64_t` (which means moving these values does not save any time than copying them) in the json string, there are 3 ways to parse it into a struct, and their performance costs are listed as follows.

1. `glz::raw_json_view` together with `glz::custom` to customize its read function. Extra costs: iterating over the string to construct the `view` of it
2. `std::array` together with `glz::custom` to customize its read function. Extra costs: memcpy the data from a temporary `std::array` on the stack to the given address
3. nested callback (`std::function`) with `glz::custom` to customize its read function. Extra costs: building the `std::function`, one call via function pointer

In the very beginning I try to add the nested callback, I naively thought that a lambda expression could be used as the nested callback and thus these extra costs could also be avoided mostly. I failed on this, and introducing `std::function` as its substitute, which brings costs as you pointed out.
I make this modification for that I often use `glaze` to parse json strings with very large array. Later I think this optimization could somehow be done, and make a prototype of it as the PR shows.
For your information.

justend29:
That's great, thanks. It illustrates exactly what you're trying to avoid such that the problem can be best targeted.

I'll mention some thoughts on your points, in case that helps:
- arrays are trivially copyable blocks on the stack. "Moving" arrays is just copying the elements, since there's no handle to the memory block that can be moved - they *are* the memory block. In any case, copying the memory is extraordinarily fast, as shown by many benchmarks.
- having *quite large* arrays should sometimes be avoided in preference for allocating containers to avoid stack overflows. Process stack size is limited and fixed to a few MB. As such, should the arrays be large enough that the copy should be avoided, it's likely best to use a `std::vector` to avoid stack overflows. They offer moves in a couple of CPU instructions, and the heap allocation costs can be mitigated through allocators, for which I would recommend off-the-shelf polymorphic allocators (`std::pmr::vector`).

Thanks, again. I'll keep thinking of solutions.

justend29:
@SlaaneshChosen I'll close this for now in favor of the alternative options presented. We can open a separate issue to deliberate specific ideas, should we need it.

Issue #508, opened 2023-11-07
support for std::bitset<>


WillisMedwell:
Support for std::bitset

```c++
auto nums = std::bitset<10>(2);
std::cout << glz::write_json(nums) << '\n'; 
```

fantastic library btw!

justend29:
Specifically, how would you want that serialized and how would you want that deserialized?

A string of 1's & 0's or an array of 1's & 0's would both make sense for humans. As for compactness, an array of the largest possible integers or a string of bytes, not characters, would be the smallest.

stephenberry:
I feel like a string of 1s and 0s would be a good balance of human readable and more compact than an array.

justend29:
Seems reasonable. If MSB is on the left, it would be enacting the same behaviour as [to_string()](https://en.cppreference.com/w/cpp/utility/bitset/to_string).

WillisMedwell:
I think a string is a good approach, as you could also add a decorator (suffix/prefix) if you wanted too as well.

stephenberry:
Yeah, matching [to_string()](https://en.cppreference.com/w/cpp/utility/bitset/to_string) behavior like `"10101100"` would be great. We could allow `"0b10101100"`, but I'm a bit cautious about this, because any time you allow multiple formats, it becomes more confusing to document and is slower if this is a runtime option. A user could easily add their own custom serializer/de-serializer to include `0b`. But, I would just have Glaze read/write as `"10101100"` for now.

WillisMedwell:
Yeah on second thought, thats a good call

justend29:
@stephenberry Supporting bitset for binary formats seems like a no-brainer. Do bitsets map to BEVE?

stephenberry:
Bitsets do map to BEVE. From the spec:
```
Boolean arrays are stored using single bits for booleans and packed to the nearest byte.
```

stephenberry:
Support has been added for JSON and BEVE read/write of std::bitset as of #526

PR #526, opened 2023-11-16
std::bitset support

stephenberry:


stephenberry:
Adds support for #508

PR #530, opened 2023-11-19
msvc iterator fix

huangminghuang:
The PR fixes the problem in MSVC where the string_view::iterator is not convertible to pointer directly.


stephenberry:
Thanks for this!

PR #532, opened 2023-11-21
Full blown reflection for aggregate structs in clang

stephenberry:
This uses clang's __builtin_dump_struct to enable fully featured (no macros, no user code) reflection for aggregate initializable structs. Only works on clang, no gcc or msvc.


stephenberry:
@justend29, @jbbjarnason 

I'm not sure if you guys use clang, but check out this black magic: [tests/reflection_test/reflection_test.cpp](https://github.com/stephenberry/glaze/blob/clang/tests/reflection_test/reflection_test.cpp)

This branch enables real reflection, no macros, no user code, for clang. If a `glz::meta` is provided, that will override this behavior, but otherwise you can just read/write structs to/from JSON without a `glz::meta`.

What are your thoughts on using `__builtin_dump_struct` in this way? I want glaze to play nice with future C++ reflection, so this was a good test case for that. The beauty of this approach is that we still get nearly the same compile time optimization using this reflection. The only overhead is needing to copy pointers due to the loss of member variable pointers from the structured binding transform.

stephenberry:
This works in Xcode 15 and later versions of clang, but I'm seeing if I can get it to work with Xcode 14.2, which is used for these actions.

jbbjarnason:
This is awesome, I really hope that the c++ committee makes this as a standard, that is reflection of members. 

> What are your thoughts on using __builtin_dump_struct in this way?

Looking at the docs: https://clang.llvm.org/docs/LanguageExtensions.html#builtin-dump-struct it is said to be for debugging purposes, so I would suggest to mention it in docs.

Can we put an `#if` for the feature flag `__has_builtin(__builtin_dump_struct)` ?


Would it be possible to implement to_tuple with tuple_cat ?

stephenberry:
Yeah, it looks like we'll be getting real reflection in C++26 (see [Herb Sutter's blog](https://herbsutter.com)), but in the meantime this could be nice and we could invisibly replace this behavior with C++26 reflection when available.

> Can we put an #if for the feature flag __has_builtin(__builtin_dump_struct) in reflect.hpp ?

Great idea!

> Would it be possible to implement to_tuple with tuple_cat?

No, `to_tuple` cannot be built with tuple_cat because we can't convert a struct directly to a tuple, instead we need to use structured bindings. But, this isn't a big problem, we just write out a large enough `to_tuple` function.

justend29:
This seems great, guys. Especially since it can be "invisibly replaced with C++26 reflection." I've been following the standards committee on the reflection proposals, and it all looks promising. I'm not sure I'll have time to experiment with this clang-specific addition in the next couple days, so I'll have to leave it to you two.  Great find.

PR #535, opened 2023-11-22
Dead code cleanup and simplifications

stephenberry:


stephenberry:
- Moves `glz::tuplet::get` to `glz::get`, simplifying code.
- Simplifies code that didn't build with MSVC 2019, as we now only support MSVC 2022

PR #537, opened 2023-11-22
removed fixed_string.hpp and using string_literal

stephenberry:


stephenberry:
@jbbjarnason
I'm cleaning up the codebase and realized we had both `string_literal` and `basic_fixed_string` classes that did the same thing. I've replaced the `basic_fixed_string` with `string_literal` and deleted `basic_fixed_string.hpp`. The `string_literal` is slightly simpler because it is not templated on the `char_type`, but I think this is fine because everywhere we are using it we are using C++ char array literals.

Let me know if you think there are any downsides to this cleanup or if it will affect your code.

jbbjarnason:
Great cleanup!

jbbjarnason (approved):

PR #557, opened 2023-11-29
expected include fix for gcc and other minor cleanup

stephenberry:


include/glaze/util/parse.hpp
```diff
@@ -124,22 +124,22 @@ namespace glz::detail
       }
    }
 
-   GLZ_ALWAYS_INLINE auto has_zero(const uint64_t chunk) noexcept
+   GLZ_ALWAYS_INLINE constexpr auto has_zero(const uint64_t chunk) noexcept
```
justend29, line 127: @stephenberry 
just for reference constexpr implies inline on function declarations. Template functions are always inline, too, including those created with `auto`.

stephenberry: Yeah, I was adding the constexpr specifier. But, I still keep GLZ_ALWAYS_INLINE, because this does force inlining on clang, so the compiler can't decide to not inline it, although in these small function cases I don't think it's necessary.

justend29: Sounds great!

PR #558, opened 2023-11-30
Int reading experiments

stephenberry:


stephenberry:
This is slower for small integers, which I do not like.

stephenberry:
Experimented using `std::from_chars` for fast paths without '.', 'e', or 'E', but it is slower.

PR #560, opened 2023-11-30
Adding short circuit error check

stephenberry:


stephenberry:
Not necessary because skip_ws checks for errors.

Issue #469, opened 2023-10-28
Evaluate viability of reducing input buffer requirements


justend29:
## Existing input buffers

The input parameter to every Glaze function that parses input from a range, as opposed to iterators, requires the input buffer to be a contiguous range ([example](https://github.com/stephenberry/glaze/blob/main/include/glaze/core/read.hpp#L49)). These functions dispatch other glaze functions that parse input from iterators, which therefore expect these iterators to lie within a single contiguous buffer, too.

The majority of operations on the input iterator would be satisfied by a multi-pass iterator, and sometimes even forward iterators.
Nevertheless, for iterators meeting only these concepts, various parsing operations would be much slower. For example, the use of `std::distance()`, including to construct `glz::parse_error`s' location index. 

## Limitations

With this premise, there seem to be two limitations:
1. The entire input buffer must be available in memory before parsing
     i. this is a regular limitation of DOM-based parsers instead of streaming parsers, like SAX parsers.
2. The entire input must be in a _contiguous_ buffer for parsing.

Now, from the outset, it may seem like a niche case to not have the entire input in a contiguous buffer.  However, it's extraordinarily common in networking to receive data in smaller chunks. In certain cases, clients aren't aware of the total response size, and read data one chunk at a time (streaming). Even when the response body is of a known size, reading the body from IO buffers is invariably accomplished through a series of smaller reads using buffer(s) of known size. The results of these reads can be either handled as they're received to reuse a single buffer, or buffered until all the smaller reads are done. The latter case is commonly buffered into a ([non-contiguous buffer](https://2122.beastdocs.prtest.cppalliance.org/libs/beast/doc/html/beast/ref/boost__beast__multi_buffer.html), since reallocating the entire buffer and copying the data is slow. asio's coins buffers of this style "buffer sequences", but this isn't exclusive to asio.

The technique of reading huge/unsized blobs one buffer/page at a time is a technique that extends to any IO operation, including reading files. With files, there are more utilities for the operating system to abstract the paging behind system calls, but that disappears with asynchronous IO.

## Consideration

So, this ticket is simply to evaluate the viability of reducing Glaze's requirements on input buffers. Maybe certain pieces of the library may need to be changed, possibly changing `glz::parse_error::location` to an iterator, but maybe not.

[...]

PR #564, opened 2023-12-01
Using memmove in dtoa

stephenberry:


justend29:
@stephenberry from your experiences, how has the performance of direct `mem*` calls compared to standard algorithms for Glaze? The algorithms intend to use the best implementation possible given the iterator types. How do they compare for contiguous chars, which would hopefully be replaced with a `mem*` call internally, although it isn't mandated?

I'm asking because of our efforts towards non-contiguous buffers geared towards I/O. If the standard library algorithm implementations really do use the best implementation, and they support non-contiguous iterators, they'd go a long way in offering that support.

justend29:
@stephenberry Fantastic answer, thank you. This makes sense.
Yes, I saw your previous post on Issue #469, but was curious if the dispatching nature of the standard containers could alleviate the need to have chunked buffers or could supplement it. I'm totally on board.

PR #572, opened 2023-12-03
Using PROJECT_IS_TOP_LEVEL to control compiler and linker optimization flags (MSVC)

taita2104:
When building on MSVC, adding some optimization flags _tout-court_ may cause some to clash with others already set previously.

In this case the `/GL` option was conflicting with `/ZI` in Debug builds, causing compilation errors and the need to manually disable the flag manually, which is clearly a workaround.

With this PR I added a CMake switch allowing the user to choose whether to activate the optimization flags or not.
Since this is an `INTERFACE` library, all options are then exposed to all the projects using this one and it seemed to me bad practice to make everybody inherit these properties by default; therefore I defaulted it to `OFF`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ include(cmake/variables.cmake)
 add_library(glaze_glaze INTERFACE)
 add_library(glaze::glaze ALIAS glaze_glaze)
 
+
 if (MSVC)
   string(REGEX MATCH "\/cl(.exe)?$" matched_cl ${CMAKE_CXX_COMPILER})
   if (matched_cl)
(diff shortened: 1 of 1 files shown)


justend29:
The existing compiler flags should be enclosed in `if(PROJECT_IS_TOP_LEVEL)` because no library should introduce compiler flags into the consumer's build. Compiler flags are not (usually) necessary to use any library, and are ideally for the builder to control.

This seems like the root of the problem which you are encountering. Would adding this fix resolve your issues? 
Can you describe how you are consuming glaze? It must be with some derivative of `add_subdirectory`, eh?

justend29:
@taita2104 
Oh, fantastic. However, it requires CMake 3.21 as the minimum version, so we'll have to add that to the `cmake_minimum_required` call. I think the CI runners use a recent enough version of CMake for that.

As for the required MSVC optimization flags, do you have any suggestions to make introducing them more robust? We wouldn't want them to conflict with consumer flags. I'm not personally familiar with MSVC, so I'm not sure if there are other flags that are mutually exclusive with these. For example, should we check for existing flags and warn when there's a conflict?

justend29:
@stephenberry Do you have concerns about raising the minimum CMake version to 3.21? Any important environments stuck on 3.18?

justend29:
@taita2104  One last question, sorry. Is `/GL`  required on MSVC for a Debug configuration, too, or is this exclusive to a Release build?

justend29:
@taita2104 Right. So, should we also wrap the release-only flags in generator expressions to automatically omit them from debug builds? Something like `$<$<NOT:$<CONFIG:Debug>>:/GL>` to grab `Release`, `RelWithDebugInfo`, etc.?

taita2104:
@justend29 sounds good

stephenberry:
> @stephenberry Do you have concerns about raising the minimum CMake version to 3.21? Any important environments stuck on 3.18?

I don't have a concern with raising the minimum CMake to 3.21

taita2104:
@justend29 @stephenberry anything left in order to finalize this one?

justend29:
> @justend29 @stephenberry anything left in order to finalize this one?

Do you think you'd have a chance to add those generator expressions to avoid optimizing flags in debug builds? I don't have ready access to MSVC.

taita2104:
@justend29 done. Optimizations are now applied only when compiling top-level and in `Release` on `MinSizeRel` configurations.

justend29:
@taita2104 Thanks so much for catching this and for going through the process of resolving it! All the best.

PR #575, opened 2023-12-04
using a char table for writing strings

stephenberry:


justend29:
I'd be interested to see benchmarks for this change.
It's essentially a battle between an explicit lookup table and a (likely) lookup table for the switch case.

justend29:
Rock. On.

PR #577, opened 2023-12-05
Binary RPC REPE

stephenberry:
This is implementation of the new [REPE](https://github.com/stephenberry/repe) binary RPC specification using BEVE


stephenberry:
@jbbjarnason, I added a unit test that doesn't build yet, but I'll keep adding to it to describe the kind of interface we want.

I'm thinking of adding custom data as a `std::vector<std::byte>` to the header. This will require a double parse for the header, but I think this makes sense. The double parse allows users to parse the header knowing exactly what it should look like every time. And, it makes it more efficient to skip.

The message itself is still templated on the body, so you can decode/encode straight to memory.

include/glaze/binary/rpc.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include "glaze/binary/write.hpp"
+
+namespace glz::repe
+{
+   constexpr uint8_t delimiter = 0b00000'110;
+   
+   struct header final
+   {
+      uint8_t version = 0b00000'110; // the REPE version
+      bool error = false; // whether an error has occurred
```
jbbjarnason, line 15: Should the `error` be something else than bool?
How does this member relate to the `error` struct below

tests/binary_rpc/binary_rpc.cpp
```diff
@@ -0,0 +1,48 @@
...
+      // This syntax works, but we can wrap this into a cleaner interface. Let this be the low level syntax.
+      my_request_data params{55, "hello"};
+      glz::repe::message<my_request_data> msg{glz::repe::header{.method = "func"}, params};
+      
+      std::string buffer;
+      glz::write_binary(msg, buffer);
+      
+      params = {};
+      msg.header = {};
+      expect(!glz::read_binary(msg, buffer));
```
jbbjarnason, line 37: Can read_binary return an error with `partial` as an error code? In terms of networking, so we would know to `async_receive` more data into the buffer before trying again. 

Secondly would it be beneficial to be able to only read the header?
```cpp
glz::repe::header header{};
glz::read_binary(header, buffer);
```

jbbjarnason (commented):
looking good

stephenberry:
The new REPE specification was implemented in other pull requests, so this one is being closed. Much progress and improvements have been made on the repe::registry.

PR #579, opened 2023-12-05
Pure compile time reflection for GCC and MSVC and more powerful reflection for Clang

stephenberry:
Adds pure reflection for GCC and MSVC, and makes the Clang reflection more powerful. This reflection does not require constexpr types.

GCC 10 must be deprecated, and I think it is time to.
MSVC doesn't currently support classes like std::optional


stephenberry:
Note that this approach should almost work for clang, but it seems Apple clang is more restrictive with the extern template solution. So, we will continue using __builtin_dump_struct for clang

stephenberry:
Seems to be an active bug for gcc: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66918

PR #580, opened 2023-12-05
Fix SEGFAULT introduced by writing empty strings.

justend29:
Additionally,
 - updated the README for the latest opts members.
 - removed the unnecessary nesting of `if constexpr (char_t<T>)` in the string writing function for JSON.
 - lots of tests for empty strings. These previously reproduced the segfaulting behavior w/o these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ out/
 **/bin/
 **/build*/
 **/lib*/
+**/*-build*/
 
 # CMake spesific settings
 CMakeUserPresets.json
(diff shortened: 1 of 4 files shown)


justend29:
is MSVC even a real compiler?

justend29:
I'm merging this w/o a review since it fixes a serious error. Should there be changes requested, I'm happy to introduce them later.

stephenberry:
@justend29, fantastic! Thanks so much for fixing this. I just saw warning for it and was about to look into it.

PR #582, opened 2023-12-05
detail reflection via reflection header

justend29:
I avoid "convenience headers" because it bloats compile times.
So, when trying the new pure reflection for GCC, I couldn't find out why my aggregate wasn't reflecting. It was the lack of `#include <glaze/reflection/reflect.hpp>` in my test.

This simply documents the header that provides these new utilities.

Great work!

### DIFF
--- a/docs/macros-for-structs.md
+++ b/docs/macros-for-structs.md
@@ -1,6 +1,8 @@
 # Struct Registration Macros
 
-Glaze supports reflection for aggregate initializable structs, and for member object pointers. So, these macros no longer save typing and have much less flexibility. They will be deprecated in the future.
+Glaze supports reflection for aggregate initializable structs, and for member object pointers. So,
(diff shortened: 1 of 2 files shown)


stephenberry:
Excellent!

stephenberry:
@justend29 I don't think the reflection code adds too much bloat. So, I'm going to make the various json and binary read/write files include the reflection, and then we can implement reflection read in the appropriate read file and write in the appropriate write file. This should reduce bloat when only reading or writing. And, I need these dependencies for binary reflection as well.

stephenberry:
The need to add reflect = true should be rare. It should only occur when using a concept like is_specialization_v, where the user type is templated and can take any argument. Types that satisfy other concepts like maps or arrays won't need it.

Also, I'm thinking the name should be changed to glaze_reflect, so that it doesn't conflict with other's code.

PR #590, opened 2023-12-06
Simplified object key parsing and fixed issue with key parsing for variants

justend29:
A failing test case was added. 
The issue stems from the fact that the two overloads of `parse_object_key` have different behaviour for the closing quote. 
I would say that it's a mistake to have this changing behviour, as all callers of `parse_objet_key` then need to offer subsequent code to match both overloads.

@stephenberry  looking into solutions now

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -10,7 +10,6 @@
 #include <iterator>
 #include <locale>
 #include <ranges>
-#include <sstream>
 #include <type_traits>
 
 #include "glaze/core/common.hpp"
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -1507,9 +1507,9 @@ namespace glz
       }
 
       // Key parsing for meta objects or variants of meta objects.
+      // We do not check for an ending quote, we simply parse up to the quote
       // TODO We could expand this to compiletime known strings in general like enums
       template <class T, auto Opts, string_literal tag = "">
-         requires(Opts.error_on_unknown_keys)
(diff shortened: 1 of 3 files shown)


stephenberry:
I agree that this should be simplified. The reason for the two approaches is for performance with error_on_unknown_keys.

justend29:
The variant parsing seems to simply not account for the second overload, as it expects the closing quote to be consumed. 
Can you detail how not consuming the final quote helps performance when `error_on_unknonw_keys = false`?

jbbjarnason:
fyi, I am also having issues with enums within variants, I think its separate issue and is a regression from my constexpr PR (https://github.com/stephenberry/glaze/pull/512). I will open a PR with a fix shortly.

stephenberry:
> The variant parsing seems to simply not account for the second overload, as it expects the closing quote to be consumed. Can you detail how not consuming the final quote helps performance when `error_on_unknonw_keys = false`?

We want to be able to look at the length of keys in a struct and use that to optimize the key parsing. If error_on_unknown_keys is false then we cannot error when we do not find a quote at the end because we don't automatically parse escapes if the struct's keys don't have escapes. For the error_on_unknown_keys = false case we need to rewind the buffer and parse the key again, taking into account that there may be escapes in the key. The object key reader may have skipped the quote on the unknown key.

This is a bit more expensive for parsing unknown keys, but much faster for parsing known keys, and that is the case we are trying to optimize for.

stephenberry:
@justend29 I'm going to push some changes to this pull request for your review.

stephenberry:
@justend29 All tests are now passing and this reduced a ton of duplicate code. Thanks for bringing this up.

PR #591, opened 2023-12-06
fix read of enum(string) within variants

jbbjarnason:
regression from this PR https://github.com/stephenberry/glaze/pull/512


stephenberry (approved):
This looks good to me, thanks!

Issue #500, opened 2023-11-04
Compile error bugs with `glz::custom`


sam20908:
1: Reduced repro:
```cpp
struct b {
  void read(const std::string&) {}
  std::vector<std::string> write() { return {}; }
  struct glaze {
    static constexpr auto value = glz::custom<&b::read, &b::write>;
  };
};
struct a {
  std::vector<b> b_;
  struct glaze {
    static constexpr auto value = &a::b_;
  };
};
void f() {
  a a_{};
  std::string buf;
  (void)glz::write_json(a_, buf);
}
```

2: Reduced repro:
```cpp
struct b {
  struct glaze {
    static constexpr auto value = glz::custom<[](auto &obj, const std::string &) {}, [](auto &obj) { return 0; }>;
  };
};
struct a {
  std::vector<b> b_;
  struct glaze {
    static constexpr auto value = &a::b_;
  };
};
void f() {
  a a_{};
  std::string buf;
  (void)glz::read_json(a_, std::string{});
}
```

sam20908:
> the lambda also is expected to return a reference of the thing to be read or written.

I assumed the `glz::custom` worked the same way in the past, where it would take a `std::string` argument for the string to be parsed. Why did the behavior change?

stephenberry:
- Thanks, I will do some testing with MSVC when I get a chance.
- Let me think over lambda handling for `glz::custom` a bit more. I'll get back to you on this.

sam20908:
If it helps, I've taken a "modified scenario" from my commercial project (Yes I'm using Glaze commercially and have credited it, great library!)
```cpp
template <> struct glz::meta<trie_node> { // trie tree
  using T = some_struct;
  static constexpr auto value = custom< //
      [](auto &obj, const std::string &data) {
        std::vector<std::string> adj_names;
        if (glz::read_json(adj_names, data)) {
          return;
        }
        obj.adj.reserve(override_names.size());
        for (const auto &override_name : override_names) {
          obj.adj.push_back(new trie_node{.name = override_name});
        }
      },
      [](auto &obj) {
        std::vector<std::string> adj_names;
        override_names.reserve(obj.adj.size());
        for (const auto &adj_name : obj.adj) {
          adj_names.emplace_back(adj_name);
        }
        return adj_names;
      }>; //
};
```

As you can see, it is almost the same as having member functions to do the job. It takes in an `obj` parameter, so using something like `std::invoke` would resolve the call in all cases when passing in the object.

[...]

PR #595, opened 2023-12-07
Supporting buffer input for custom and lambdas

stephenberry:


stephenberry:
Provides support for #500 -> so long as the lamba accepts a `const std::string_view` or `const std::string&`

PR #605, opened 2023-12-08
add gcc 13 to workflows, is it this easy?

jbbjarnason:


stephenberry:
@jbbjarnason, thanks for adding gcc 13 to testing, super useful!

PR #608, opened 2023-12-08
Support equality comparison on opts

justend29:
The compelling use case is where serialized JSON is already available, as generated with a given set of opts, and it's to be determined whether the available JSON can be reused or should be regenerated.


stephenberry:
Looks good to merge

PR #609, opened 2023-12-09
Member function const support

lamweilun:
Found an issue where custom read/write was not accepting a `const` getter member function. Added more specialization that takes in `const` member function in `type_traits.hpp`
Eg.
```C++
class Foo
{
public:
  int get_i() const { return i; }  // This only works if it's non-const
  void set_i(int I) { i = I; }

private:
  int i = 0;
};

template <>
struct glz::meta<Foo>
{
  static constexpr auto value = object("i", custom<&Foo::set_i, &Foo::get_i>);
};

// The above snippet would not compile in the latest release
```

### DIFF
--- a/include/glaze/util/type_traits.hpp
+++ b/include/glaze/util/type_traits.hpp
@@ -67,11 +67,23 @@ namespace glz
       using type = Result (*)(void*, std::decay_t<Args>&&...);
    };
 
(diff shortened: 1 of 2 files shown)


stephenberry:
@lamweilun
Excellent! Thanks for this fix! I added a unit test as well.

PR #637, opened 2023-12-18
Fix writing maps with internally skipped null members

justend29:
Previously, invalid JSON was serialized for `writable_map`s when "internal" nulls were skipped. 
Commas were missed for every entry following a skipped entry, producing JSON like `{"hello":4"there":"bob"}`.

The logic previously wrote entry separators when the previous entry was **not** skipped. However, as soon as the first entry is written, an entry separator can be written in front of every subsequent entry. The serialization only needs to avoid these separators while it's starting the object - i.e skipping **leading** entries - meaning only a `{` has been serialized. After that, no matter how many internal entries are skipped, there's always a serialized entry that needs separation.

### DIFF
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -542,21 +542,18 @@ namespace glz
                };
 
                auto it = std::begin(value);
-               [[maybe_unused]] bool previous_skipped = write_first_entry(it);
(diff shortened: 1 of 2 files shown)


justend29:
@stephenberry The MSVC job is hanging again. Should we just wait or do you have some insight about resolving that?

PR #640, opened 2023-12-19
Cleanup of string reading

stephenberry:


stephenberry:
This optimization results in slower parsing of short strings. It is better for very long strings and single reads, but I'm unsure of whether an intermediate buffer should be used.

PR #669, opened 2023-12-30
include<expected> before feature test macro

justend29:
The `__cpp_lib_expected` feature testing macro is defined in the <expected> header, so the header is first required.


stephenberry:
Ah, thanks for this fix. I had changed the code because I keep getting annoyed by MSVC warnings for the header.

PR #700, opened 2024-01-09
multi_includer test

stephenberry:


stephenberry:
This test really isn't needed. The problem was stemming from BEVE not handling hostname_include

Issue #695, opened 2024-01-09
call to consteval function 'glz::get_name<&ProjectSettings::identity>' is not a constant expression


Wolfleader101:
Hello,

When using clang on windows I get the following error
`call to consteval function 'glz::get_name<&ProjectSettings::identity>' is not a constant expression`

![image](https://github.com/stephenberry/glaze/assets/25214835/764edf0c-a170-4a73-afff-e243992fede6)

![image](https://github.com/stephenberry/glaze/assets/25214835/607a3312-7fda-4871-8f4b-79dccbfb2270)

![image](https://github.com/stephenberry/glaze/assets/25214835/329e46a8-6aac-41a4-b0e1-948d24c37332)

![image](https://github.com/stephenberry/glaze/assets/25214835/3e0ff9e9-2241-4f52-b3c9-12516e63189b)

Usage:
```cpp
struct ProjectSettings
{
    std::string identity = "";
    std::string name = "";
    std::string version = "0.1.0";
    std::string description = "";
    std::string author = "";
    std::string assetPath = "";
};

    ProjectSettings projectSettings{};
    auto ec = glz::read_file_json(projectSettings, "assets/demoApp.json", std::string{});
```

stephenberry:
Thanks for reporting this. I don't currently have a good way of testing clang on windows. I have a few questions:

Notice the `#if defined(_MSC_VER)`, should this be defined for clang?

What versions of clang for windows are you using?

Wolfleader101:
> Thanks for reporting this. I don't currently have a good way of testing clang on windows. I have a few questions:
> 
> Notice the `#if defined(_MSC_VER)`, should this be defined for clang?
> 
> What versions of clang for windows are you using?

Hey,

I tried playing around with it yesterday, and I think I found the problem.
I thought at first maybe my clang install was busted as `_MSC_VER` was being defined, but  `__clang__` is also defined, and after some googling I found that its due to clang trying to stay compatible with MSVC compiled code.

My Clang Version for reference.

![image](https://github.com/stephenberry/glaze/assets/25214835/412a39fc-5ddc-4abc-8252-5cc309a3678c)

I believe the appropriate fix is to have:
`#if defined(_MSC_VER) && !defined(__clang__)`

This is linked to PR #701

stephenberry:
Thanks for solving this. I merged your pull request.

PR #701, opened 2024-01-10
patch for clang-cl bug

Wolfleader101:
A patch for the issue #695 

Clang on Windows defines `_MSC_VER` to be compatible with `MSVC`, however, `get_name_msvc` can only be called when compiling with `MSVC`.


stephenberry:
Thanks for this fix!

PR #706, opened 2024-01-10
Support member object reflection with approx.hpp and compare.hpp

stephenberry:


stephenberry:
This uses old, less flexible code before the `glz::reflect` refactor. Therefore closing this pull request and new code could be used to implement these features.

PR #708, opened 2024-01-11
Cleanup of jsonrpc

stephenberry:
I'm working on a significant cleanup of the JSON RPC code. The aim is to make the API simpler and more straightforward for users.

- Using reflection for tests, examples, and member object reflection for some internals

- Hiding `expected` from the `server.on` registration. The user can just return their result type or an `rpc::error` and the internals will wrap that call to produce an `expected`

### DIFF
--- a/docs/json-rpc.md
+++ b/docs/json-rpc.md
@@ -44,15 +44,15 @@ int main() {
        client;
    
     // One long living callback per method for the server
-    server.on<"foo">([](foo_params const& params) -> glz::expected<foo_result, rpc::error> {
+    server.on<"foo">([](foo_params const& params) {
         // access to member variables for the request `foo`
         // params.foo_a 
         // params.foo_b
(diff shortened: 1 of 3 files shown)


stephenberry:
@jbbjarnason, If you could look over these changes and let me know if they are good to go, that would be great! I think I'm going to stop here for this merge request.

I currently have an internal JSON RPC that is much simpler and faster than this implementation, which my company uses in the field. But, it doesn't conform to JSON RPC 2.0

I'm trying to simplify this implementation to see if there is a way to synthesize the two approaches. But, I will probably end up initially adding another JSON RPC approach to Glaze and work on synthesizing as we develop them further.

Also, I will be working more on binary RPC with BEVE, but that will probably happen later this spring.

jbbjarnason (commented):
Looks good !

PR #709, opened 2024-01-12
fix reading empty struct with reflection

huangminghuang:
The PR fixes the compile error to read an empty struct without explicit meta object.


stephenberry:
Fantastic! I had missed this logic error.

PR #713, opened 2024-01-18
Add minify method to complement prettify method

BowDown097:
Was surprised to see a prettify method here but no minify method. Needed one for a program I'm writing that decrypts and encrypts JSON data so I wrote one and figured I'd contribute it back.

### DIFF
--- a/include/glaze/json/prettify.hpp
+++ b/include/glaze/json/prettify.hpp
@@ -11,6 +11,58 @@ namespace glz
    {
       enum class general_state : uint32_t { NORMAL, ESCAPED, STRING, BEFORE_ASTERISK, COMMENT, BEFORE_FSLASH };
 
+      inline void handle_other_states(const char c, general_state& state) noexcept
+      {
+         switch (state) {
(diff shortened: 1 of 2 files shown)


BowDown097:
While I'm at it, I had some questions using this library:
1. Why is 3 the default indent for the prettify method? I have never seen a JSON file with a 3 indent, and pretty much every formatting implementation I've ever seen has 4 as the default, if there is one at all.
2. Should the header file be renamed to something like formatting.hpp now that minifying has been added to it? I didn't want to just do it because it would be a breaking change, but it could be a good idea.

stephenberry:
Awesome, thanks so much for the contribution! To answer your questions:

Three spaces is just what the company I first developed the library at used. But, I do think it should probably be four to be more standardized.

I’ll probably put minify into its own header and have prettify.hpp and minify.hpp with some more core helper files for each. And then we could have formatting.hpp that includes both. But, this can be done at a later date. My intention is to create highly optimized versions of these functions in the future, but this is fine for the first step.

I will make note these changes to implement in the future.

PR #717, opened 2024-01-29
Added opts to exception based json file write.

isaki:
This change mirrors the read version.

### DIFF
--- a/include/glaze/exceptions/json_exceptions.hpp
+++ b/include/glaze/exceptions/json_exceptions.hpp
@@ -40,7 +40,7 @@ namespace glz::ex
    template <auto Opts = opts{}, class T>
    void read_file_json(T& value, const sv file_name, auto&& buffer)
    {
-      const auto ec = glz::read_file_json(value, file_name, buffer);
+      const auto ec = glz::read_file_json<Opts, T>(value, file_name, buffer);
       if (ec == glz::error_code::file_open_failure) {
          throw std::runtime_error("file failed to open: " + std::string(file_name));
       }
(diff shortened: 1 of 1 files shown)


include/glaze/exceptions/json_exceptions.hpp
```diff
@@ -107,7 +107,7 @@ namespace glz::ex
       return glz::write_jsonc(std::forward<T>(value));
    }
 
-   template <class T>
+   template <auto Opts = opts{}, class T>
    void write_file_json(T&& value, const std::string& file_name, auto&& buffer)
    {
       const auto ec = glz::write_file_json(std::forward<T>(value), file_name, buffer);
```
jbbjarnason, line 113:
```suggestion
      const auto ec = glz::write_file_json<Opts>(std::forward<T>(value), file_name, buffer);
```

I guess you would need to propagate the template parameter down to the actual write right?

isaki: This isn't necessary for the read version, which seems surprising (which is basically what I copied). It did build and test successfully locally (gcc, linux) as well, though I did not yet get a chance to test if it works in the code base I needed this for.

```
   template <auto Opts = opts{}, class T>
   void read_file_json(T& value, const sv file_name, auto&& buffer)
   {
      const auto ec = glz::read_file_json(value, file_name, buffer);
      if (ec == glz::error_code::file_open_failure) {
         throw std::runtime_error("file failed to open: " + std::string(file_name));
      }
      else if (ec) {
         throw std::runtime_error("read_file_json error: " + glz::format_error(ec, buffer));
      }
   }
```

isaki: I've gone and added the template arguments in all the places I mentioned (and an additional function I noticed).

stephenberry:
Thanks for this improvement!

Issue #715, opened 2024-01-26
"ThreadSanitizer: data race" when using reflection


Oipo:
Using v2.0.6 with multiple threads reading/writing json at the same time as well as reflection, leads to thread sanitizer giving data races. Without thread sanitizer, this segfaults on my machine.

[glaze_test.cpp.txt](https://github.com/stephenberry/glaze/files/14061842/glaze_test.cpp.txt)
[output.txt](https://github.com/stephenberry/glaze/files/14061839/output.txt)

Adding the following completely removes these issues:

```c++
template <>
struct glz::meta<TestMsg> {
    using T = TestMsg;
    static constexpr auto value = object(
            "id", &T::id,
            "val", &T::val
    );
};
```

stephenberry:
Thanks for giving an example and pointing this out.

stephenberry:
So, the issue was that I misunderstood `constinit` and thought it was implicitly thread_local when static, which it is not. I have an easy fix, which just adds thread_local to these `constinit` values. However, MSVC has a bug and won't compile. I'm going to try to find a workaround, but if I can't, MSVC just won't support multithreading and reflection. I'll merge this fix soon and push a version bump.

stephenberry:
This has been resolved for Clang and GCC with the merge of #727. I added a new issue to file an issue with MSVC for them to fix their compiler issue with `static thread_local constinit`

stephenberry:
This has been fixed for MSVC as well as of #730. We just avoid using constinit for MSVC, which will likely result in slower runtime performance (from previous testing), but is necessary until MSVC fixes the associated compiler bug.

Oipo:
Super, thanks. Do you have a link to the bug report for MSVC? I'll upvote it.

PR #727, opened 2024-02-08
Multi-threading fix for reflection for GCC and Clang

stephenberry:
Using thread_local for tuple_of_ptrs


stephenberry:
Fixes #715

stephenberry:
This only fixes multi-threading with reflection for GCC and Clang. MSVC is broken when it comes to `static thread_local constinit`

PR #771, opened 2024-02-20
issue 768

arturbac:
Changed a bit proposed earlier code to fix issue with buffer overflow on testing for comma on last csv record at the end of line as csv is not required to have lf at the end.


stephenberry:
Thank you so much for this pull request and for including unit tests as well. Much appreciated!

arturbac:
not using unit tests is a crime ;-)

PR #773, opened 2024-02-20
fix buffer overflow:

arturbac:
With this fix all tests are green with clang asan
10: =================================================================
10: ==28606==ERROR: AddressSanitizer: heap-buffer-overflow
10: WRITE of size 8 at 0x5080000001f9 thread T0
10:     #1 0x55be8ea855e0 in char const* glz::detail::parse_string<8ul,[..]include/glaze/util/parse.hpp:874:10
10:     #2 0x55be8ea83aa5 in void glz::detail::from_json<[..]/include/glaze/json/read.hpp:580:29

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -572,7 +572,7 @@ namespace glz
                   if (escaped) {
                      static constexpr auto Bytes = 8;
 
-                     const auto length = round_up_to_multiple<Bytes>(size_t(it - start));
+                     const auto length = round_up_to_multiple<Bytes>(size_t(it - start))+Bytes;
                      value.resize(length);
 
                      const char* c;


arturbac:
I would like to suggest adding profile with asan build
```
                "CMAKE_C_COMPILER": "clang-17",
                "CMAKE_CXX_COMPILER": "clang++-17",
                "CMAKE_CXX_FLAGS" : "-ggdb -fvisibility=default -fno-omit-frame-pointer -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=undefined",
                "CMAKE_EXE_LINKER_FLAGS" : "-fsanitize=address -fsanitize=undefined"
```
And running it always before creating release.

stephenberry:
Ah, in looking at the address sanitizer issue recently for writing I had realized this could happen but had forgotten how this read code works. Thanks so much for the fix.

The issue is that the length of the string, even when rounded up to the nearest 8 byte value, does not include the null pointer to terminate the string. Because the null pointer may live at the first byte of the 8 byte uint64_t swar value, we need 7 more bytes to memcpy safely. I would expect just adding 7 would solve the issue, but I'm happy adding an even 8 bytes and I'll merge this now.

arturbac:
>  I would expect just adding 7 would solve the issue, but I'm happy adding an even 8 bytes and I'll merge this now.
With heap allocation on linux 64 bit platform You will not get lower allocation than rounded to multiply of 8 bytes.

PR #781, opened 2024-02-21
REPE support for structs of functions

stephenberry:
This merge allows users to register structs of functions that use reflection and add the functions to the server's RPC methods.

This also updates the cli_menu to write out and read in JSON, which allows it to return any types and makes it compatible with RPC methods.

### DIFF
--- a/include/glaze/reflection/reflect.hpp
+++ b/include/glaze/reflection/reflect.hpp
@@ -7,8 +7,14 @@
 #include "glaze/reflection/get_name.hpp"
 #include "glaze/reflection/to_tuple.hpp"
 
+#include <initializer_list>
+
 namespace glz
 {
+   // If you want to make an empty struct reflectable, add this constructor:
+   // my_struct(glz::make_reflectable) {}
+   using make_reflectable = std::initializer_list<bool>;
+   
    namespace detail
    {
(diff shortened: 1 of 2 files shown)


stephenberry:
@jbbjarnason, I'm adding the ability to register structs with functions, which will translate these into RPC methods with the member variable name reflected as the method name. However, I want this to work with nested structures, so that functions can be appropriately grouped. I also plan to automatically generate read/write calls to member variables placed in these structures.

To work with nested structures it just makes sense to use JSON pointer syntax for method names. It gives us a standard way that works well with JSON to refer to the method we want to call. The only downside is that instead of just have a function called "sum", to be compliant it will now need to be "/sum". But, I don't think it is a bad price to pay for the flexibility and auto-generation that becomes possible.

You can look at the test code from this pull request to see how the nested syntax works.

My question for you is do you see any potential pitfalls with using JSON pointer syntax for RPC method names? How would this affect your RPC use cases?

stephenberry:
@jbbjarnason, I guess since JSON Pointer syntax must begin with `/`, we can still support function names that don't begin with the forward slash as just top level functions to call.

jbbjarnason:
I don't see any problems with it. And it should not become a problem for our use case

PR #799, opened 2024-02-24
Optimize glz::any to avoid using virtual

stephenberry:


stephenberry:
@arturbac, I you're able, I could use another pair of eyes looking at this. I'm reworking an implementation of `any` from here [https://github.com/kocienda/Any](https://github.com/kocienda/Any)

The reason I need a custom any is that I need a `data()` function to obtain a `void*` to the underlying data. Because this version of `any` has small object optimization, there there may or may not be additional allocations.

In `json_test.cpp` all of my tests pass except for:
```c++
"poly pointer"_test = [] {
    dog d{};
    glz::poly<animal> a{&d};

    a.call<"eat">();

    expect(d.age == 1);
    expect(&a.get<"age">() == &d.age);
 };
```

> glz::poly provides polymorphism without inheritance

I'd like to figure this out so that I can merge these changes and remove the virtual destructor. No problem if you don't have the time, but thanks for brining up this to fix.

arturbac:
Will try soon take a look at repo impl, currently I am bit busy.

arturbac:
Did You solve it ?

```
/home/artur/projects/glaze/build/tests/json_test> /home/artur/projects/glaze/build/tests/json_test/json_test
/home/artur/projects/glaze/include/glaze/util/dump.hpp:168:34: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:44:28: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/artur/projects/glaze/include/glaze/util/dump.hpp:168:34 
/home/artur/projects/glaze/include/glaze/util/dump.hpp:76:34: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/string.h:44:28: note: nonnull attribute specified here
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/artur/projects/glaze/include/glaze/util/dump.hpp:76:34 
Suite 'unnamed suite': all tests passed (6558 asserts in 328 tests)
Suite 'global': all tests passed (0 asserts in 0 tests)
```

arturbac:
As You probably know this code is invalid.
And it is difficult to understand.
any uses storage to store in buffer value of dog *
raw_ptr holds address to that storage holding value of pointer
In decltype(auto) call(Args&&... args)
there is call to arguments op reinterpreting this raw_ptr to dog instead to dog* and dereferencing it.

reinterpret_cast<T*>(ptr)->*MemPtr
Once again it this to difficult to fix for me but I see this examining pointers
dog *     0x7ffaf05da220
storage * 0x7f6a209da298 holding value of dog *
eat       0x7f6a209da298 eating on reinterpret caster storage as dog *
And one note
aligned_storage is deprecated in c++23 use for ex std::byte storage[]

stephenberry:
Thanks for this insight and the note about `aligned_storage` deprecation. I'll take another stab at this when I get the chance.

PR #807, opened 2024-03-04
Supporting references in structs with reflection

stephenberry:


stephenberry:
This doesn't seem possible due to the compiler not seeing references within the extern type being valid. We can properly count members on clang, but this also has issues on gcc for references.

schaumb:
Does the original [union](https://github.com/boostorg/pfr/issues/90#issuecomment-1597313072) method work with references?

PR #841, opened 2024-03-18
Matching glz::schema names to JSON schema specification

stephenberry:
Breaking change to `glz::schema`, to match variable names with the JSON schema specification. This also moves us towards using pure reflection in the future if we want to.


stephenberry:
@jbbjarnason, this may affect your codebase. So, I wanted to make you aware of this change.

PR #863, opened 2024-03-22
custom read/write test

stephenberry:


tests/json_test/json_test.cpp
```diff
@@ -7086,6 +7086,48 @@ suite expected_tests = [] {
...
+         value.str += "write";
+         write<json>::op<Opts>(value.str, args...);
+      }
+   };
+}
+
+suite custom_struct_tests = [] {
+   "custom_struct"_test = [] {
+      custom_struct obj{};
+      std::string s{};
```
arturbac, line 7122: try changing it to const

`std::string const s{};`

stephenberry: You can't use const for your buffer if you are writing, because it may need to grow the buffer.

arturbac: my bad, I was thinking about custom_struct obj{};
But I am busy at work and I made mistake

stephenberry: No problem

PR #864, opened 2024-03-22
Cleanup

stephenberry:
- removing static to help the compiler optimize away compile time temporaries

Added `glz::read_json_supported`, `glz::write_json_supported`, `glz::read_binary_supported`, and `glz::write_binary_supported` concepts to check if your type is compatible with Glaze.


tests/binary_test/binary_test.cpp
```diff
@@ -38,6 +38,9 @@ struct glz::meta<my_struct>
    );
 };
 
+static_assert(glz::write_binary_supported<my_struct>);
+static_assert(glz::read_binary_supported<my_struct>);
```
arturbac, line 42: Note: If this is annoying to You that I report You such things like with doc and this one, I will stop reporting anything, just say, as I am not sure You want that reports.

I think You are not testing real use case with std::declval<T>() witch is rvalue refrence T and code in reality expects lvalue reference T 

Those pass
```
static_assert(glz::read_json_supported<rc_t>);
static_assert(glz::detail::read_json_invocable<glz::opts{},rc_t ,glz::context,char const *,char const *>);
```
Those don't
```
static_assert(glz::read_json_supported<rc_t &>);
static_assert(glz::detail::read_json_invocable<glz::opts{},rc_t &,glz::context &,char const *&,char const *&>);
```

```
read.hpp:83:33: error: static assertion failed due to requirement 'false_v<Aqurat::rc_tmpl<unsigned short> &>': Glaze metadata is probably needed for your type
   83 |                   static_assert(false_v<T>, "Glaze metadata is probably needed for your type");
      |                                 ^~~~~~~~~~
glz::detail::read<10>::op<opts{..}, rc_tmpl<unsigned short> &, glz::context &, const char *&, const char *&>' requested here
```
As You see it is expecting **rc_tmpl<unsigned short> &** lvalue reference
Because I made mistake and mixed type with perfect forwarding getting as result rvalue arg

```
template<typename value_type>
struct from_json<Aqurat::rc_tmpl<value_type>>
  {
  using rc_type = Aqurat::rc_tmpl<value_type>;
  template<auto Opts>
  static void op(rc_type && rc, auto &&... args)
  
```
instead of one of 2
```
  static void op(auto && rc, auto &&... args)
  static void op(rc_type & rc, auto &&... args)
```

arturbac: read concept is actually wrong as it rejects, valid op
static void op(rc_type & rc, auto &&... args)
so only working case is having perfect forwarding auto && on op it passes concept and will work as T & during real write.

stephenberry: Go ahead and keep reporting, it is helpful. I just threw these together, and you're correct that we should test with lvalues. I can update the concepts.

arturbac: I think joining used concept in internals with user one would be better than writting from scratch
Something like this
```
template<typename T>
concept read_json_supported = glz::detail::read_json_invocable<glz::opts{}, std::add_reference_t<T>,
glz::context,char const *,char const *>;
```

PR #865, opened 2024-03-22
Using internal concepts for read/write supported

stephenberry:


stephenberry:
@arturbac, using internal concepts

include/glaze/json/write.hpp
```diff
@@ -1487,8 +1487,5 @@ namespace glz
    }
 
    template <class T>
-   concept write_json_supported = requires {
-      detail::to_json<std::remove_cvref_t<T>>::template op<opts{}>(std::declval<T>(), context{}, std::string{},
-                                                                   size_t{});
-   };
+   concept write_json_supported = detail::write_json_invocable<glz::opts{}, std::add_lvalue_reference<T>, glz::context, std::string, size_t>;
```
arturbac, line 1490: IMHO std::add_const_t<std::add_lvalue_reference_t<T>> ? 
it should accept type const & in to_json not only &

stephenberry: Agreed

Issue #874, opened 2024-03-25
missing check on to_json<T> and from_json when partial specializing with meta<>


arturbac:
I asked You on other discussion for help You answered only to first question suggesting meta<T>
So i took of the constraint if constexpr (write_json_invocable<Opts, T, Ctx, B, IX>)
to see what will fail.
And it fails because of ambiguity even when having meta<T> for partial specializations
So for write and read(similar) I think You are missing **glaze is missing this** below
What is interesting when I change to bounded enum so more constrained enum a subset of normal enum it works out of the box ..
But for normal enum You need that fixes below : (I dont need them but for glaze I think it would be nice to respect own meta<> customisations by users

```
      template <class T>
         requires(std::is_enum_v<std::decay_t<T>> && !glaze_enum_t<T> 
         && !meta<std::decay_t<T>>::custom_write // <----- glaze is missing this
         )
      struct to_json<T>
      {
         template <auto Opts, class... Args>
         GLZ_ALWAYS_INLINE static void op(auto&& value, is_context auto&& ctx, Args&&... args) noexcept
         {
            write<json>::op<Opts>(static_cast<std::underlying_type_t<std::decay_t<T>>>(value), ctx,
                                  std::forward<Args>(args)...);
         }
      };
```
Do You agree ?

I took off constraint to check here
 ```
     template <>
      struct write<json>
      {
         template <auto Opts, class T, is_context Ctx, class B, class IX>
         GLZ_ALWAYS_INLINE static void op(T&& value, Ctx&& ctx, B&& b, IX&& ix) noexcept
         {
            // if constexpr (write_json_invocable<Opts, T, Ctx, B, IX>) {
               to_json<std::remove_cvref_t<T>>::template op<Opts>(std::forward<T>(value), std::forward<Ctx>(ctx),
                                                                  std::forward<B>(b), std::forward<IX>(ix));
            // }
            // else {
               // static_assert(false_v<std::remove_cvref_t<T>>, "Glaze metadata is probably needed for your type");
            // }
         }
      };
```

and testing my galze support for single enum so I will not need to use glz::enuemrate as I will use simple_enum::enum_view for automatic enumerations for all bounded_enums at once
And simple_enum with galze can be symbiotically working and galze support would be then included in simple enum

[glaze_json_enum_name.hpp](https://github.com/arturbac/simple_enum/blob/glaze_support/include/simple_enum/glaze_json_enum_name.hpp)

[...]

PR #875, opened 2024-03-25
fix partial custom specialization on enumerations #874

arturbac:
#874


arturbac:
Wait I didnt check if t works for me ..

arturbac:
it is ok, works as expected for me and You UT are passing too.

stephenberry:
Looks good, thanks!

Issue #895, opened 2024-03-30
Generic JSON parsing wraps the parsed value in an extra array.


issuefiler:
# Bug

```
std::expected<glz::json_t, glz::parse_error> document
	{glz::read_json<glz::json_t>(input_string)};
std::string buffer{};
glz::write_json(*document, buffer);
```

when the `input_string` (a `std::string_view` with `.remove_prefix(…)` applied) is “`{"type":"A","service_set_identifier":"","passphrase":""}`”, the parsed `glz::json_t` ends up with “`[{"passphrase":"","service_set_identifier":"","type":"A"}]`”, the original value wrapped in an array.

I’ve just stumbled upon this; not sure if this is reproducible.

issuefiler:
I cannot reproduce it with a simplified code… Let me see if this is a bug on my side.

issuefiler:
# Minimal reproducible code

When you copy(-construct) a `glz::json_t`, the copy gets wrapped in an extra array.

```C++
std::string output;
glz::write_json(glz::json_t{*(glz::read_json<glz::json_t>("{}"))}, output);
// {} → [{}]
```

stephenberry:
Thanks for bringing up this issue. I added a test case for it on this pull request: #897. The weird thing is that this only occurs on GCC, and not MSVC or Clang. I'm not yet sure which compiler is correct. Maybe GCC is wrong? I think GCC is seeing this as an initializer list rather than a copy construction.

stephenberry:
I have not been able to figure out a way to fix GCC in this case. I am fairly convinced that this is a GCC bug, because it should select the constructor that takes the type directly over the initializer list version. But, GCC is instead choosing to use the initializer list constructor even though no additional braces were added to make it an initializer list.

The cross platform solution is to use parentheses `()` for constructing your `json_t`. The following will produce the desired result:
```c++
glz::write_json(glz::json_t(*(glz::read_json<glz::json_t>("{}"))), output);
```

I'm going to convert this issue into a discussion for future reference. If you wanted, you could report this bug to GCC.

PR #897, opened 2024-04-01
Added json_t copy construction test

stephenberry:
- Also adding moves in constructors to help performance


stephenberry:
Test case for #895

PR #904, opened 2024-04-02
Improved read/write supported concepts

stephenberry:
The `read_json_supported`, `write_json_supported`, and corresponding binary concepts now simply check for the existence of a template specialization for the underlying to/from functor. This simplifies the concepts, makes them more understandable, and reduces compile time. If we were to use simpler concepts that tried to invoke `op`, then we would be asking the compiler to see if it can instantiate versions of templates that we weren't even using, which is a bad idea from a compilation standpoint (both for performance and potential obscure issues in differences).

This update also adds concepts to the core read/write functions that take in configuration options. This allows us to reduce the depth of the template errors and make the output more manageable.

### DIFF
--- a/include/glaze/binary/beve_to_json.hpp
+++ b/include/glaze/binary/beve_to_json.hpp
@@ -455,14 +455,14 @@ namespace glz
                const auto complex_type = complex_header & 0b0000000'1;
                if (complex_type) {
                   // complex array
-                  const auto tag = complex_header & 0b111'00000;
+                  const auto number_tag = complex_header & 0b111'00000;
                   const auto n = int_from_compressed(it, end);
                   dump<'['>(out, ix);
                   for (size_t i = 0; i < n; ++i) {
                      dump<'['>(out, ix);
-                     beve_to_json_number<Opts>(tag, ctx, it, end, out, ix);
+                     beve_to_json_number<Opts>(number_tag, ctx, it, end, out, ix);
                      dump<','>(out, ix);
-                     beve_to_json_number<Opts>(tag, ctx, it, end, out, ix);
+                     beve_to_json_number<Opts>(number_tag, ctx, it, end, out, ix);
                      dump<']'>(out, ix);
                      if (i != n - 1) {
                         dump<','>(out, ix);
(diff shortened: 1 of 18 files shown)


stephenberry:
@arturbac, I'm tagging you here because I'm updating how `read_json_supported` and `write_json_supported` work. They are more general now, and they have been applied to the top level read/write functions as well as the core read/write functions.

They now only check for the existence of a template specialization for the underlying to/from functor. This simplifies the concepts and makes everything more straightforward.

Let me know if you expect these changes to affect your use case in a detrimental way, or if you think they look good. Thanks!

arturbac:
Checking

arturbac:
I've checked fast to json/from json/schema and rpc calls after removing from simple_enum specialization of all enums leaving only partial for bounded_enum.
IMHO looks good, names are handled by simple_enum for bounded enums and rest is handled by glaze with integrals, including rpc schema
```
enum class Color : int8_t
  {
  Green,
  Blue,
  Red = -1
  };

consteval auto adl_enum_bounds(Color)
  {
  using enum Color;
  return simple_enum::adl_info{Red, Blue};
  }
enum class ColorUnbounded: int8_t
  {
  Green,
  Blue,
  Red = -1
  };
```
```
{
  "type": [
    "object"
  ],
  "properties": {
    "color": {
      "$ref": "#/$defs/Color"
    },
    "color2": {
      "$ref": "#/$defs/ColorUnbounded"
    },
    "value": {
      "$ref": "#/$defs/int32_t"
    }
  },
  "additionalProperties": false,
  "$defs": {
    "Color": {
      "type": [
        "string"
      ],
      "oneOf": [
        {
          "const": "Red"
        },
        {
          "const": "Green"
        },
        {
          "const": "Blue"
        }
      ]
    },
    "ColorUnbounded": {
      "type": [
        "number",
        "string",
        "boolean",
        "object",
        "array",
        "null"
      ]
    },
    "int32_t": {
      "type": [
        "integer"
      ]
    }
  }
}
```

stephenberry:
Great, thanks for checking! I'll merge this in.

PR #913, opened 2024-04-12
Custom JSON RPC error support/testing

### LINKED ISSUE #912: Is that possible for custom error to pop up as rpc::error in json-rpc?
Try to return custom errors from json-rpc server methods like 
```
struct MyParams {};
struct MyResult {};
using MyServer = rpc::server<
    rpc::method<"MyMethod", MyParams, std::expected<MyResult, rpc::error>>>;
using MyClient = rpc::client<
    rpc::method<"MyMethod", MyParams, std::expected<MyResult, rpc::error>>>;

int main() {
  MyServer server;
  server.on<"MyMethod">([](MyParams) { return rpc::error{}; });

  MyClient client;
  auto [req, _] =
      client.request<"MyMethod">(1, MyParams{}, [](auto val, auto id) {});
  auto resp = server.call(req);

  std::println("resp: {}", resp);
}
```
but the response actually use the returned error as the result payload, i.e. the result look like
```
resp: {"jsonrpc":"2.0","result":{"unexpected":{"code":0,"message":"No error"}},"id":1}
```
instead of using the error as json-rpc error, like
```
resp: {"jsonrpc":"2.0","error":{"code":0,"message":"No error"},"id":1}
```

Is that possible to integrate error handling with the method handler?

stephenberry:


stephenberry:
Closing because this is stale and needs more research

PR #916, opened 2024-04-13
Implemented a new SWAR-enabled comparison function to accelerate string comparisons.

RealTimeChris:
Achieves the following numbers under Windows-x64 MSVC 2022, where the top values are using memcmp and the bottom values are using the new compare function:
2-Characters:
2800ns
2400ns
3-Characters:
3100ns
2600ns
4-Characters:
4100ns
2600ns
5-Characters:
3900ns
2600ns
6-Characters:
5000ns
2600ns
7-Characters:
5000ns
2800ns
8-Characters:
5200ns
3600ns
9-Characters:
4400ns
2800ns
10-Characters:
4300ns
3500ns
11-Characters:
4500ns
3300ns
12-Characters:
5100ns
3100ns
13-Characters:
5600ns
3300ns
14-Characters:
6200ns
3300ns
15-Characters:
6400ns
3600ns
16-Characters:
6500ns
3900ns
17-Characters:
4700ns
3100ns
31-Characters:
6400ns
5000ns
32-Characters:
7400ns
5500ns
64-Characters:
7400ns
6900ns

### DIFF
--- a/include/glaze/util/compare.hpp
+++ b/include/glaze/util/compare.hpp
@@ -1,0 +1,68 @@
+// Glaze Library
+// For the license information refer to glaze.hpp
+
+#pragma once
+
+#include <cstdint>
+
+namespace glz
+{
+   template <typename... types>
+   struct type_list;
+
+   template <typename value_type, typename... rest>
+   struct type_list<value_type, rest...>
+   {
+      using current_type = value_type;
+      using remaining_types = type_list<rest...>;
+      static constexpr uint64_t size = 1 + sizeof...(rest);
+   };
+
+   template <typename value_type>
+   struct type_list<value_type>
+   {
+      using current_type = value_type;
+      static constexpr uint64_t size = 1;
+   };
+
+   template <typename type_list, uint64_t Index>
+   struct get_type_at_index;
+
+   template <typename value_type, typename... rest>
+   struct get_type_at_index<type_list<value_type, rest...>, 0>
+   {
+      using type = value_type;
+   };
+
+   template <typename value_type, typename... rest, uint64_t Index>
+   struct get_type_at_index<type_list<value_type, rest...>, Index>
+   {
+      using type = typename get_type_at_index<type_list<rest...>, Index - 1>::type;
+   };
+
+   using integer_list = type_list<uint64_t, uint32_t, uint16_t, uint8_t>;
+
+   template <uint64_t index = 0, typename char_type01, typename char_type02>
+   GLZ_ALWAYS_INLINE bool compare(char_type01* string1, char_type02* string2, uint64_t lengthNew)
+   {
+      using integer_type = typename get_type_at_index<integer_list, index>::type;
+      static constexpr uint64_t size{sizeof(integer_type)};
+      integer_type value01[2]{};
+      while (lengthNew >= size) {
+         std::memcpy(value01, string1, sizeof(integer_type));
+         std::memcpy(value01 + 1, string2, sizeof(integer_type));
+         lengthNew -= size;
+         string1 += size;
+         string2 += size;
+         if (value01[0] != value01[1]) {
+            return false;
+         }
+      }
+      if constexpr (index < integer_list::size - 1) {
+         return compare<index + 1>(string1, string2, lengthNew);
+      }
+      else {
+         return true;
+      }
+   }
+}
--- a/include/glaze/util/hash_map.hpp
+++ b/include/glaze/util/hash_map.hpp
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <glaze/util/compare.hpp>
 #include <algorithm>
 #include <array>
 #include <bit>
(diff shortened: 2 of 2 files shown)


stephenberry:
@RealTimeChris, so I was able to improve on your code and make a version that is consistently faster for both short and longer strings. However, in benchmarking, I found that my new version beat std::memcmp for short strings, but std::memcmp still wins out for longs strings (probably because it can work on 16 byte chunks). Your new version was equal to std::memcmp for short strings and much slower (about half the speed) for long strings.

Here is a quick bench of your algorithm, my new version, and std::memcmp: https://quick-bench.com/q/IwKRemoPfvI8t2CYqiZR1nXgHEM

I'd be curious how you're benchmarking. You need to be very careful to not let compilers optimize away parts of your benchmark.

RealTimeChris:
> @RealTimeChris, so I was able to improve on your code and make a version that is consistently faster for both short and longer strings. However, in benchmarking, I found that my new version beat std::memcmp for short strings, but std::memcmp still wins out for longs strings (probably because it can work on 16 byte chunks). Your version was equal to std::memcmp for short strings and much slower (about half the speed) for long strings.
> 
> Here is a quick bench of your algorithm, my new version, and std::memcmp: https://quick-bench.com/q/IwKRemoPfvI8t2CYqiZR1nXgHEM
> 
> I'd be curious how you're benchmarking. You need to be very careful to not let compilers optimize away parts of your benchmark.

|               ns/op |                op/s |    err% |     total | benchmark
|--------------------:|--------------------:|--------:|----------:|:----------
|            2,479.16 |          403,362.10 |    4.2% |      3.13 | `memcmp_test - 2 char`
|            1,728.24 |          578,623.85 |    1.9% |      2.13 | `new_test - 2 char`
|            1,748.35 |          571,967.90 |    1.3% |      2.10 | `original_test - 2 char`
|            3,843.80 |          260,159.08 |    3.1% |      4.57 | `memcmp_test - 4 char`
|            1,963.16 |          509,382.01 |    0.9% |      2.34 | `new_test - 4 char`
|            1,733.65 |          576,816.83 |    1.7% |      2.15 | `original_test - 4 char`
|            4,892.61 |          204,389.80 |    6.1% |      5.92 | `memcmp_test - 8 char`
|            2,737.77 |          365,260.64 |    1.1% |      3.27 | `new_test - 8 char`
|            3,249.41 |          307,748.53 |    1.9% |      3.95 | `original_test - 8 char`
|            6,605.11 |          151,398.04 |    0.7% |      7.91 | `memcmp_test - 16 char`
|            3,680.56 |          271,697.93 |    0.7% |      4.43 | `new_test - 16 char`
|            3,697.01 |          270,488.54 |    1.2% |      4.47 | `original_test - 16 char`
|            7,873.37 |          127,010.36 |    1.2% |      9.52 | `memcmp_test - 32 char`
|            3,805.94 |          262,746.85 |    0.9% |      4.54 | `new_test - 32 char`
|            4,444.17 |          225,014.13 |    1.0% |      5.31 | `original_test - 32 char`

With the following code (Btw I found that using 'while' instead of 'if' with your code speeds it up a little, in addition to you forgot to add the uint8_t length):
https://github.com/RealTimeChris/Jsonifier/blob/nanobench-test/Tests/main.cpp

stephenberry:
Thanks for the additional benchmarks. I'm confused as to why the `uint8_t` loop is needed, because we can just check the last character instead of doing a `memcpy`, since the length that is left can only be one or zero.

I would also recommend using std::string for your benchmark, because it makes it harder for the compiler to optimize things away (can't as easily see the compile time size). It might now make a difference, but it's worth checking.

I'm surprised the while loop is appearing faster for you, as in compiler explorer I saw more complex assembly generated than the `if`. I'll have to look into it more.

I'll also note that you don't need to default initialize your `value01`, because you're about to write into the entire memory. The compiler will probably optimize this away in Release, but it would probably make Debug faster.

RealTimeChris:
> Thanks for the additional benchmarks. I'm confused as to why the `uint8_t` loop is needed, because we can just check the last character instead of doing a `memcpy`, since the length that is left can only be one or zero.
> 
> I would also recommend using std::string for your benchmark, because it makes it harder for the compiler to optimize things away (can't as easily see the compile time size). It might now make a difference, but it's worth checking.
> 
> I'm surprised the while loop is appearing faster for you, as in compiler explorer I saw more complex assembly generated than the `if`. I'll have to look into it more.
> 
> I'll also note that you don't need to default initialize your `value01`, because you're about to write into the entire memory. The compiler will probably optimize this away in Release, but it would probably make Debug faster.

I just figured that if the while loop was running faster than the if loop it would make sense to use it for the uint8_t loop as well although I suppose we can just direct compare it and see how that goes.

RealTimeChris:
> Clang compiles to using `bcmp` on Linux when calling `std::memcmp`, which I see in the assembly. This makes it faster than than our memcpy approach for long strings. However, `bcmp` has been deprecated on linux.
> 
> Clang using `bcmp` - https://news.ycombinator.com/item?id=21019495 `bcmp` deprecated - https://man7.org/linux/man-pages/man3/bcmp.3.html (even though it is faster than memcmp and not actually the same)
> 
> Clang is faster with `if` (GCC is the same for `if` and `while`), MSVC is the outlier with `while` being faster. But, MSVC just generates worse assembly overall.
> 
> This memcpy approach is faster for shorter strings on all platforms.
> 
> Those are my current findings.

Is it that memcpy is faster on shorter strings - or when the length of the string breaks down evenly into perfect uint-x sizes? IE: 2, 4, or 8 bytes?

stephenberry:
I've just created a new merge request with an even faster version of string comparison. See #925
I realized that if the count is greater than 8 we can just keep working on 8 byte chunks.

@RealTimeChris, let me know if you find any downsides to this approach.

PR #943, opened 2024-04-22
Add `build2` package reference to `README.md`

helmesjo:
Hi! I've packaged this (amazing) library for [build2](https://build2.org), so figured I might as well add it as a reference.
You can find the package [here](https://github.com/build2-packaging/glaze).

`v2.4.0` has already been released and I'm just about to publish `v2.5.3`.

You can see all the `v2.5.3` CI builds [here](https://queue.cppget.org/libglaze/2.5.3) (60 different configurations, freely available as part of `build2`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,14 @@ find_package(glaze REQUIRED)
 target_link_libraries(main PRIVATE glaze::glaze)
 ```
 
+### [build2](https://build2.org)
+
+- Available on [cppget](https://cppget.org/libglaze)
+
+```
+import libs = libglaze%lib{glaze}
+```
+
 ### Arch Linux
 
 - AUR packages: [glaze](https://aur.archlinux.org/packages/glaze) and [glaze-git](https://aur.archlinux.org/packages/glaze-git)


stephenberry:
Sweet, thanks for the support. I'll merge this in.
FYI, I noticed the dependency question mark: `libasio ^1.29.0 ?`
Glaze's test code depends on boost ut, Eigen, and the standalone asio. However, the primary code base has no dependencies and only requires the appropriate Eigen or asio headers to be available if using that corner of the Glaze library. Most developers use Glaze without dependencies and Glaze never brings in dependencies via its CMake when linked.

helmesjo:
> Sweet, thanks for the support. I'll merge this in. FYI, I noticed the dependency question mark: `libasio ^1.29.0 ?` Glaze's test code depends on boost ut, Eigen, and the standalone asio. However, the primary code base has no dependencies and only requires the appropriate Eigen or asio headers to be available if using that corner of the Glaze library. Most developers use Glaze without dependencies and Glaze never brings in dependencies via its CMake when linked.

Great! Regarding `libasio` it is set as a conditional dependency (see [here](https://github.com/build2-packaging/glaze/blob/main/libglaze/build/root.build#L14)), and is brought in if `config.libglaze.repe_rpc = true` (either by consumer or whatever the default is, and I did set it to `true` by default but in hindsight maybe a bit premature since that feature is still WIP according to the docs). But generally the `build2` approach is to prefer "batteries included" and instead opt out of features.

Issue #956, opened 2024-04-24
Split `make_reflectable` to a separate header


Chaoses-Ib:
> ## Mixing glz::meta reflection with pure-reflected structs
> If you have an struct with constructors you may need to use `glz::meta` to define the reflection. But, now your type might break the counting of the numbers of members when included in another struct.
> 
> To fix this, add a constructor for your struct to enable the proper counting: `my_struct(glz::make_reflectable) {}`

The problem is, `make_reflectable` is declared in `reflection/reflect.hpp`, which requires `char8_t` to be available and doesn't compile with [`/Zc:char8_t-`](https://learn.microsoft.com/en-us/cpp/build/reference/zc-char8-t?view=msvc-170). If `make_reflectable` could be split to a separate header, this problem could be avoided and make Glaze less intrusive.

Chaoses-Ib:
In case anyone has similar issues, here is a workaround:
```cpp
namespace glz
{
   struct dummy;
   using make_reflectable = std::initializer_list<dummy>;
}


struct V2
{
   float x{};
   float y{};

   V2(glz::make_reflectable) {}
   V2() = default;

   V2(V2&&) = default;
   V2(const float* arr) : x(arr[0]), y(arr[1]) {}
};
```

stephenberry:
I think the compiler is confused. `char8_t` is used in `glaze/core/common.hpp`, but currently only used in `glaze/json/study.hpp`.

If I move `basic` and `basic_ptr` into `glaze/json/study.hpp`, then `char8_t` won't be used for the majority of users, since `glaze/json/study.hpp` is not included anywhere else in Glaze.

stephenberry:
I just merged #957 into main. Let me know if this solves your problem.

Chaoses-Ib:
It works now. Thanks for your quick fix.

stephenberry:
Excellent!

PR #957, opened 2024-04-24
Moving basic and basic_ptr into study.hpp

stephenberry:
This should remove the need for `char8_t` support for those who don't use `study.hpp`


stephenberry:
Should address #956

PR #1001, opened 2024-05-16
Cleanup

stephenberry:
- dump_newline_indent to clean code an slightly optimize

### DIFF
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -453,8 +453,7 @@ namespace glz
          if (!empty_range(value)) {
             if constexpr (Opts.prettify) {
                ctx.indentation_level += Opts.indentation_width;
-               dump<'\n'>(args...);
-               dumpn<Opts.indentation_char>(ctx.indentation_level, args...);
+               dump_newline_indent<Opts.indentation_char>(ctx.indentation_level, args...);
             }
 
             auto it = std::begin(value);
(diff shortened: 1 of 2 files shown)


include/glaze/util/expected.hpp
```diff
@@ -59,18 +59,13 @@ namespace glz
 #if defined(__cpp_lib_expected)
 namespace glz
 {
-   template <class expected_t, class unexpected_t>
-   using expected = std::expected<expected_t, unexpected_t>;
+   template <class Expected, class Unexpected>
+   using expected = std::expected<Expected, Unexpected>;
 
-   template <class unexpected_t>
```
jbbjarnason, line 65: @stephenberry removing this broke compilation with clang 18, it seems that the ctad for alias is not supported until clang 19. Could this particular change be reverted?

For reference, https://github.com/stephenberry/glaze/pull/237

stephenberry: @jbbjarnason 
Oh, thanks for pointing this out. I figured since clang 15 built it wouldn't be broken later. The problem is that GCC 14 is broken without this change. Before GCC 14 the code works.

This is a weird case where GCC and Clang introduced bugs in future versions that contradict one another. We can just add a macro to switch between GCC and Clang. If you make this change feel free to merge it ASAP, otherwise I might not be able to get to it until next week.

jbbjarnason: I can make the change next Tuesday

stephenberry: Cool, I'll see if I can make some time today to fix it. Thanks for being willing to help.

stephenberry: @jbbjarnason
I realized I really needed to setup actions for testing clang 16, 17, and 18 so that we don't run into these issues again. I'm actively working on this here: #1006

stephenberry: #1006 merged

PR #1006, opened 2024-05-17
Run actions for Clang 16, 17, & 18 on linux

stephenberry:


stephenberry:
@jbbjarnason
I've merged this fix, which now passes with address sanitization on Clang 16, 17, and 18. I'll release a new version soon.

There were some sanitizer warnings about the thread_pool (which you probably aren't using), which I'll look into later, but I think it is the unit test library that is creating the problem (I've been dealing with issues with boost-ut).

jbbjarnason:
Great thanks 👍 

Yes you are right, we are currently not using the thread_pool.

Well done

PR #1012, opened 2024-05-20
Cleanup of dump.hpp

stephenberry:
- Removed dead code
- More generic handling of buffer byte types, better support for std::byte
- Merged handling of raw buffer function calls and vector_like buffers


stephenberry:
GCC14, why do you have to segfault here?

PR #1013, opened 2024-05-21
JsonSchema add advanced flag and unit support

jbbjarnason:
Two additional json schema features,

- format https://www.learnjsonschema.com/2020-12/format-annotation/format/
- out of specification
  - unit ascii 
  - unit unicode
  - advanced flag

The out of specification is up for discussion, the use case for units relates to usage of https://github.com/mpusz/mp-units units library, to reflect on the actual type and scaler of each number. The use case of `advanced` is generation of configuration view in UI, so the programmer writing logic can mark parameter advanced so it would not show in default view.

I will follow up with small json schema changes after this PR.

### DIFF
--- a/include/glaze/json/schema.hpp
+++ b/include/glaze/json/schema.hpp
@@ -26,14 +26,12 @@ namespace glz
    namespace detail
    {
       enum struct defined_formats : std::uint8_t;
-      struct out_of_spec final
+      struct ExtUnits final
       {
-         std::optional<std::string_view> unit_ascii{}; // ascii representation of the unit, e.g. "m^2" for square meters
+         std::optional<std::string_view> unitAscii{}; // ascii representation of the unit, e.g. "m^2" for square meters
          std::optional<std::string_view>
-            unit_unicode{}; // unicode representation of the unit, e.g. "m²" for square meters
-         std::optional<bool>
-            advanced{}; // flag to indicate that the parameter is advanced and can be hidden in default views
-         constexpr bool operator==(const out_of_spec&) const noexcept = default;
+            unitUnicode{}; // unicode representation of the unit, e.g. "m²" for square meters
+         constexpr bool operator==(const ExtUnits&) const noexcept = default;
       };
    }
    struct schema final
@@ -78,8 +76,10 @@ namespace glz
       // properties
       std::optional<std::span<const std::string_view>> enumeration{}; // enum
 
-      // out of spec
-      std::optional<detail::out_of_spec> outOfSpec{};
+      // out of json schema specification
+      std::optional<detail::ExtUnits> extUnits{};
+      std::optional<bool>
+         extAdvanced{}; // flag to indicate that the parameter is advanced and can be hidden in default views
 
       static constexpr auto schema_attributes{true}; // allowance flag to indicate metadata within glz::object(...)
 
@@ -116,7 +116,8 @@ namespace glz
                                                    "maxContains", &T::maxContains, //
                                                    "uniqueItems", &T::uniqueItems, //
                                                    "enum", &T::enumeration, //
-                                                   "outOfSpec", &T::outOfSpec);
+                                                   "extUnits", &T::extUnits, //
+                                                   "extAdvanced", &T::extAdvanced);
       };
    };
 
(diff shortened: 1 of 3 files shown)


include/glaze/json/quoted.hpp
```diff
@@ -107,6 +107,16 @@ namespace glz
          }
       };
 
+      template <class T>
+      struct from_json<raw_t<T>>
```
jbbjarnason, line 111: This is needed for schema examples, was there any reason for this not being implemented?

jbbjarnason:
converting to draft while I resolve issues with workflows

include/glaze/json/quoted.hpp
stephenberry: Nope, just never needed to deserialize it before, so it was overlooked.

stephenberry:
@jbbjarnason, sweet! More JSON schema support and tests are always appreciated.

As per `std::variant<bool, std::int64_t, std::uint64_t, double, std::string_view>;` I assume your issue is that this is not auto-deducible. Are you reading into the JSON schema? I understand your issue in this context.

I'm tempted to add support for reading these kinds of variants, where if a `.` is encountered it is parsed as the floating point value in the variant, and if a `-` is encountered (and its not floating point) then it is parsed as a signed integer. And, if it is not floating point or signed, then it is parsed as an unsigned integer.

But, in your case, could you simply use `double`? Or, do you need integer support?

stephenberry:
I'm okay duplicating for now, until we see a clear path forward or add more advanced support for mixed number types.

jbbjarnason:
> I'm okay duplicating for now, until we see a clear path forward or add more advanced support for mixed number types.

I skipped the default value test for the time being.

stephenberry:
I would rather not use the name `outOfSpec` for a units. I could see the desire to add many more "outOfSpec" fields to our JSON schema documents. So, I think it is better to name things based on what purpose they serve.

I think we should also choose a naming convention that will not collide with JSON schema, to avoid JSON schema adding a colliding name in the future and breaking the code. JSON schema starts field names with `$` or a lowercase character, so I would suggest we start our "out of spec" extensions with "Ext".

My recommendation would be to rename your `outOfSpec` here to `ExtUnits`, because you are adding unit support, which I think is excellent.

The ability to mark things `advanced` is also helpful.

jbbjarnason:
> I would rather not use the name `outOfSpec` for a units. I could see the desire to add many more "outOfSpec" fields to our JSON schema documents. So, I think it is better to name things based on what purpose they serve.
> 
> I think we should also choose a naming convention that will not collide with JSON schema, to avoid JSON schema adding a colliding name in the future and breaking the code. JSON schema starts field names with `$` or a lowercase character, so I would suggest we start our "out of spec" extensions with "Ext".
> 
> My recommendation would be to rename your `outOfSpec` here to `ExtUnits`, because you are adding unit support, which I think is excellent.
> 
> The ability to mark things `advanced` is also helpful.

I have seen somewhere that the key is prefixed with `x-` for out of spec support. However, I do not have strong opinion on this so `Ext` is fine with me.

In terms of  `advanced`, it is not coupled with units support, would you suggest making `advanced` member of `schema` struct, then as `ExtAdvanced` ?

stephenberry:
`x-` isn't a valid C++ prefix, so `x_` would be better to support C++ reflection. But, I could see fields like `x_axis` being a thing, so I think `Ext` is a better prefix.

I do like the idea of `ExtAdvanced` since it is generally useful.

jbbjarnason:
> `x-` isn't a valid C++ prefix, so `x_` would be better to support C++ reflection. But, I could see fields like `x_axis` being a thing, so I think `Ext` is a better prefix.
> 
> I do like the idea of `ExtAdvanced` since it is generally useful.

I see your point, good catch!

I have changed it, I followed the convention of camel casing with lower case first letter, would you rather like upper case for the external members?

stephenberry:
I could go both ways, `ExtAdvanced` is a little safer from colliding with future JSON schema, but I don't expect JSON schema to add new fields that begin with `ext`

stephenberry:
@jbbjarnason, I think this looks great, go ahead and merge it!

Issue #981, opened 2024-05-06
Support 0 and 1 as boolean values


stephenberry:
### Discussed in https://github.com/stephenberry/glaze/discussions/978

<div type='discussions-op-text'>

<sup>Originally posted by **scalpel4k** May  5, 2024</sup>
I know it's not officially supported to use 0 and 1 as boolean values. However, reality looks a bit different and I know a few situations especially in the embedded world where this is common practice.

It would be really cool to have a compile time option to convert between numeric values and booleans. It would be way easier than writing custom converters all the time....

thx Michi </div>

stephenberry:
Support has been added in #1022 . You can use the compile time option `bools_as_numbers = true` or the wrapper by the same name.

PR #1022, opened 2024-05-22
Wrapper simplification and bools_as_numbers option

stephenberry:
- Added opts_wrapper_t, which allows new `glz::opts` based wrappers to be created.
- Added `bools_as_numbers` option to `glz::opts` for #981


stephenberry:
@jbbjarnason, just wanted to make you aware of this improvement, which is adding `raw` wrapper handling to reading as a side affect. This makes it easier to create new wrappers for booleans in `glz::opts`.

PR #1038, opened 2024-05-29
 json schema enhancements

jbbjarnason:
Couple of enhancements:
- Always null types write null as `const` to json schema payload
- `const` supports "any" type now for enums and numbers for example
- (Major) change, introduced attributes of type `schema` into `schematic` to embed info to each type, however, it bloates the code a bit but does not duplicate the variables. Note I moved to using enumeration from schema instead of its duplicate, however, maybe it should be duplicated it's a bit simpler to work with vector instead of span. `constant` was also moved but is the same.
This change is breaking change, is it okay?
- Correctly handling  directly accessed glaze conversion e.x. `struct glz::meta<T>{ static constexpr value = &T::some_var; };`. In case that `some_var` is constexpr the `const` is populated 
- Add min and max for numbers
- Minor enhancements for enum
- Objects, validate ref and statically assert the name string.


include/glaze/json/schema.hpp
```diff
@@ -537,10 +612,15 @@ namespace glz
                   }
                   else if constexpr (std::is_convertible_v<additional_data_type, schema>) {
                      ref_val = get<comment_index>(item);
+                     validate_ref<name_v<val_t>>();
                      ref_val.ref = join_v<chars<"#/$defs/">, name_v<val_t>>;
                   }
                }
 
+               if (!def.type) {
```
jbbjarnason, line 620: I don't remember why this was needed but it was required for our frontend at the time which I do not currently have access to. This is trivial change though so I don't see a problem with it. Please comment if you feel otherwise

stephenberry: I don't see a problem with it

stephenberry:
Lots of good changes, it's nice to have the `schema` attributes included in `schematic`

include/glaze/core/common.hpp
jbbjarnason, line 1: @stephenberry could you review these changes? `has_static_size` was behaving differently clang on Linux so I added more restrictions to it.

stephenberry: This looks fine, I'm just curious why the `value_type` needs to be trivially copyable?

jbbjarnason: Maybe the trivially copyable type trait is not the correct one, my assumption is from the use case from binary write, I glimpsed at this code:
```cpp
               if constexpr (has_static_size<T>) {
                  constexpr auto num_bytes = (value.size() + 7) / 8;
                  std::array<uint8_t, num_bytes> bytes{};
                  for (size_t byte_i{}, i{}; byte_i < num_bytes; ++byte_i) {
                     for (size_t bit_i = 7; bit_i < 8 && i < value.size(); --bit_i, ++i) {
                        bytes[byte_i] |= uint8_t(value[i]) << uint8_t(bit_i);
                     }
                  }
                  dump(bytes, args...);
               }
```
And thought to myself that you are probably using this to write the bytes from some fixed size container value, and to safely `memcpy` the data from the value the value_type would need to be trivially copyable. 

For reference, the concept before behaves differently on clang vs gcc or msvc https://godbolt.org/z/7jYT9e38T

stephenberry: Ah, I hadn't looked at its use case. I think your changes make sense.

PR #1055, opened 2024-06-06
Add fallback and option to boost asio

SiebrenW:
Do tell if you need me to add some documentation or if you think this implementation is a bit "leaky". I could put the using asio = boost::asio inside of the glz scope for instance.


stephenberry:
Yeah, that's a good idea. Put the namespace alias inside the `glz` namespace.

stephenberry:
Thanks for this! I'll merge it in.

PR #1064, opened 2024-06-09
fix make_map_impl

huangminghuang:
This PR fixes the implementation of `make_map_impl()` when `n >= 64`.


stephenberry:
Interesting, thanks!

PR #1098, opened 2024-06-13
Allow variants of variants

shiretu:


stephenberry:
Thanks for this, will have to check out why it isn't building later.

shiretu:
I see that is failing in some other places. In my use case, it works fine: I have 3 levels of variants. Yes, I know, it is not fine, but what can I do!? This is how the data is modelled from server side :(

stephenberry:
I'll look into it when I have some time, I think it should be possible to get working.

shiretu:
Thank you! It would be very helpful.

stephenberry:
There is a lot more to variant handling that this pull request tackles. I can see what you're trying to accomplish, and I'd like to get variants of variants working, but we need to look at all sub-variants and count the various types in each of the variants. A sub-variant may just have numerical types and therefore not contribute to the deduction map of the higher level variant, or it may just have one element that contributes.

I'm doing more work on variants now. I just added support for variants that contain std::shared_ptr and std::unique_ptr.

I'm going to close this pull request, but I appreciate a demonstration of your aim and will try to tackle this in a more robust manner.

Thanks!

PR #1105, opened 2024-06-20
JSON concepts

stephenberry:
Adding JSON concepts for simple type checking in the user facing API:
```c++
static_assert(glz::json_string<std::string>);
static_assert(glz::json_string<std::string_view>);
static_assert(glz::json_object<my_struct>);
static_assert(glz::json_array<std::array<float, 3>>);
static_assert(glz::json_boolean<bool>);
static_assert(glz::json_number<float>);
static_assert(glz::json_integer<uint64_t>);
static_assert(glz::json_null<std::nullptr_t>);

```

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -2,7 +2,7 @@
 
 ### What happens when the JSON input is missing objects? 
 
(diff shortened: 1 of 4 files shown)


stephenberry:
Can't use `is_object` type of naming because we want these functions available for the `json_t` API. Also, while these are applicable for BEVE, they are more JSON focused because they follow JSON typing.

PR #1106, opened 2024-06-20
glz::format_to for numbers

stephenberry:
Improves performance of `glz::mustache`


include/glaze/util/itoa.hpp
```diff
@@ -132,7 +132,7 @@ namespace glz
 
    template <class T>
       requires std::same_as<T, std::remove_cvref_t<int32_t>>
```
jbbjarnason, line 134: Should probably be:
```suggestion
requires std::same_as<std::remove_cvref_t<T>, int32_t>
```

stephenberry: Thanks! Fixing in #1120

PR #1107, opened 2024-06-20
Cross platform sockets

stephenberry:
# Cross platform sockets
- Adding `glz::socket` and `glz::server` for TCP networking
- Builds upon these low level tools to replace `glz::asio_server` with `glz::repe_server` and `glz::asio_client` with `glz::repe_client`

> [!IMPORTANT]
> This code is cleaner and more future proof using `std::format`. When this is merged we will drop support for GCC 12 to support `std::format`.

## Removing asio
This change replaces the `glz::asio_server` and `glz::asio_client` with `glz::repe_server` and `glz::asio_client`.
- Removes the asio third party dependency
- Allows recursive request calls with clients
- Allows us to develop networking for Glaze at a faster pace
- Allows us to optimize networking
- Allows tighter integration with RPC and network protocols (REPE, etc.)


stephenberry:
@meftunca, I'm pinging you to let you know about this socket development for Glaze. After running into limitations with asio, wanting to push performance and ease of use, and realizing that asio does not accept much outside influence, we've decided to implement our own cross platform sockets in Glaze.

We're able to use a simpler approach by not needing to support older versions of C++ like asio. And, we can more cleanly handle various protocols and formats by using the same template options. The socket class is still very abstract, but it enables us to send either JSON or binary without the need for an additional protocol. But, we can also build a protocol like REPE on top of it (which I'm also refactoring) so that we can more efficiently handle RPC calls and more.

meftunca:
I wrote a simple socket service with io_uring and kqueue and I was able to handle more than 150k requests on average.  My personal interpretation is that libraries like asio will continue to be developed by people with different concerns to fit certain patterns. Doing an independent development on Glaze is really the right conclusion. 

Is there any planning or roadmap for the Socket project?

meftunca:
https://github.com/meftunca/muxpp
 To be honest, I'm not very knowledgeable about socket programming, but I wrote a learning project anyway. I designed an ExpressJS-like interface. I was hoping to complete phase 1 by finishing client-server communication by routing the final requests to the relevant EndPoints :D

meftunca:
It can also be more performant to route large requests to coroutines for workflow management, congestion reduction and performance.  I am sure it will be more performant if coroutine support is provided as an option.

meftunca:
After analyzing some socket code,  realized that you use threads very generously :) At first, I used threads generously in this way and I lost performance, let alone the performance I expected.

stephenberry:
Thanks for the feedback. Yes, currently the threading is on a per client basis, which isn't the best approach for a lot use cases. I plan to add a thread-pool solution as well as coroutines in the future. This is just an initial implementation to figure out the API and architecture.

sam20908:
Just wondering: does this plan to support websockets? Or is there a way I can attach existing websocket to this?

stephenberry:
@sam20908, yes within the next few months we plan to add Websocket support. The goal is to have the Websocket protocol separate from the socket code as well, so either the protocol or the socket code could be swapped out.

Right now we're working on cross platform coroutines with will be added to Glaze to support efficient server implementations.

sam20908:
Nice to hear! I realized that is there a way one can use their own connection? E.g. I use MBedTLS to create a TLS connection (adding TLS support for Glaze sounds like pain)

stephenberry:
Yes, I'd like to use templates on the sockets and create a general wrapping interface that will allow any networking library to work with the rest of the protocols, coroutines, etc.

Nother62:
Hey,

Is there an ETA for when this PR will be merged into the main branch? I would love to use this for my current project, which would improve my current implementation of RPC.

Thanks

stephenberry:
We're still figuring out the best approach to handling the underlying concurrency/scheduling. It's hard to say when this will get merged. I would say it will probably take until the end of September. We've decided to use https://github.com/NVIDIA/stdexec, which has been accepted into C++26. It could be sooner, but I don't want to disappoint.

sam20908:
btw im wondering if theres a reason why you're implementing ur own sockets instead of having the users getting their own sockets setup and just parse the json response using this library?

stephenberry:
The socket implementation has to be somewhat tied to the scheduler (I think). I wish https://github.com/dietmarkuehl/stdnet were further along, but it isn't mature and is not cross platform. As much as possible we want to separate the sockets, scheduler, payload spec, and RPC spec, but we need cross platform versions of all these that we can iterate on to figure out a general solution.

sam20908:
I don't understand, there are cross platform libraries that handles socket connection. What does this do that other socket libraries don't

stephenberry:
Currently Glaze uses asio, and I intend to still allow asio sockets. I've also experimented with libuv sockets. But, cross platform sockets is the easiest part of this implementation and it would be nice to not need third party dependencies.
There's also the fact that we'd like to eventually handle HTTP3 with QUIC, which asio doesn't have support for. One can argue that just simple UDP sockets can be used underneath a QUIC implementation, but then why bring in a big dependency like asio to use only a tiny piece of it. There are also some optimizations that get easier to manage when it's possible to manage the socket code oneself, particularly when it comes to header handling in various protocols. The project I'm working on has networking needs for various context, web UI, hardware interfaces with extremely low latency, and more, so there's a lot to consider.

sam20908:
ok that makes more sense, i look forward to use it in my project when websocket support comes lol

wilkolbrzym-coder (commented):
# Review — PR #1107

I went file by file through the visible diff.

---

## 1. `include/glaze/network/server.hpp`

### 1.1 Callback signature mismatch

`server.hpp` invokes the callback as:

```cpp
callback(thread_pool.get_scheduler(), socket{client_fd}, active);
```

But `repe_server.hpp` supplies:

```cpp
server.accept([this](socket&& socket, auto& active) {
```

This will not compile. The callback receives three arguments but the lambda accepts two.

`socket_test.cpp` uses yet another signature:

```cpp
server.async_accept([](auto sched, glz::socket client_socket, std::atomic<bool>& active) {
```

At least one of these call sites is guaranteed to fail.

---

### 1.2 Thread pool may be constructed with zero threads

```cpp
exec::static_thread_pool thread_pool{std::thread::hardware_concurrency() / 2};
```

`hardware_concurrency()` may return `0` or `1`, producing zero-sized thread pool.  
No `std::max(1, ...)` guard exists.

---

### 1.3 Destructor may deadlock

```cpp
~server()
{
   active.store(false);
   scope.request_stop();
   thread_pool.request_stop();
   stdexec::sync_wait(scope.on_empty());
}
```

If a task does not respond to stop, or if destruction happens from a thread owned by `thread_pool`, `sync_wait` can block indefinitely.

---

### 1.4 Windows timeout is mishandled

```cpp
n = WSAWaitForMultipleEvents(1, &event_fd, FALSE, 10, FALSE);
```

`WSA_WAIT_TIMEOUT` is not equal to `WSA_WAIT_FAILED`. The branch:

```cpp
if (n == GLZ_WAIT_FAILED) {
#if ...
#else
   if (n == WSA_WAIT_TIMEOUT) continue;
#endif
```

puts the timeout check inside the `n == GLZ_WAIT_FAILED` branch. Therefore, the timeout path never executes `continue`; instead, it falls through to `WSAEnumNetworkEvents` every 10 ms.

This may work, but it is clearly not what the code intended.

---

### 1.5 Error information is lost

```cpp
const auto ec = accept_socket.bind_and_listen(port);
if (ec) {
    return {ip_error::socket_bind_failed, ip_error_category::instance()};
}
```

The original error code is discarded. Every bind/listen failure is reported as `socket_bind_failed`.

---

## 2. `include/glaze/network/repe_server.hpp`

### 2.1 Already mentioned callback mismatch

The two-argument lambda cannot be called by `server.hpp`.

### 2.2 Receive loop continues on non-fatal errors

```cpp
while (active) {
   if (auto ec = receive(socket, buffer)) {
       if (ec.value() == ip_error::client_disconnected) {
           return;
       }
   }
   else {
       ...
   }
}
```

If `receive` returns an error other than `client_disconnected`, the loop continues immediately. Depending on the error, this can become a busy-spin and repeatedly log the same failure.

---

## 3. `include/glaze/network/repe_client.hpp`

### 3.1 Wrong error code in `init()`

```cpp
if (not socket.no_delay()) {
    return {ip_error::socket_bind_failed, ip_error_category::instance()};
}
```

`no_delay()` is not related to socket binding. This should be a socket-option error, not `socket_bind_failed`.

### 3.2 Buffer is not cleared between send and receive

Example from `get`:

```cpp
repe::unique_buffer ubuffer{buffer_pool.get()};
auto& buffer = ubuffer.value();

send<Opts>(socket, buffer);
receive<Opts>(socket, buffer);
```

The same buffer is used for transmission and then handed directly to receive.  
Unless `receive` explicitly clears or resizes the buffer first, stale transmitted data may remain.

### 3.3 `call(repe::header&& header)` bypasses the normal REPE request path

This overload writes:

```cpp
glz::write_json(std::forward_as_tuple(std::move(header), nullptr), buffer);
```

while all other methods use:

```cpp
repe::request<Opts>(...)
```

That is inconsistent and may not produce the same wire format as the other calls.

### 3.4 Several methods return `repe::error_t` but swallow the actual `std::error_code`

Errors from `send` and `receive` are converted to `{ec.value(), ec.message()}`.  
If `ec.message()` is empty or generic, important context is lost.

---

## 4. `include/glaze/network/socket_io.hpp`

### 4.1 Async send/receive hide errors

```cpp
return exec::repeat_effect_until(
    stdexec::then([&]() {
        auto result = sock.send(buffer, bytes_sent);
        if (result == socket::io_result::error) {
            throw std::system_error(...);
        }
        return result == socket::io_result::completed;
    }) |
    stdexec::upon_error([](auto&&) { return true; }));
```

`upon_error` converts an exception into `true`, terminating the repeat effect **as if it succeeded**.

The caller of `async_send` cannot know whether the send failed.

The same pattern exists in `async_receive`.

### 4.2 Senders capture references to local objects

`async_send`:

```cpp
stdexec::let_value([&sock, buffer = std::string(buffer)]() mutable {
    size_t bytes_sent = 0;
    return exec::repeat_effect_until(
        stdexec::then([&]() {
            auto result = sock.send(buffer, bytes_sent);
            ...
        })
    );
});
```

The inner lambda captures `sock`, `buffer`, and `bytes_sent` by reference.

The sender may outlive the current function call. If those referenced objects are destroyed, this is use-after-free.

The same applies to:

- `async_connect`
- `async_bind_and_listen`
- `async_receive`
- `async_receive_value`
- `async_send_value`

### 4.3 `async_receive_value` ignores the received header

```cpp
auto [header, buffer] = std::move(result);
auto ec = glz::read<Opts>(value, buffer);
```

The `header` variable is never used to verify the payload length.  
A truncated or oversized frame may still be parsed as success.

### 4.4 Host-endian header transmission

```cpp
uint64_t header = uint64_t(buffer.size());
return async_send(sched, sckt,
    std::string_view{reinterpret_cast<char*>(&header), sizeof(header)});
```

The header is the raw host-endian `uint64_t`.  
There is no explicit conversion to network byte order or a portable format.

### 4.5 Static thread-local buffers in synchronous `send_value` / `receive_value`

```cpp
static thread_local std::string buffer{};
```

This is safe only for the simplest sequential use.  
Nested or re-entrant calls on the same thread can overwrite the buffer.

---

## 5. `cmake/stdexec.cmake`

### 5.1 Unpinned moving dependency

```cmake
set(branch_or_tag "main")
set(url "https://github.com/NVIDIA/stdexec.git")

execute_process(
    COMMAND git clone --depth 1 --branch "${branch_or_tag}" --recursive "${url}" "${target_folder}"
)
```

This fetches `main` at configure time. No commit SHA, tag, or content hash.  
Every configure can produce a different dependency version.

### 5.2 Requires `git` and network

The build is not hermetic. A header-only library should not silently clone a repository during CMake configure.

### 5.3 `stdexec_INCLUDE_DIR` leaks to consumers

Root `CMakeLists.txt`:

```cmake
target_include_directories(
    glaze_glaze ...
    INTERFACE ...
    "$<BUILD_INTERFACE:${stdexec_INCLUDE_DIR}>"
)
```

This means any consumer of `glaze::glaze` now needs `stdexec` include path available.

---

## 6. Root `CMakeLists.txt`

### 6.1 MSVC interface flags are baked at configure time

```cmake
if (MSVC)
  string(REGEX MATCH "\/cl(.exe)?$" matched_cl ${CMAKE_CXX_COMPILER})
  if (matched_cl)
    target_compile_options(glaze_glaze INTERFACE "/Zc:preprocessor" /permissive- /Zc:lambda /Zc:__cplusplus)
```

This is the same consumer-side leak described in #2818.  
Flags are decided by Glaze’s own compiler and exported to every consumer.

### 6.2 Unused vendored CPM.cmake

`cmake/CPM.cmake` is added but no code uses it.  
`stdexec` is fetched through a hand-written `git clone`, not through CPM.  
This is dead code and increases maintenance surface.

---

## 7. Tests

### 7.1 `tests/socket_test/socket_test.cpp` contains `std::abort()`

```cpp
suite make_server = [] {
   std::abort(); // IMPORTANT: comment this out when testing...
```

This suite kills the test process unconditionally.  
It is not a test; it is a hard kill switch.

### 7.2 `std::vector<std::future<void>> threads(n_clients);` is pre-filled

```cpp
std::vector<std::future<void>> threads(n_clients);
...
for (size_t id{}; id < n_clients; ++id) {
    threads.emplace_back(std::async(...));
}
```

The vector initially contains `n_clients` default-constructed `std::future<void>` objects.  
`emplace_back` then adds `n_clients` more. Final size is `2 * n_clients`.

Later:

```cpp
for (auto& t : threads) {
    t.get();
}
```

The first `n_clients` futures have no shared state. Calling `get()` on them throws `std::future_error`.

This is a deterministic test bug, independent of networking.

### 7.3 Port 8080 is hardcoded

Both server and client use `8080`.  
A parallel CI job or a local service using that port causes flaky failure.

### 7.4 No `server_thread.join()`

The server thread is never joined.  
The program relies on global destructors and process exit to clean up.

### 7.5 The test does not actually test async sender paths

Most `async_*` calls are commented out:

```cpp
/*auto send_sender = glz::async_send_value(...);
...
*/
```

The current test runs only the synchronous `send_value`/`receive_value` paths.

---

## 8. `tests/repe_server_client`

### 8.1 Old Asio includes remain in renamed files

The rename diff still shows transitions from:

```cpp
#include "glaze/ext/glaze_asio.hpp"
```

to:

```cpp
#include "glaze/network/repe_client.hpp"
```

That is fine if the final file is correct, but the diff itself demonstrates the branch has been repeatedly patched.

### 8.2 Server and client CMake include path changes

Client/server CMake files remove:

```cmake
${asio_SOURCE_DIR}/asio/include
```

and replace it with plain `include`.

Given that `glaze_asio.hpp` is deleted, this is expected.  
But it means there is no migration story for existing Asio users.

---

## 9. `include/glaze/core/error.hpp`

### 9.1 `message()` depends on `nameof` for enum conversion

```cpp
return std::string{nameof(error_code(ec))};
```

If `ec` contains an out-of-range value, the returned string may be empty or misleading.  
For a system-level error category used by networking, this is fragile.

---

## 10. `include/glaze/ext/glaze_asio.hpp` deletion

The entire Asio compatibility layer is removed.  
There is no deprecation header, forwarding shim, or release note.

For a public library, this is a hard source break.  
Even if Asio is being intentionally removed, the PR does not provide a migration path.

---

## 11. GitHub workflow

`gcc.yml` changes matrix from `[12, 13, 14]` to `[13, 14]`.

That is consistent with the README:

> GCC 12 is not supported due to lack of `std::format`.

No issue.

But the fact that this PR has **0 workflow checks** means none of the above was caught.

---

## 12. Overall assessment

This is not a code review candidate in its current form.

It combines:

- new socket primitives,
- stdexec integration,
- removal of Asio,
- new REPE client/server,
- CMake dependency fetching,
- test renames,
- missing CI.

There are deterministic compile-time bugs, deterministic test bugs, undefined behaviour risks, and supply-chain risks.

---

## 13. Recommendation

Close this PR in its current form.

If the goal is still to replace Asio, split the work into focused PRs:

1. Socket primitives only, with all platforms building in CI.
2. Synchronous REPE client/server using those primitives.
3. stdexec integration with pinned dependency and error propagation.
4. Asio removal with a deprecation shim or migration guide.
5. Networking tests that can run safely in CI.

cc @stephenberry

stephenberry:
@wilkolbrzym-coder, I will close this work. I have an alternative to asio in development and in good shape, but I don't have the time to release it yet.

PR #1116, opened 2024-06-26
to_json_schema nullable_t throws instead of segfault

jbbjarnason:
this can occur in special cases when user has it's own to_json_schema for custom type, where the user forgets to assign type to the schematic


stephenberry:
Nice, that's smarter

PR #1121, opened 2024-06-28
Reflection macro improvements

stephenberry:
Create an enum with the provided fields and names. A nameof reflection function is made.
Example: GLZ_ENUM_MAP(color, "Red", red, "Green", green, "Blue", blue);
nameof(color::red) == "Red"

### DIFF
--- a/include/glaze/core/common.hpp
+++ b/include/glaze/core/common.hpp
@@ -362,6 +362,11 @@ namespace glz
 
       template <class T>
       concept glaze_enum_t = glaze_t<T> && is_specialization_v<meta_wrapper_t<T>, Enum>;
+      
+      template <class T>
+      concept has_nameof = requires(T t) {
+          { glz::nameof(t) } -> std::convertible_to<std::string_view>;
+      };
 
       template <class T>
       concept glaze_flags_t = glaze_t<T> && is_specialization_v<meta_wrapper_t<T>, Flags>;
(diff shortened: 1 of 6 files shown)


stephenberry:
@jbbjarnason, I've created some new enumeration macros (GLZ_ENUM AND GLZ_ENUM_MAP), primarily for internal use in Glaze. These macros compile much faster (~20X faster) than the more flexible `glz::meta` enumerate approach. They don't have to do any complex tuple handling with lots of templates. I'm using this for the `error_code` and for `defined_formats`.

I'm pointing this out to you because I changed the underlying type for `defined_formats` from a `uint8_t` to a `uint32_t`. The reason is because it keeps the macros simpler. We use `defined_formats` in a `std::optional` so the alignment will remove the benefits of `uint8_t`. I just wanted these reflection macros to use a `uint32_t` because it can be faster in some contexts and we don't have to worry about running out of space in a `uint8_t`.

Just making you aware of this change in case it adversely affects your codebase.

jbbjarnason:
> @jbbjarnason, I've created some new enumeration macros (GLZ_ENUM AND GLZ_ENUM_MAP), primarily for internal use in Glaze. These macros compile much faster (~20X faster) than the more flexible `glz::meta` enumerate approach. They don't have to do any complex tuple handling with lots of templates. I'm using this for the `error_code` and for `defined_formats`.
> 
> I'm pointing this out to you because I changed the underlying type for `defined_formats` from a `uint8_t` to a `uint32_t`. The reason is because it keeps the macros simpler. We use `defined_formats` in a `std::optional` so the alignment will remove the benefits of `uint8_t`. I just wanted these reflection macros to use a `uint32_t` because it can be faster in some contexts and we don't have to worry about running out of space in a `uint8_t`.
> 
> Just making you aware of this change in case it adversely affects your codebase.

Sounds awesome, great job!

PR #1130, opened 2024-07-02
Silence MSVC warning about including `<expected>`

helmesjo:
Check the feature macro `__cpp_lib_expected >= 202202L` as per the [documentation](https://en.cppreference.com/w/cpp/feature_test) to silence the warning:

> warning STL4038: The contents of <expected> are available only with C++23 or later.


stephenberry:
Thanks!

PR #1150, opened 2024-07-15
Custom error messages

stephenberry:
Adds `std::string_view custom_error_message` to `glz::context` and `glz::error_ctx` to allow custom error propagation for reading/writing.

### DIFF
--- a/include/glaze/core/context.hpp
+++ b/include/glaze/core/context.hpp
@@ -37,20 +37,26 @@ namespace glz
             invalid_variant_string, //
             no_matching_variant_type, //
             expected_true_or_false, //
+            // Key errors
             unknown_key, //
+            missing_key, //
+            // Other errors
             invalid_flag_input, //
             invalid_escape, //
             u_requires_hex_digits, //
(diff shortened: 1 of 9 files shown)


stephenberry:
I think making the `glz::context` class more generally useful is better than having various context classes.

stephenberry:
@scalpel4k, does this addition of a std::string_view that you can customize with your own errors solve your problem? Does the variable name sound good, and is there anything else you feel is necessary? We can also expand the feature further in the future.

scalpel4k:
it's certainly less hackish than using current_file as a workaround.

However, if someone would like to handle certain error types that are not covered by the standard error_codes this isn't really the best approach I could think of. Not that I have something concrete in mind right now, just thinking ....

Maybe there could be something using inheritance or a template parameter using concepts, to make the system really customizable.

stephenberry:
Yeah, I think using a concept would be the most flexible solution. But, I also need to give it some more thought before making something too generic that it is hard for Glaze to handle custom errors and build utilities to support them.

PR #1206, opened 2024-07-23
json::rpc example with working expected error handling

arturbac:
I took me few h to determine how to handle errors in json rpc server.call so I decided to fix documentation and make actually working example. IMHO new users of glaze can benefit from it.
The question is if it is acceptable for You to include source code of example and not only brief md documentation as there are not examples at all in glaze except very detailed and complicated unit tests which are not exactly the 'examples' users usually are searching for.

- providing working source code of json::rpc example with expected error handling
- updating documentation on json-rpc


docs/rpc/json-rpc.md
```diff
@@ -6,6 +6,8 @@ Compile time specification of JSON-RPC methods making it unnecessary to convert
 
 ### Example
 
+full wrodking example can be seen in source code [examples/json-rpc.cc](examples/json-rpc.cc)
```
jbbjarnason, line 9: Spelling mistake with working

jbbjarnason (approved):
Looks good

stephenberry:
This looks great! I'm just wondering if we should by default not build the example in CMake, but rather have an option that is selected. I would just have for those who include Glaze to have another project show up that they might not care about.

Another extremely minor request: if `json-rpc.cc` could be renamed to `json_rpc.cpp` to match Glaze's naming conventions.

arturbac:
> This looks great! I'm just wondering if we should by default not build the example in CMake, but rather have an option that is selected.

I don't get it is not build by default at all .. `EXCLUDE_FROM_ALL`

stephenberry:
My bad, I missed `EXCLUDE_FROM_ALL`. Thanks!

stephenberry (approved):

arturbac:
I don't have write access to squash it.

stephenberry:
Thanks so much! I'll squash and merge.

stephenberry:
@arturbac, whenever I use `FetchContent_Declare(glaze ...)` and `target_link_libraries(${PROJECT_NAME} glaze::glaze ...)` I get the `json-rpc` project in my IDE. How can I prevent this for my own projects and others?

jbbjarnason:
> @arturbac, whenever I use `FetchContent_Declare(glaze ...)` and `target_link_libraries(${PROJECT_NAME} glaze::glaze ...)` I get the `json-rpc` project in my IDE. How can I prevent this for my own projects and others?

Maybe just move the example include line in cmake to developer mode only or separate cmake option?

stephenberry:
I added a CMake option in #1231

PR #1207, opened 2024-07-23
Add exhaustive float testing and tighten float fuzzer for denormals

pauldreik:
This adds an exhaustive test for roundtripping floats.
It takes around one cpu hour running with sanitizers, about five cpu minutes without. CI runners have four cores, so it takes a minute or two to run.

Since the floating point serializer was changed to one that roundtrips denormal numbers, I tightened the check in the existing fuzzer. It is good to lock in that improvement.

I also fixed a tiny spelling error and changed std::async to std::thread since async did not behave well on one of my machines, it got stuck for a minute or so without any reason.

### DIFF
--- a/.github/workflows/exhaustive_float.yml
+++ b/.github/workflows/exhaustive_float.yml
@@ -1,0 +1,52 @@
+name: exhaustive float json tests
+
+on:
+  push:
+    branches:
+    - main
+    - feature/*
+    paths-ignore:
+    - '**.md'
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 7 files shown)


stephenberry:
Really nice! Thanks for adding this. It is good to lock in this roundtrip behavior and nice to pass exhaustive tests!

stephenberry:
Is this ready to be merged?

pauldreik:
> Is this ready to be merged?

yes!

PR #1208, opened 2024-07-23
Redundant static def

stephenberry:


stephenberry:
Handled in another merge that updated fast_float

PR #1213, opened 2024-07-24
Contextual Sentinels

stephenberry:


include/glaze/core/read.hpp
```diff
@@ -72,12 +92,34 @@ namespace glz
...
          goto finish;
       }
-
-      if constexpr (use_padded) {
-         detail::read<Opts.format>::template op<is_padded_on<Opts>()>(value, ctx, it, end);
+      
+      if constexpr (json_read_object<T>) {
+         // Require closing `}` and use as sentinel
+         --end;
+         if (*end != '}') [[unlikely]] {
```
arturbac, line 99: if end points one past end this is UB in c++ standard.
From I see in auto read_iterators it can set end from span<Buffer> as end += buffer.size();
so such things like here
```
if (*end != '}') 
```
is UB by definition in c++ standard and buffer read overflow.

stephenberry: Notice the `--end`, so we aren't de-referencing `end += buffer.size();`

stephenberry:
Closing this contextual sentinel approach in favor of more flexible generic approach in #1262

PR #1219, opened 2024-07-25
adjust-naive_map_max_size

bryan-felipe:


stephenberry:
Bumping the max size to 128 would use 8192 bytes of memory for the underlying table. I don't think this is too much. If that 128 element structure contained 128 pointers (or uint64_t) it would be 1024 bytes large. And, we only have one `static constexpr` map for every instance of the structure it is hashing.

I think this is good, and it will result in better performance for larger tables.

Do you have any additional thoughts before I merge?

bryan-felipe:
```
struct MyStruct
{
	int m1;
	int m2;
	int m3;
	int m4;
	int m5;
	int m6;
	int m7;
	int m8;
	int m9;
	int m10;
	int m11;
	int m12;
	int m13;
	int m14;
	int m15;
	int m16;
	int m17;
	int m18;
	int m19;
	int m20;
	int m21;
	int m22;
	int m23;
	int m24;
	int m25;
	int m26;
	int m27;
	int m28;
	int m29;
	int m30;
	int m31;
	int m32;
	int m33; // error
};

void deserialize_objects(const std::string& json)
{
	std::vector<MyStruct> v;
	auto result = glz::read_json(v, json);
}

```

not at the moment, my initial idea was to inherit structs to use with std::variant, as it was not possible, I needed to create a struct with all possible members, but I got an error starting with member 33, and I went to investigate the what could it be, then I reached this constant that was preventing

bryan-felipe:
```
struct var_base 
{
	int m1;
};

struct var_a : var_base
{
	bool m2;
};

struct var_b : var_base
{
	float m3;
};

struct var_c : var_base
{
	std::string m4;
};

using var_variant = std::variant<var_base, var_a, var_b, var_c>;

void deserialize_objects(const std::string& json)
{
	std::vector<var_variant> v;
	auto result = glz::read_json(v, json);
}
```

Doing something like that isn't possible with glaze yet, right?

stephenberry:
Oh, that looks like a bug, there should be a limit of 128 elements using pure reflection, and actually no limit to the number of keys. I'll look into this.

stephenberry:
@bryan-felipe, Glaze supports variants, but not inheritance. It depends what you're trying to do with variants, but my company uses them all the time to separate options between structs.

bryan-felipe:
```
[
  {
    "m1": 1,
    "m4": "Title",
    "children": [
      {
        "m1": 2,
        "m2": true
      },
      {
        "m1": 3,
        "m3": 0.5
      }
    ]
  },
  {
    "m1": 4,
    "m2": false,
    "children": []
  }
]

struct var_base 
{
	int m1;
	std::vector<var_base> children;
};

struct var_a : var_base
{
	bool m2;
};

struct var_b : var_base
{
	float m3;
};

struct var_c : var_base
{
	std::string m4;
};
```

My idea was to read a json file and derealize linked objects, more or less like this, is there anything available for this purpose at the moment?

stephenberry:
This should be doable, but in trying to hack something together I think I ran into a Glaze bug.

Note that your `std::vector<var_base>` won't be able to deduce children because `var_base` doesn't have the higher level members. This also will leak memory because it isn't a pointer and there are no virtual destructors.

A better approach is to use a `std::vector<std::variant<var_a, var_b, var_c>>`, well almost. You see, we can't have a variant within a type that contains itself. Instead we need: 

```c++
struct var_a;
struct var_b;
struct var_c;

using variant_t = std::variant<std::shared_ptr<var_a>, std::shared_ptr<var_b>, std::shared_ptr<var_c>>;
```

This should be deducible in Glaze and valid C++. Here is a full example on compiler explorer: https://gcc.godbolt.org/z/nfbx5e3s8

However, as you can see this doesn't build right now. I think that Glaze isn't seeing through the `std::shared_ptr` type in the variant. I'll have to fix this.

stephenberry:
I'm going to merge this in, because I think the change makes sense. But, I've also opened two new issues based on what you've encountered here: #1221 and #1220

stephenberry:
@bryan-felipe, support for std::shared_ptr in std::variant has been added in #1098

PR #1222, opened 2024-07-26
add integration with oss fuzz

pauldreik:
[oss fuzz](https://google.github.io/oss-fuzz/ ) is a great project. They fuzz with different sanitizers and fuzzers with a lot of cpu power. Also, their automated handling of minimization and bug filing is very helpful.

I added integration to it mostly to be able to easily run memory sanitizer. That is useful on it's own.

If you agree, I will make a request on oss fuzz to accept glaze.

Here is a [draft branch](https://github.com/pauldreik/oss-fuzz/tree/glaze) I speculatively put your email there, let me know if you want any changes.

There is not much more work to be expected, once it works (and I expect it to, since this is header only) it  usually just runs.

### DIFF
--- a/fuzzing/ossfuzz.sh
+++ b/fuzzing/ossfuzz.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+#
+# this is the entry point for build.sh on oss fuzz, so that changes in how to
+# build the fuzzers does not require changes in oss fuzz but instead can
+# be done in the glaze repository.
+
+set -eux
+
+$CXX --version
+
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks so much for recommending oss fuzzing. I think this sounds excellent. So, go ahead and let me know what I need to do to integrate this.

pauldreik:
@pauldreik  oss-fuzz wonders if there are any prominent users of this library, would be great if you can answer there: https://github.com/google/oss-fuzz/pull/12294

stephenberry:
Thanks! I just replied on that thread.

PR #1223, opened 2024-07-27
change internal linkage on header constants to external

arturbac:
- changing linkage on header constants to external as internal causes warnings
- also fix invalid comma in example

### DIFF
--- a/docs/rpc/json-rpc.md
+++ b/docs/rpc/json-rpc.md
@@ -74,7 +74,7 @@ int main() {
     // you can assign timeout for the request in your event loop
     auto timeout = [uuid, &client]() {
         decltype(auto) map = client.get_request_map<"foo">();
-        if (map.contains(uuid));
+        if (map.contains(uuid))
             map.erase(uuid);
     };
     timeout();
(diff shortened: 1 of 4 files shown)


stephenberry:
Thanks for the fix!

However, I'm confused about the `inline constexpr` change.
As per cppreference:
> A constexpr specifier used in a function or [static](https://en.cppreference.com/w/cpp/language/static) data member(since C++17) declaration implies inline.

What compiler is giving you linker warnings?

arturbac:
> > A constexpr specifier used in a function or [static](https://en.cppreference.com/w/cpp/language/static) data member(since C++17) declaration implies inline.
> 
> What compiler is giving you linker warnings?

static data member of class is not the same as translation unit variable, they are not part of classes/structs but free header declared variables.
So AFIR constexpr variable in translation unit defaults to internal linkage not external.
And it is reported by clang
obraz

arturbac:
BTW it doesn't matter as no one is going to compare 2 same type function objects from 2 translation units with theirs addresses but it is annoying when You have in Your code warnings turned on and You see such warning from dependency being shown ..

arturbac:
my warning section on 300lines file that is using glaze is dominated by this
obraz

stephenberry:
Ah, I had read some misinformation and missed that this only applied to `data members`. There are a number of places in Glaze where I'll have to add inline, but this is a great start. Thanks so much!

PR #1247, opened 2024-08-07
String support for std::array<char, N>

stephenberry:
# Breaking change for `std::array<char, N>`
`std::array<char, ...>` is now handled as a JSON string `"abc"` rather than an array of characters `["a", "b", "c"]`


stephenberry:
@nato4r, My only concern is losing the ability to use `std::array<char, N>` as an array of characters `["a", "b", "c"]`. This is not a typical use case, but it is still within the realm of useful.

We could keep `std::array<char, N>` as an array of characters `["a", "b", "c"]` and add a `glz::stack_string<char, N>` that is treated as a JSON string with an underlying `std::array<char, N>` buffer. The downside I see for a custom Glaze class is that I would want hardware interfaces to just use `std::array` and not require a type change to a Glaze specific type.

I'm currently leaning towards changing the default behavior of `std::array<char, N>` to behave like a JSON string, and then add a Glaze wrapper that allows it to be interpreted as an array of characters in special cases.

tests/json_test/json_test.cpp
```diff
@@ -9464,6 +9464,26 @@ suite shark_variant = [] {
...
+      std::string_view value = "";
+      expect(not glz::read_json(value, R"("hello")"));
+      expect(value == "hello");
+   };
+};
+
+suite array_char_tests = [] {
+   "array<char... value"_test = [] {
+      std::array<char, 8> value{};
+      expect(not glz::read_json(value, R"("hello")"));
```
sjanel, line 9478: Open question : should a `std::array<char, N>` considered as a string with a **maximum** capacity of `N` or a string of **exactly** `N` ? For me, it would make more sense to only accept exact lengths of `N`. As an array is not a vector, it always have a fixed length, and thus it should return an error if the value length is under `N`.

include/glaze/json/read.hpp
```diff
@@ -1028,30 +1028,33 @@ namespace glz
...
-               GLZ_VALID_END();
             }
+            else if constexpr (array_char_t<T>) {
+               const size_t n = it - start;
+               if ((n + 1) > value.size()) {
+                  ctx.error = error_code::unexpected_end;
+                  return;
+               }
+               std::memcpy(value.data(), start, n);
+               value[n] = '\0';
```
sjanel, line 1054: I would actually not expect that. For me, it should behave like a string of fixed-length without the null terminating char. If I have a `std::array<char, N>`, I would expect a string of exactly N chars. This follows-up my previous comment about how I would see the semantics of the std::array<char, N> parsing case.

PR #1263, opened 2024-08-13
Binary (BEVE) support for non-null terminated buffers

stephenberry:
For BEVE handling we combine our error handling with our non-null terminated handling. So, for BEVE we don't care whether a buffer is null terminated anymore, we just always handle the data as if it is not, because we wouldn't get significant performance improvements like we do with JSON, because null character is valid binary so we can't make decisions on this value.


stephenberry:
Merged into #1262

PR #1275, opened 2024-08-17
Use FetchContent_MakeAvailable for asio and eigen

kaladron:
CMake 3.30 deprecates FetchContent_Populate as CMP0169.  FetchContent_MakeAvailable is already used in tests/CMakeLists.txt so this should not cause any problems for people on older versions of CMake.

Tested:
Built and ran tests using default container with LLVM-19 and CMake 3.30.2 on x86_64 without eigen installed on the image.


stephenberry:
@kaladron, as you can see these changes break the CI tests. Any idea of how to fix this?

stephenberry:
No, it's not you, I'm pretty sure it is Eigen

kaladron:
OK, it turns out I was seeing the failures in vscode, vscode just doesn't give ANY notice on ctests failing.  Sorry about that.  I've adjusted this so that Eigen's tests aren't recursed into.  I've used C-f to look for "Not run", and also "failed" to ensure all of them say "0 tests failed".

stephenberry:
Thanks for working on this. There are still some issues with Eigen, as you can see in the Action results:

```
CMake Error at tests/eigen_test/CMakeLists.txt:15 (set_target_properties):
  set_target_properties can not be used on an ALIAS target.
```

kaladron:
I think I just need to delete the out/ directory as a test every time before I submit. =(  I'm just running tests on a fix.

kaladron:
Please take another look.  Thank you!

kaladron:
This is driving me nuts. =(  It's midnight here, I'll look again in the morning.

stephenberry:
Thanks for pounding your head against this. If you're not able to make progress I can take a look at it. But, I really appreciate you going through this.

kaladron:
No worries.  I avoided learning cmake for so long, and I've recently switched to trying to embrace it.  It's just frustrating when it builds, you run tests and get all green checkmarks, and then it fails in presubmit! =)

Especially as this is a nothing patch to avoid a warning with a newer minimum cmake.  My actual goal is to export a modules interface which only needs cmake 3.28, not the 3.30 I thought - which is only needed for import std.  But I'll need to figure this out to submit a bigger patch, so...

stephenberry:
I've removed FetchContent_Populate with #1650, so this pull request can now be closed.

PR #1298, opened 2024-09-14
fixes min/max macro conflict in dragonbox.hpp

lambwheit:
undef original min & max definitions before redefinition in detail namespace,
use parentheses to prevent macro expansion in detail::stdr::numeric_limits<T>::max()


stephenberry:
Thanks for this! I'll submit an issue with the dragonbox repository concerning this next week. I'll merge this into main for now.

PR #1318, opened 2024-09-23
Adding BMI2 instruction set testing

stephenberry:


stephenberry:
This may be useful in the future, but for now other SWAR optimizations have shown themselves to be extremely fast and generic.

PR #1323, opened 2024-09-24
Renaming binary to beve

stephenberry:
# Major API (naming) breaking changes

> [!IMPORTANT]
>
> This is only an API (naming) change. No internal logic was changed.

## Format Specifiers
Format in `glz::opts` are now all caps. This is to disambiguate from structs with similar names.
`binary` -> `BEVE`
`json` -> `JSON`
`ndjson` -> `NDJSON`
`csv` -> `CSV`

> [!IMPORTANT]
>
> If you set the format in `glz::opts` the naming changes from `glz::opts{.format = glz::json}` to `glz::opts{.format = glz::JSON}`

## to/from specializations
For those who use custom serialization and have specialized `to_json<T>` and `from_json<T>`, these have been changed to be more generic. Use `to<JSON, T>` and `from<JSON, T>` instead.

> This will significantly reduce internal code duplication across formats.

### Binary (BEVE) name changes
Glaze used to refer to the BEVE format as `binary`. This was confusing and would become even more confusing as we add more binary formats.

Pretty much everywhere `binary` was referred to is now referred to as `beve`. So `write_binary` becomes `write_beve`.

- The file paths `glaze/binary` have been renamed to `glaze/beve`
- The header `glaze/binary.hpp` has been renamed to `glaze/beve.hpp`


stephenberry:
@arturbac, I'm making you aware of this breaking change that is coming in Glaze v3.5 as it will affect simple_enum. But, the required changes should be minor.

PR #1327, opened 2024-09-25
Reflection API

stephenberry:
# Breaking internal change for `glz::refl` -> `glz::reflect`

Formalizing the `glz::reflect<T>` API.
- Removes the inline definition of `refl` and renames `refl_info` to `reflect`
- Changes `N` to `size` for the number of fields


stephenberry:
This also adds `base64.hpp` because I made a mistake not making another branch.

PR #1343, opened 2024-09-30
SIMD support for string writing

stephenberry:
Adds SIMD support for writing strings. Currently adds AVX 256.

Adds a glaze_ENABLE_AVX2 option to CMake. By default this is on for anything but Apple. For Arm or other cross-compilation cases this needs to be turned off.


stephenberry:
@Bolderaysky, I recently merged this into main, but I haven't officially released it yet. This adds AVX2 support when available for faster string writing in some contexts.

I just want to make sure that you're aware of this change and the new option `glaze_ENABLE_AVX2` in case you are cross compiling for a system without AVX2 support.

Issue #1348, opened 2024-10-02
string_t concept should have append method instead of push_back


sjanel:
This concept currently requires that a `string_t` type has a `push_back` method:

```cpp
// this concept requires that T is string and copies the string in json
template <class T>
concept string_t = str_t<T> && !std::same_as<std::decay_t<T>, std::string_view> && has_push_back<T>;
```

but I noticed that in the code where `string_t` is used, we do not need `push_back`, but rather `resize`, `append` and / or `assign` (sometimes only `assign`, for instance when `raw_string` is `true`). I think we would need to simplify that as it makes the custom read for a custom struct hard to define.

sjanel:
I have created a test PR to illustrate the issue. It's WIP though (unit test to be added).

stephenberry:
Thanks for this. I'll try to get these updates merged soon.

PR #1349, opened 2024-10-02
[Custom json read] - Fix string_t concept (it is not based on push_back in practice)

sjanel:
Test PR to attempt to fix [this issue](https://github.com/stephenberry/glaze/issues/1348).

Basically this PR changes the `string_t` concept previously based on `has_push_back` with the methods actually used:

```cpp
// this concept requires that T is string and copies the string in json
template <class T>
concept string_t = str_t<T> && !std::same_as<std::decay_t<T>, std::string_view> && has_push_back<T>;
```
becomes:

```cpp
template <class T>
concept has_resize = requires(T t, typename T::size_type sz) { t.resize(sz); };

template <class T>
concept has_append = requires(T t, typename T::const_iterator it) { t.append(it, it); };

template <class T>
concept has_assign = requires(T t, const typename T::value_type* v, typename T::size_type sz) { t.assign(v, sz); };
   
template <class T>
concept string_t = str_t<T> && !std::same_as<std::decay_t<T>, std::string_view> &&
                                 (has_assign<T> || has_resize<T> || has_append<T>);
```

### DIFF
--- a/include/glaze/concepts/container_concepts.hpp
+++ b/include/glaze/concepts/container_concepts.hpp
@@ -25,9 +25,7 @@ namespace glz
 
    template <class T>
    concept has_empty = requires(T v) {
-      {
-         v.empty()
-      } -> std::convertible_to<bool>;
+      { v.empty() } -> std::convertible_to<bool>;
    };
 
    template <class T>
@@ -38,9 +36,7 @@ namespace glz
 
    template <class T>
    concept has_capacity = requires(T t) {
-      {
-         t.capacity()
-      } -> std::integral;
+      { t.capacity() } -> std::integral;
    };
 
    template <class T>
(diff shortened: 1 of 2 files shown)


stephenberry:
This looks great! Thank you!

fregate:
```
template <class T>
concept has_resize = requires(T t, typename T::size_type sz) { t.resize(sz); };
```
already in <glaze/concepts/container_concepts.hpp> as `concept resizable<T>`

```
   template <class T>
   concept resizable = requires(T v) { v.resize(0); };
```

looks same

stephenberry:
Thanks for pointing this out. I'll eliminate the duplication.

sjanel:
It actually also needs `.data()` with resize based on the code. Created [this PR](https://github.com/stephenberry/glaze/pull/1372)

PR #1354, opened 2024-10-03
Improve error_on_const_read logic for arrays and more direct array reading

stephenberry:


stephenberry:
Too much was going on here and this is old. Closing, and may revisit at some point.

PR #1357, opened 2024-10-03
Allow NDJSON in the skip_array requires statment.

### LINKED ISSUE #1356: glz::read for NDJSON with specifc options fails to build
Minimal reproducible example:

```C++
struct Foo  {
	int x;
};


int main ([[maybe_unused]] const int argC, [[maybe_unused]] char* argV[] ) {
	std::vector<Foo> assets{};
	const auto error = glz::read<glz::opts{.format = glz::NDJSON, .error_on_unknown_keys = false, .validate_skipped = true}>(assets, "{\"x\":1}\n{\"x\":2}");
}
```

The example fails to build (Glaze v3.6.1 and GCC 14.2) with the error (along with a lot of noise):
> glaze/include/glaze/json/skip.hpp:158:29: error: no matching function for call to ‘skip_array<glz::opts{100, '\001', '\000', '\000', '\001', '\001', '\000', '\000', ' ', 3, '\001', '\000', '\001', '\000', '\000', '\001', '\000', 0, glz::float_precision::full, '\000', '\000', '\000', '\000', '\000', '\000', '\000', '\001', '\000', '\000', '\001', '\001', 0}>(glz::context&, const char*&, const char*&)’

I don't see why the combination of options should be invalid, so I assume this is a bug. I guess the reason is the requires statement in skip.hpp:63 wrongly excluding NDJSON. See the accompanying pull request.

Haatschii:
This hotfix fixes #1356. However I am not sure this is way to solve it, so please feel free to reject the pull request if there is a better solution.

### DIFF
--- a/include/glaze/json/skip.hpp
+++ b/include/glaze/json/skip.hpp
@@ -60,7 +60,7 @@ namespace glz::detail
    }
 
    template <opts Opts>
-      requires(Opts.format == JSON)
+      requires(Opts.format == JSON || Opts.format == NDJSON)
    void skip_array(is_context auto&& ctx, auto&& it, auto&& end) noexcept
    {
       if constexpr (!Opts.validate_skipped) {


stephenberry:
This looks like a good change. I'll merge it in.

PR #1372, opened 2024-10-09
[Concept update] string_t with resize also needs to have .data()

sjanel:
`resize()` is used with `data()`. Update the concept requirements for `string_t`.
Also clang-formatted the code according to the `.clang-format` file

### DIFF
--- a/include/glaze/concepts/container_concepts.hpp
+++ b/include/glaze/concepts/container_concepts.hpp
@@ -25,9 +25,7 @@ namespace glz
 
    template <class T>
    concept has_empty = requires(T v) {
-      {
-         v.empty()
-      } -> std::convertible_to<bool>;
+      { v.empty() } -> std::convertible_to<bool>;
    };
 
    template <class T>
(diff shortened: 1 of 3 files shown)


stephenberry:
@sjanel, I prefer your formatting for concepts. Do you know how to update the clang format in Glaze to get this cleaner formatting?

stephenberry:
Thanks for this!

sjanel:
> @sjanel, I prefer your formatting for concepts. Do you know how to update the clang format in Glaze to get this cleaner formatting?

I use clang-format v18, based on the `.clang-format` file of `glaze` repository. Maybe your clang-format version is different ? What we could do to enforce formatting of files in the repo is to use [this github action](https://github.com/jidicula/clang-format-action) and add a new workflow whose aim is simply to fail if the files are not formatted with the desired clang-format version + clang-format file.

stephenberry:
Ah, I think I just need to do a version bump on the clang-format action runner. `main` is always auto-formatted, Glaze is just using an old version.

PR #1385, opened 2024-10-16
Fix JSON RPC documentation link to json-rpc.cpp source

pentatonick:
Fix broken link to json-rpc.cpp example in the docs.


stephenberry:
Thanks!

PR #1392, opened 2024-10-21
Use jkj::dragonbox::to_decimal_ex

stephenberry:
Supersedes #1316


stephenberry:
@jk-jeon, Glaze will now be using `jkj::dragonbox::to_decimal_ex`. This change gets it working. Thanks for your awesome library!

PR #1394, opened 2024-10-22
Fix json parsing from arrays to set like types with custom structs

sjanel:
Hello,

Several things in this PR:

 - I noticed a bug in the parsing of json arrays towards set like types. It seems that a skip of whitespaces is needed (`GLZ_SKIP_WS();`). It is done correctly for `vector` like types below. I tested the fix on my custom struct and it works fine. I could not figure it out with a unit test though (I don't know how new tests pass without the fix in `read.json`).
 - Could you please help me figuring out why the commented code added in this PR (in `json_test.cpp` does not compile? It seems that the `to` JSON custom function excepts this signature instead (it works with below code, however, `value` is const so it's not possible to apped anything to it):
 
 ```
      template <auto Opts, is_context Ctx, class B, class IX>
      static void op(auto&& value, Ctx&& c, B&& b, IX&& ix)
      {
         write<JSON>::op<Opts>(value.str, c, b, ix);
      }
```
Thanks a lot in advance.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -1250,6 +1250,7 @@ namespace glz
                   return;
                }
                GLZ_MATCH_COMMA;
+               GLZ_SKIP_WS();
             }
          }
       };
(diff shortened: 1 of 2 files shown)


stephenberry:
@sjanel, thanks for submitting this pull request. I'll look into it ASAP, though it may be a few days until I have the time.

stephenberry:
@sjanel, I was able to look at the issue and thankfully its simple. For writing we have to iterate through the container, however `std::set` iterators (even from non-const containers) are const qualified (see https://en.cppreference.com/w/cpp/container/set/begin). This means that our custom serializer must take in a const reference: `const custom_struct& value`.

This means that you cannot mutate the set elements on writing. If you think about it, it makes sense, because mutating the values would need to change their location in the set. So, a std::set is probably the wrong type if you're needing to mutate on write.

I've updated the unit test to build and pass. I'll merge in this fix, thanks so much!

PR #1398, opened 2024-10-23
Put fast_float and dragonbox into glz namespace

amosbird:
Move fast_float and dragonbox libraries under glz namespace to prevent version conflicts when projects using different versions of these dependencies are linked together. This change improves compatibility and reduces potential linking issues.


stephenberry:
Thanks for this, it's a smart change.

Issue #1410, opened 2024-10-26
can't skip null on custom lambda


JimjunZhang:
Hello,
I am using custom Lambda to output a field or skip based on a condition, but I found it can't skip the null.

sample code:
```
struct my_struct
{
    int i = 287;
    double d = 3.14;
   
};
template <>
struct glz::meta<my_struct> {
    using T = my_struct;
    static constexpr auto read_test = [](T& s, const std::string& input) {};
    static constexpr auto write_test =
        [](auto& s) ->std::unique_ptr<std::string> {
        if (287 != s.i)
            return std::make_unique<std::string>("expected: not 278");
        else
            return std::unique_ptr<std::string>();
        };
    static constexpr auto value = object(
        "i", & T::i,
        "d", & T::d,
        "test", glz::custom<read_test, write_test>
    );
};
```

Use it:
```
my_struct obj;
std::string buffer;
auto ec = glz::write < glz::opts{ .prettify = true } > (obj, buffer);
```
The output is:
{
   "i": 287,
   "d": 3.14,
   "test": null
}
But I would expect the null can be skipped.  Is it to add this feature for custom lambda?  thanks!

stephenberry:
Thanks for reporting this issue. Will look into it soon.

stephenberry:
So, this will require a significant amount of compile time programming, because we need to deduce the type of the return of the custom function. If the type is nullable then we need to decode the value within the object reading code and determine if the value is null so that we can skip writing out keys.

Because this fix would affect object serialization code, I'm going to spend some time considering this issue and tackle it when I have a good amount of time to determine a flexible solution. So, I would expect it to be a while until this feature is added. But, thanks for bringing this up!

stephenberry:
This issue was inadvertently fixed in commit 20e53e56 on December 13, 2024 ("Mustache improvements and moving get_member out of detail namespace").

The key change was in glaze/core/reflect.hpp:130:

```c++
template <size_t I>
using type = member_t<V, decltype(get<I>(values))>;
```

By using member_t, the field_t<T, I> type alias now correctly resolves to the lambda's return type rather than the lambda type itself. This allows the existing null_t concept checking to work correctly for lambda return values.

Verification tests have been added in: https://github.com/stephenberry/glaze/pull/2059

PR #1438, opened 2024-11-11
implement skip writing out default

huangminghuang:
# Motivation

The primary goal of this feature is to mimic protobuf behavior, where fields with default values are omitted from the wire format. Additionally, a new printout option allows users to override this behavior if they wish to include fields with default values. To support this flexibility, the `skip_null_members flag` in Opts has been redefined as a bitmask, allowing independent control over skipping null and default values.

# Key Changes
- Conditional Skipping with `custom` Wrapper: The `custom` wrapper now enables users to define a function that determines whether a field should be omitted based on specific conditions. An associated bitmask has been introduced to allow compile-time checks, which control whether user-defined functions should be evaluated based on the `Opts::skip_null_members` setting during serialization.
- New `skip_write_default` Wrapper: A `skip_write_default` wrapper provides a concise way to skip fields with default values, streamlining the setup for this common use case.

# Note
This feature can be used to resolve IS #1410

### DIFF
--- a/include/glaze/core/wrappers.hpp
+++ b/include/glaze/core/wrappers.hpp
@@ -39,26 +39,51 @@ namespace glz
          };
       }
 
+      struct always_write_t
+      {
+         bool operator()() const { return true; }
+      };
+
       // custom_t allows a user to register member functions (and std::function members) to implement custom reading and
       // writing
-      template <class T, class From, class To>
+      template <class T, class From, class To, class Skippable, class SkipMask>
       struct custom_t final
       {
          static constexpr auto glaze_reflect = false;
+         static constexpr auto glaze_wrapper = true;
+         static constexpr auto glaze_skip_write_mask = SkipMask::value;
          using from_t = From;
          using to_t = To;
          T& val;
          From from;
          To to;
+         constexpr bool write_skippable() const { return Skippable()(val); }
       };
 
-      template <class T, class From, class To>
-      custom_t(T&, From, To) -> custom_t<T, From, To>;
-
-      template <auto From, auto To>
+      template <auto From, auto To, auto WriteSkippable, uint8_t SkipMask>
       inline constexpr auto custom_impl() noexcept
       {
-         return [](auto&& v) { return custom_t{v, From, To}; };
+         return [](auto&& v) {
+            using skippable_t = decltype(WriteSkippable);
+            using skip_mask_t =
+               std::conditional_t<std::is_same_v<skippable_t, always_write_t>, std::integral_constant<uint8_t, 0>,
+                                  std::integral_constant<uint8_t, SkipMask>>;
+            return custom_t<std::remove_reference_t<decltype(v)>, decltype(From), decltype(To), skippable_t,
+                            skip_mask_t>{v, From, To};
+         };
+      }
+
+      struct deduct_default_t {};
+
+      template <auto MemPtr, auto Default, class T>
+      bool is_default(const T& val)
+      {
+         if constexpr (std::same_as<decltype(Default), deduct_default_t>) {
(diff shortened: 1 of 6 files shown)


huangminghuang:
I am glad you like it. However, from the [github action run](https://github.com/stephenberry/glaze/actions/runs/11784241534/job/32823028575) , there is a race condition in repe_test which I don't think is introduced by my change. You may need to tackle it.

stephenberry:
@huangminghuang, yes I saw that race condition recently and need to tackle it.

stephenberry:
One other aspect of this pull request that we might want to change is the behavior of the `skip_null_members` options when the `skip_default_flag` is set. Currently in Glaze the `skip_null_members` option is globally applied, which means that you don't have to use `glz::custom` to get the behavior and you can use it with reflected types that have no `glz::meta`. I feel like we want the `skip_default_flag` to work the same way, so as to not require a `glz::meta` nor require `glz::custom`. I do like the fine grained approach of being able to customize the behavior for each field, but often this is desired as a global effect for the `glz::write` call.

This requires some more work, but I think it will be worth it in the long run. If you don't feel comfortable tackling this, I can work on it when I have time. But, it would be great if you were able to mimic the behavior of the other options and get global behavior working.

include/glaze/core/wrappers.hpp
```diff
@@ -71,14 +71,16 @@ namespace glz
...
+      // Used for determining the default value
       struct deduct_default_t
       {};
 
       template <auto MemPtr, auto Default, class T>
       bool is_default(const T& val)
       {
          if constexpr (std::same_as<std::decay_t<decltype(Default)>, deduct_default_t>) {
-            return val.*MemPtr == T{}.*MemPtr;
+            static thread_local auto defaulted = T{};
```
huangminghuang, line 82: I believe the `thread_local` is unnecessary because C++ requires static variable initialization to be thread safe since C++11. The statement might be better as follows.
```
static const auto defaulted = T{}.*MemPtr;
return val.*MemPtr == defaulted;
```
I expected the compiler can be smart enough not to construct unused data members this way; however, based on my experiment on compiler explorer, no compiler can do that.

huangminghuang:
> One other aspect of this pull request that we might want to change is the behavior of the `skip_null_members` options when the `skip_default_flag` is set. Currently in Glaze the `skip_null_members` option is globally applied, which means that you don't have to use `glz::custom` to get the behavior and you can use it with reflected types that have no `glz::meta`. I feel like we want the `skip_default_flag` to work the same way, so as to not require a `glz::meta` nor require `glz::custom`. I do like the fine grained approach of being able to customize the behavior for each field, but often this is desired as a global effect for the `glz::write` call.
> 

I think we should consider the semantic carefully about this. The current design is to use `glz::custom` to declare a field has a default value, and use the `glz::opts` to control whether we should skip the default. If I understand you right, you want that every field has a default value implicitly and can be skipped using `skip_default_flag`. In the case, you may want a way to declare a field cannot be skipped anyway; use `glz::custom` with `skip_mask=0` may work. However,  I feel we need two different flags: `skip_custom_default_flag` and `skip_implicit_default_flag` to differentiate between the two scenarios.

include/glaze/core/wrappers.hpp
stephenberry: You are correct, because we aren't changing state, we can have a single, global static variable. I'm just used to requiring `thread_local` for mutation. We can also make it `const`.

stephenberry:
I agree, any general option for skipping defaults would need an opt-out option, like you've noted with glz::custom.

stephenberry:
This is now supported with the `skip_default_members = true` compile time option

PR #1444, opened 2024-11-20
Asio improvements

stephenberry:
Adding a `run_async` option for the `glz::asio_server` and simplifying the `stop()` method (removing unnecessary thread).

`glz::asio_client::init()` now returns a `repe::error_t` for consistency

Removed exceptions from `send_buffer` and `receive_buffer`

Also fixes/hides a number of MSVC warnings.


stephenberry:
Note that MSVC build is broken due to an MSVC bug in 17.12.1 that I reported and was fixed last night, but has not been released yet.

PR #1446, opened 2024-11-21
Adding currency csv test

stephenberry:
Handling for escaped string reading (at least for column wise CSV)


stephenberry:
Note that MSVC build is broken due to an MSVC bug in 17.12.1 that I reported and was fixed last night, but has not been released yet.

PR #1449, opened 2024-11-22
Simplifying JSON read code

stephenberry:
## Breaking change for escaped unicode keys

Removed the option `escaped_unicode_key_conversion`, as it added significant complexity to the underlying code and required inefficient parsing due to heap allocations and resulted in inefficient hashing. Removing this option does not break any JSON conformance.

If you require both unescaped and escaped versions for a key, then register both of these keys with the same member pointer within `glz::meta`. This gives far better performance, but should still generally be avoided.

Added "glaze/json/escape_unicode.hpp", which provides `glz::escape_unicode`. This results in a compile time escaped version of the input string.

An example:
```c++
template <>
struct glz::meta<unicode_escaped_t>
{
   using T = unicode_escaped_t;
   static constexpr auto value = object("😀", &T::smile, escape_unicode<"😀">, &T::smile);
};
```


include/glaze/util/string_literal.hpp
```diff
@@ -30,7 +30,11 @@ namespace glz
       constexpr string_literal& operator=(const string_literal&) noexcept = default;
       constexpr string_literal& operator=(string_literal&&) noexcept = default;
 
-      constexpr string_literal(const char (&str)[N]) noexcept { std::copy_n(str, N, value); }
+      constexpr string_literal(const char (&str)[N]) noexcept {
+         for (size_t i = 0; i < N; ++i) {
```
sjanel, line 34: This is to avoid include of `<algorithm>` ?

stephenberry: @sjanel, yes I'm slowly trying to simplify large header includes where they aren't really necessary.

PR #1453, opened 2024-11-23
don't use _umul128, fallback to umul128_generic in MSYS2/MINGW32

jalius:
Currently seeing build errors for glaze on Windows 10/11 with MSYS2, UCRT64, GCC 14.1.0 :

> C:\Users\Jalius\repos\glaze\src>g++ main.cpp --std=c++23 -I../include
> In file included from ../include/glaze/util/parse.hpp:15,
> (TRUNCATED)
>                  from main.cpp:1:
> ../include/glaze/util/atoi.hpp: In function 'constexpr glz::detail::value128 glz::detail::full_multiplication(uint64_t, uint64_t)':
> ../include/glaze/util/atoi.hpp:174:20: error: '_umul128' was not declared in this scope
>   174 |       answer.low = _umul128(a, b, &answer.high); // _umul128 not available on ARM64
>       |                    ^~~~~~~~
> 

Failling back to`umul128_generic` conditionally based on `__MINGW32__` resolves this.

### DIFF
--- a/include/glaze/util/atoi.hpp
+++ b/include/glaze/util/atoi.hpp
@@ -170,7 +170,7 @@ namespace glz::detail
       // But MinGW on ARM64 doesn't have native support for 64-bit multiplications
       answer.high = __umulh(a, b);
       answer.low = a * b;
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for this fix!

PR #1460, opened 2024-12-02
Support for nullable value types

stephenberry:
This adds support for optional like types that cannot provide an `operator bool()` because that operator is reserved for some other action. This adds a `nullable_value_t` concept that only checks for `value()` and `has_value()`.

### DIFF
--- a/docs/nullable-types.md
+++ b/docs/nullable-types.md
@@ -1,0 +1,32 @@
+# Nullable Types
+
+Glaze provides concepts to support custom nullable types. In order to serialize and parse your custom nullable type it should conform to the `nullable_t` concept:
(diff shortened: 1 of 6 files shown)


scalpel4k:
Nice, this will do the trick for me. Should be mentioned somewhere in the docs, I think.

thanks a lot!

scalpel4k:
I've tested it with our code base and it works as intended

stephenberry:
Great! I'm about to merge this. I added support for reading nullable value types as well, as long as `reset()` and `emplace()` methods are provided. If you don't have these methods on your type then you can just keep using your custom parse specialization.

PR #1468, opened 2024-12-03
Make glz::async_map copyable and movable

stephenberry:
DO NOT MERGE YET. MUST DETERMINE IF THE MAP ITSELF SHOULD BE SHARED OR JUST THE ELEMENTS.

```c++
// The async_map is copyable, but it only performs shallow copies
// This is to allow values like `std::atomic<int>` (and other thread safe types) to be held
// So, copying an async_map will result in a shared reference to the data
// In a multi-threading context this is often the desired behavior as thread safe logic
// is typically attempting to shared state across threads
```

### DIFF
--- a/include/glaze/thread/async_map.hpp
+++ b/include/glaze/thread/async_map.hpp
@@ -23,7 +23,7 @@ namespace glz
    struct async_map
    {
      private:
-      std::vector<std::unique_ptr<std::pair<K, V>>> items;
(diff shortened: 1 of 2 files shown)


stephenberry:
Decided that the state should internally be shared, because the point of this map is to share data across threads. This change was made so that the class is now `glz::shared_async_map`, hence closing this pull request.

Issue #1503, opened 2024-12-16
problems inheriting from `glz::json_t`


benine203:
Illustrative testcase
=====================
```cpp
#include <cassert>
#include <print>

#include "glaze/glaze.hpp"

struct GLZ : public glz::json_t
{
   GLZ() = default;
   GLZ(GLZ const& other) : glz::json_t(other) {}
   GLZ(GLZ&& other) : glz::json_t(std::move(other)) {}

   operator double() const
   {
      if (!is_number())
         throw std::runtime_error(std::format("{}: cannot convert {} to double", __PRETTY_FUNCTION__, dump().value()));
      return 0.0;
   }
};

int main()
{
   GLZ glz;
   glz["key"] = 3.14;
   glz["abc"] = "xyz";

   assert((std::is_convertible<GLZ, glz::json_t::val_t>::value));
   assert((!std::same_as<glz::json_t, GLZ>));

   GLZ glz2{std::move(glz)};

   std::print("copy : {}\n", glz2.dump().value());
}

```

Ouput
=====
```
~/src/glaze/build$ c++ -g -ggdb3 -std=gnu++23 -I../include tst2.cc 
.
.         ^
1 warning generated.
~/src/glaze/build$ ./a.out 
terminate called after throwing an instance of 'std::runtime_error'
  what():  double GLZ::operator double() const: cannot convert {"abc":true,"key":3.14} to double
Aborted
```

With libc++
```
libc++abi: terminating due to uncaught exception of type std::runtime_error: double GLZ::operator double() const: cannot convert {"abc":true,"key":3.14} to double
Aborted
```

Tracing the crash
=================
Following a backtrace leads to [this code](https://github.com/stephenberry/glaze/blob/a62bc98dcc14a2eb40d6f328f26b8bd3cba977d5/include/glaze/json/json_t.hpp#L172-L177) being called upon invocation of copy constructor.

`val_t` is a variant, and the checks 
```
 template <class T>
    requires std::convertible_to<T, val_t> &&
             (!std::same_as<json_t, std::decay_t<T>>)
  }
 ```
 

 allow it to go through, with T being the GLZ class.
 
 Solution proposed
 =================
 his patch proposes to weaken the check to `std::derived_from` instead and allows for derived classes providing conversion operators

 ```diff
 diff --git a/include/glaze/json/json_t.hpp b/include/glaze/json/json_t.hpp
index 696221ef..660c718e 100644
--- a/include/glaze/json/json_t.hpp
[...]

PR #1504, opened 2024-12-16
allow subclassing json_t

benine203:
corresponds to issue #1503

### DIFF
--- a/include/glaze/json/json_t.hpp
+++ b/include/glaze/json/json_t.hpp
@@ -25,6 +25,8 @@ namespace glz
    // Generic json type.
    struct json_t
    {
+      virtual ~json_t() {}
+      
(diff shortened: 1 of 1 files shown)


stephenberry:
@benine203, thanks for this pull request. What do you think about adding a virtual destructor to `glz::json_t` if we are going to allow users to inherit from it?

benine203:
> @benine203, thanks for this pull request. What do you think about adding a virtual destructor to `glz::json_t` if we are going to allow users to inherit from it?

great, even better right now I just do a lot of static typecasting etc... most of the code is on a repo account

stephenberry:
Cool, I'll merge this as soon as everything passes. I added the virtual destructor.

PR #1515, opened 2024-12-20
Removing make_static

stephenberry:
This can done done when GCC 12 support is deprecated


stephenberry:
We'll revisit this eventually, but it isn't critical, so I'm closing the pull request.

PR #1518, opened 2024-12-20
Remove constexpr requirement for glz::schema fields & cleaned JSON schema code

stephenberry:


stephenberry:
@jbbjarnason, since you helped so much with this, I wanted to let you know that I've dropped the constexpr construction requirement on `glz::schema`, this means that we can now use `std::vector` rather than `std::span` and it makes including examples and other unknown array length fields easier.

stephenberry:
Hmm, I'm not sure why GCC is failing this when clang works

PR #1525, opened 2024-12-26
[Json] - Fix enum string read for vector of pairs in concatenate mode

sjanel:
This PR fixes an `expected_quote` parsing error occurring for objects of type `vector<std::pair<MyEnum, ...>` when in concatenate mode and if user requests parsing of enums as strings.

This PR takes the assumption that 

```
std::is_enum_v<typename V::first_type> && detail::glaze_t<typename V::first_type>
```

is `true` **if and only if** the enum needs to be serialized / de-serialized as strings instead of values (based on my findings). Please help me correct the code if this assumption is wrong.

Updated the unit tests accordingly.

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -968,7 +968,7 @@ namespace glz
          }
 
          // for types like std::vector<std::pair...> that can't look up with operator[]
-         // Intead of hashing or linear searching, we just clear the input and overwrite the entire contents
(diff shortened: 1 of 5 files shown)


stephenberry:
Thanks for this pull request! I'll try to look it over soon.

stephenberry:
I think this makes sense, it allows for quoted enum parsing in vectors of pairs. I'll merge it in. Thanks!

PR #1555, opened 2025-01-07
add support for Eigen::Transform

WanZhiQiu-ac:


WanZhiQiu-ac:
> I don't think this is a Glaze issue, because it is robustly tested for floating point round-tripping (every possible float is tested and random doubles). This is probably a precision issue with the way floating point assembly is generated for Eigen. Floating point math operations with Eigen aren't guaranteed to give the exact same result across compilers.

Totally agree

WanZhiQiu-ac:
And yes I think it is ready to be merged

stephenberry:
Sweet! Merged it in.

Issue #1558, opened 2025-01-10
Template type error with some 32bit architectures


20162026:
## issue
on some 32bit architectures int32_t might not be based of int, cousing a compiler error at https://github.com/stephenberry/glaze/blob/4489f051cd68896f6592bcd02d759ea1ee894f2d/include/glaze/json/jmespath.hpp#L477

````
 return std::clamp(idx, 0, size);  <<< int32_t, int, int32_t  TYPE ERROR
````

xtensa gcc 12.2.0
https://godbolt.org/z/EhjMdP59G
arm gcc 14.2.0
https://godbolt.org/z/MMzTjb3jo

## proposal
using `int32_t{0}` instead of `0` should fix the issue and not cause any unexpected behaviour

stephenberry:
I merge in your pull request, thanks!

PR #1559, opened 2025-01-10
fix template type error

20162026:
fix template type error (#1558)


stephenberry:
Thanks for this fix! I'll merge it in.

PR #1575, opened 2025-01-24
reflect: fix uint16 compile error in find_unique_sized_index

timsjostrand:
The left-shift operator causes integer promotion which results in compiler errors on msvc with all warnings enabled.

This is just a quick fix, it might be possible to do the cast on the right hand side of the OR?


stephenberry:
I think you're correct and the cast can just be on the right side. But, this looks fine so I'll merge it in. Thanks for the fix!

PR #1590, opened 2025-02-03
fix #1588

### LINKED ISSUE #1588: Setting compiler flags in user project should be in domain of user not glaze - AVX2 injection into user project configuration
I am working on #1566 trying to figure out more and I found that
1)
CPM fails disabling flag where enabling usually works and I am not sure where is the bug with that
I simply can not disable AVX on my builds using cpm even when I hard try to do so ..
```cmake
set(glaze_ENABLE_AVX2 OFF CACHE BOOL "Disable AVX2 support")

CPMAddPackage(
    glaze
    GITHUB_REPOSITORY stephenberry/glaze
    # GIT_TAG        v3.2.6
    # GIT_TAG        v4.2.2
    GIT_TAG main
     OPTIONS "glaze_ENABLE_AVX2 OFF"
    )
```
got
```
-- CPM: Adding package glaze@ (main)
Glaze is being compiled with AVX2 instructions. Set glaze_ENABLE_AVX2 to OFF if you wish to disable this.
```

2) You are setting compiler flags injecting them into user projects in cmake/avx.cmake with interface compile flags on glaze. This is IMHO bad practice affecting user code build flags instead of own flags. Since glaze is interface it should  not set any compiler flags at all.
More over You don't need ```target_compile_definitions(glaze_glaze INTERFACE GLZ_USE_AVX2)``` at all.
It should be detected for x86_64 at compile time checking 
```cpp
#if defined(__x86_64__) || defined(_M_X64)
    #if defined(_MSC_VER)
        #include <intrin.h>
    #else
        #include <immintrin.h>
    #endif

    #if defined(__AVX2__)
        #define GLZ_USE_AVX2
    #endif
#endif
```
leaving the user to decide on his compiler flags and setting -march=native or -mavx2 ...
It is common for my company to build code on arch supporting avx2 code but doing builds for server which don't support avx2. Or to make builds targeting different platform cross compilation, so hosts ability to run avx2 means nothing at all.

So I would ask You to at least disable glaze_ENABLE_AVX2 by default at least.

arturbac:


stephenberry:
This avoids adding compiler flags to the CMake configuration, giving developers proper flexibility.

stephenberry:
@arturbac, thanks for this pull request. I deleted the old `avx.cmake` and make some minor changes. Do you think in the long run we'll want separate options for different instruction sets, such as what I've named `glaze_DISABLE_AVX2_WHEN_SUPPORTED`, or should Glaze just use a catch all `glaze_DISABLE_SIMD_WHEN_SUPPORTED`? I'm thinking we might end up with a few different options for different instruction sets in the long run, except that your recent suggestion of using built in vectorization types might make us just want one global option.

stephenberry:
There are places in Glaze where I know I can optimize code using BMI2 instructions that are not generated from generic vectorized code. But, I could add that as a separate CMake option. I think I'm going to follow your advice and use `glaze_DISABLE_SIMD_WHEN_SUPPORTED` for general SIMD instruction sets.

arturbac:
In my production code I was using approach hiding explicit intrinsic calls with forced inline functions written for generic, x86_64 variants an aarch64
So in global code everything was generic and not infected with any explicit intrinsic calls, more readable and explicit wrappers over intrinsic sets were able to be testable.
So in case of BMI2 i wold not use them directly but create 3 wrapper functions
[bmi2, ararch64 neon, generic c++ code] 

examples:
```cpp
template<int... args, typename vector_type>
constexpr vector_type shuffle_vector(vector_type a, vector_type b) noexcept
  {
#if defined(__clang__)
  return __builtin_shufflevector(a, b, args...);
#else
  using element_type = typename std::remove_reference<typename std::remove_cv<decltype(a[0])>::type>::type;
  return __builtin_shuffle(a, b, vector_type{static_cast<element_type>(args)...});
#endif
  }
```

or
```cpp
#if defined(__ARM_NEON)
[[nodiscard, gnu::always_inline, gnu::const]]
inline float64x2_t max_pd(float64x2_t a, float64x2_t b)
  {
  return vpmaxq_f64(a, b);
  }
#elif defined(__SSE2__)
using float64x2_t = __m128d;

[[nodiscard, gnu::always_inline, gnu::const]]
inline float64x2_t max_pd(float64x2_t a, float64x2_t b) noexcept
  {
  return _mm_max_pd(a, b);
  }
 #else 
 [[nodiscard, gnu::always_inline, gnu::const]]
inline float64x2_t max_pd(float64x2_t a, float64x2_t b) noexcept
{ 
return 
```
and that way i was able to avoid using direct intrinsic

stephenberry:
That's a good approach.

PR #1595, opened 2025-02-05
Add glz::sync

stephenberry:
The purpose of glz::sync is to create a thread-safe wrapper around a type. The only way to access the data is by supplying lambdas to `read` or `write` methods, which feed underlying data into the lambda. A lock is held for the duration of the call. This makes it easy to maintain thread safety without managing a mutex manually and allows the data to still be copied/moved.

```c++
struct foo { int x{}; };
glz::sync<foo> s{};
s.write([](auto& value) { value.x = 42; });
s.read([](const auto& value) { std::cout << value.x << '\n'; });
```

### DIFF
--- a/include/glaze/thread/sync.hpp
+++ b/include/glaze/thread/sync.hpp
@@ -1,0 +1,121 @@
+// Glaze Library
+// For the license information refer to glaze.hpp
+
+#pragma once
+
+#include <concepts>
+#include <mutex>
+#include <shared_mutex>
+#include <type_traits>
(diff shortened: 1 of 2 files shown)


stephenberry:
The problem with this approach is when you need to handle multiple `glz::sync` types at once.

I want to have a `glz::sync_read([](auto&&... args){}, sync1, sync2)` and `glz::sync_write([](const auto&... args){}, sync1, sync2)`, where `args...` can be various `sync` types.

Issue #1566, opened 2025-01-14
clang inlining optimisation issue with pure release static build in json parsing - possibly null termination code change caused it


arturbac:
In one of my projects when i tried to upgrade version from v3.2.5 i got an error in system tests.
I will point out everything I have found so far to narrow down it.
it affects only static linked release build, on asan/ubsan instrumented and debug sahred builds it is ok.

3.2.6 - everything was still so far ok.
3.3.0 - release of non null termianted support ~7 errors in release build of >1000 tests
in later versions up to 4.2.2 there is still one error in my system tests but it is not possible to make repro as repro json works.
error is deterministic, always the same.

I tried parsing with null termination, without, even creating resize string buffer twice the size to ensure 0.
In below test You can see that I am outputing json string for which parsing fails so this makes sure problem is not in my code with invalid json data.

```cpp
template<glz::read_json_supported T, class Buffer>
[[nodiscard]]
expected<T, glz::error_ctx> read_json(Buffer && buffer) noexcept
  {
  T value{};
  glz::context ctx{};
  glz::error_ctx const ec = read<glz::opts{.null_terminated = true}>(value, std::forward<Buffer>(buffer), ctx);
  if(ec)
    return unexpected<glz::error_ctx>(ec);
  return value;
  }

std::string data{response.jsonrpc};
    size_t orgsize{data.size()};
    data.resize(orgsize*2);
    data.resize(orgsize);
    auto geocode_res{read_json<json_rpc_response>(std::move(data))};
    namespace ut = boost::ut;

    if(!geocode_res)
      {
      ut::expect(false) << [&]
      {
        return std::format(
          "failed to decode json rpc v2 response for :{} msg:{} glz ec:{}:{} loc:{}\njson:\n{}", 
          tu.test_name,
          geocode_res.error().custom_error_message,
          int(geocode_res.error().ec),
          geocode_res.error().ec,
          geocode_res.error().location,
          response.jsonrpc
        );
      };
      return false;
      }
    json_rpc_response result{std::move(geocode_res.value())};
```
which produces when run in batch
```bash
Running "1348_GeeQD: Domaniewska 37"...
  verify.h:140:FAILED [false] failed to decode json rpc v2 response for :1348_GeeQD: Domaniewska 37 msg: glz ec:18:expected_comma loc:153
json:
[...]

PR #1597, opened 2025-02-06
Straightforward boolean write

stephenberry:
This approach avoids UB issues if developers forget to instantiate booleans. See discussion at the end of #1566


stephenberry:
@arturbac, here's the safer code I'll merge in soon.

PR #1598, opened 2025-02-06
Separate out opts

stephenberry:


stephenberry:
I don't think the inheritance approach is right, but this was an experiment.

stephenberry:
This attempt was superseded by work that was released in v5.0.0

PR #1599, opened 2025-02-07
Modularizing (#1586)

msqr1:
More to be added, this is just a draft one. Modularized with [importizer](https://github.com/msqr1/importizer), code is not reviewed, fresh after running it through.
Regular build works perfectly, you can use the same command to build and test with zero changes. Working on exporting and writing the CMake portion for modules.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ include(cmake/prelude.cmake)
 
 project(
     glaze
-    VERSION 5.0.0
+    VERSION 5.0.1
     LANGUAGES CXX
 )
 


msqr1:
@stephenberry Do you have any thoughts on the code before I move on to actually supporting modules? Also do you want to try [the new module structure](https://github.com/msqr1/importizer/issues/12) or the current one is OK?

stephenberry:
Can you explain a little bit more the new approach? Does this have to do with internal macros to the library? Glaze does have some macros that are used across the library.

Note that I noticed some weird handling of macros that are locally defined. See the modularization of `glaze/api/trait.hpp`.

```c++
module;
#include <array>
#define std_trait(x) static constexpr sv x = to_sv<std::x##_v<T>>()
#undef std_trait
export module glaze.api.trait;
```

This `std_trait` macro is defined and undefined within the same header. But, your module code adds it to the top in a manner that does nothing.

stephenberry:
Thanks for your explanation. It sounds like the new approach is better. You'll probably be dropping the old approach, right? In which case I think it's better to move to the new.

Yeah, I realize this is auto-generated from importizer. So, I figured I'd point out things that you can improve/fix. The #define/#undef is probably the best use of macros because it doesn't pollute anything, but can locally shorten code.

msqr1:
I will work on those, and let's see how it goes.

msqr1:
I implemented the shared code section as well as the #define/#undef removal. @stephenberry can you check it out?

msqr1:
@stephenberry can you check it out? I implemented the shared code section as well as the #define/#undef removal.
Let me know if you are ready for exporting and cleaning up!

stephenberry:
This is looking good. We should be able to add to the GitHub Actions and test building with modules.
I'm also curious about your thoughts on `import std`? What do you think needs to happen with compilers before it can be used?
Thanks!

msqr1:
Std module status:
GCC: [Incomplete](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2023:~:text=Standard%20Library%20Modules%20std%20and%20std.compat)
Clang: 19
MSVC: 14.38

If you want to use `import std` right now to speed up compilation, we have to drop GCC. The best way right now to do use standard library as a module is bootstrapping it. By that I mean something like this:
```cpp
module;
#include <vector>
// Other standard includes
export module stdBootstrap;
using namespace std;
export namespace std {
  using vector;
  // using ... 
}
```
We get better compile time this way. We can swap out for compilers' standard modules once GCC is ready with some Ctrl+F. And you might have guessed it, I have a setting to do just this...

msqr1:
> An additional thought is that you might want to look into a way of handling git renaming so that the pull request doesn't show every line of every file as having changed. Git has methods to rename files so that changing to `.cppm` doesn't look like the entire file changed.

I will get to this at the very end. Are you ready for cleaning? Any last fixes?

stephenberry:
I don't see any problems, so I say keep going ahead!

msqr1:
Um, just a heads up, this is going to take quite long, so be patient! I need to export, move the licence comments up, and do lots of stuff.

msqr1:
@stephenberry One more thing I forgot, glaze seems to have some headers depending on macros from other headers. While this would work in header build, it won't be in modules because modules don't leak macros. Local macros like GLZ_CSV_NL is fine. For glaze, there are a few way out: put the macros in a separate file, and include that file in those that needed the macros, put the macro definitions on the command line using -D. What do you want to do? I am seeing this happen with: GLZ_ALWAYS_INLINE, (what else?)

msqr1:
I also replaced the static_assert(false, "Eigen must be included to use glaze/ext/eigen.hpp") and static_assert(false, "standalone or boost asio must be included to use glaze/ext/glaze_asio.hpp"); with #error because my tool doesn't recognize a `static_assert(false, ...);` as a preprocessor directive, is that ok?

stephenberry:
> I also replaced the `static_assert(false, "Eigen must be included to use glaze/ext/eigen.hpp");` and `static_assert(false, "standalone or boost asio must be included to use glaze/ext/glaze_asio.hpp");` with `#error` because my tool doesn't recognize a `static_assert(false, ...);` as a preprocessor directive, is that ok?

Yes, this sounds fine.

stephenberry:
> @stephenberry One more thing I forgot, glaze seems to have some headers depending on macros from other headers. While this would work in header build, it won't be in modules because modules don't leak macros. Local macros like GLZ_CSV_NL, DLL_EXPORT are fine. For glaze, there are a few way out: put the macros in a separate file, and include that file in those that needed the macros, put the macro definitions on the command line using -D. What do you want to do? I am seeing this happen with: GLZ_ALWAYS_INLINE, [most of the macros in util/parse.hpp] (what else?)

Yeah, the macros in `util/parse.hpp` are also useful for users writing custom implementations. So, I think they'll need to exist in a header that developers can also include. I think this solution is best:

> ...put the macros in a separate file, and include that file in those that needed the macros...

Note that C++ doesn't have a mechanism of optionally returning from an inlined function call. So, macros are needed to keep code reasonably clean when writing code without exceptions and needing to return on error. Hence, the reason for most of these macros.

msqr1:
> Note that C++ doesn't have a mechanism of optionally returning from an inlined function call. So, macros are needed to keep code reasonably clean when writing code without exceptions and needing to return on error. Hence, the reason for most of these macros.

Can you elaborate further on this and give an example?

stephenberry:
Here's an example of a macro:
```c++
#define GLZ_INVALID_END(RETURN)                  \
   if constexpr (not Opts.null_terminated) {     \
      if (it == end) [[unlikely]] {              \
         ctx.error = error_code::unexpected_end; \
         return RETURN;                          \
      }                                          \
   }
```

This can be used simply by adding the line:
```c++
GLZ_INVALID_END();
```

To do this with a forced inline function call we could write the function as:
```c++
template <auto Opts>
GLZ_ALWAYS_INLINE void invalid_end(context& ctx, auto& it) noexcept {
   if constexpr (not Opts.null_terminated) { 
      if (it == end) [[unlikely]] {
      }
   }
}
```

To use this function we need to write:
```c++
invalid_end<Opts>(ctx, it);
if (bool(ctx.error)) [[unlikely]] {
   return;
}
```

Not only is this more typing, but it actually creates two checks for the error rather than just one. So, it has double the branching. Even if it is optimized away in Release builds, it still makes Debug builds slower. And, the force inline tends to increase compilation times (especially on MSVC). So, the macro approach can build much faster.

Now, consider functions that might have multiple return points. The optimizer begins to struggle more when the error paths become more complex. If we need to add this additional check after calling our function then we can end up with more branching than the macro code, and in order to force inline the function we also hurt compilation times.

stephenberry:
In looking at the error handling problem again I discovered a way to get as efficient assembly with the non-macro approach in Release mode. It looks like I can avoid using force inlining for at least the simpler functions. There will still be additional overhead in Debug builds, but I think it is worth it to eliminate macros. I'll do more experimentation later.

msqr1:
I'm also curious to how you do this. For now I don't see any way without force inlining.

msqr1:
Why don't you use an enum? It is literally just named numbers. I'm pretty sure they are also efficient.

msqr1:
That was what I had in mind, too, yesterday. Returning a bool for compilers to optimize it away

stephenberry:
I'm removing macros in this pull request: #1614

msqr1:
Just making sure, @stephenberry, the only non-local macros left are in util/inline.hpp, right? Are there any other files that provides macro definitions for others?

stephenberry:
Yes, I think the only macros that need to be shared in order to build are those in `util/inline.hpp`. Let me know if you discover otherwise.

msqr1:
I'm still trying to set up CMake modules build. It doesn't work at all, I need some help lol @stephenberry

msqr1:
Hi, I got these error while trying to build normally with clang(++)-19 and libc++:
```
glaze-modularized/tests/beve_test/beve_test.cpp:1443:12: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
 1443 |       auto operator<=>(const A1&) const = default;
      |            ^
glaze-modularized/tests/beve_test/beve_test.cpp:1441:35: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'a'
 1441 |       std::map<uint8_t, uint64_t> a{};
      |                                   ^
/usr/include/c++/v1/__threading_support:680:17: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to '__thread_id' for 1st argument
  680 | strong_ordering operator<=>(__thread_id __x, __thread_id __y) noexcept {
      |                 ^           ~~~~~~~~~~~~~~~
/usr/include/c++/v1/system_error:481:1: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'const error_code' for 1st argument
  481 | operator<=>(const error_code& __x, const error_code& __y) noexcept
      | ^           ~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/system_error:489:1: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'const error_condition' for 1st argument
  489 | operator<=>(const error_condition& __x, const error_condition& __y) noexcept
      | ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/__variant/monostate.h:32:49: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'monostate' for 1st argument
   32 | _LIBCPP_HIDE_FROM_ABI constexpr strong_ordering operator<=>(monostate, monostate) noexcept {
      |                                                 ^           ~~~~~~~~~
/usr/include/c++/v1/__utility/pair.h:444:1: note: candidate template ignored: could not match 'pair' against 'std::map'
  444 | operator<=>(const pair<_T1,_T2>& __x, const pair<_T1,_T2>& __y)
      | ^
/usr/include/c++/v1/__iterator/reverse_iterator.h:288:1: note: candidate template ignored: could not match 'reverse_iterator' against 'std::map'
  288 | operator<=>(const reverse_iterator<_Iter1>& __x, const reverse_iterator<_Iter2>& __y)
      | ^
/usr/include/c++/v1/__iterator/reverse_iterator.h:288:1: note: candidate template ignored: could not match 'reverse_iterator' against 'std::map'
/usr/include/c++/v1/tuple:1572:1: note: candidate template ignored: could not match 'tuple' against 'std::map'
 1572 | operator<=>(const tuple<_Tp...>& __x, const tuple<_Up...>& __y)
      | ^
/usr/include/c++/v1/tuple:1572:1: note: candidate template ignored: could not match 'tuple' against 'std::map'
/usr/include/c++/v1/string_view:784:1: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
  784 | operator<=>(basic_string_view<_CharT, _Traits> __lhs, basic_string_view<_CharT, _Traits> __rhs) noexcept {
      | ^
/usr/include/c++/v1/string_view:797:38: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
  797 | _LIBCPP_HIDE_FROM_ABI constexpr auto operator<=>(
      |                                      ^
/usr/include/c++/v1/string_view:797:38: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
/usr/include/c++/v1/string:4077:38: note: candidate template ignored: could not match 'basic_string' against 'std::map'
 4077 | _LIBCPP_HIDE_FROM_ABI constexpr auto operator<=>(
      |                                      ^
/usr/include/c++/v1/string:4085:1: note: candidate template ignored: could not match 'basic_string' against 'std::map'
 4085 | operator<=>(const basic_string<_CharT, _Traits, _Allocator>& __lhs, const _CharT* __rhs) {
      | ^
/usr/include/c++/v1/string:4085:1: note: candidate template ignored: could not match 'basic_string' against 'std::map'
/usr/include/c++/v1/__memory/unique_ptr.h:566:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
  566 | operator<=>(const unique_ptr<_T1, _D1>& __x, const unique_ptr<_T2, _D2>& __y) {
      | ^
/usr/include/c++/v1/__memory/unique_ptr.h:566:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
/usr/include/c++/v1/__memory/unique_ptr.h:658:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
  658 | operator<=>(const unique_ptr<_T1, _D1>& __x, nullptr_t) {
      | ^
/usr/include/c++/v1/__memory/unique_ptr.h:658:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
/usr/include/c++/v1/__memory/shared_ptr.h:1374:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
 1374 | operator<=>(shared_ptr<_Tp> const& __x, shared_ptr<_Up> const& __y) noexcept
      | ^
/usr/include/c++/v1/__memory/shared_ptr.h:1374:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
/usr/include/c++/v1/__memory/shared_ptr.h:1483:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
 1483 | operator<=>(shared_ptr<_Tp> const& __x, nullptr_t) noexcept
      | ^
/usr/include/c++/v1/__memory/shared_ptr.h:1483:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
/usr/include/c++/v1/__iterator/move_iterator.h:282:6: note: candidate template ignored: could not match 'move_iterator' against 'std::map'
  282 | auto operator<=>(const move_iterator<_Iter1>& __x, const move_iterator<_Iter2>& __y)
      |      ^
/usr/include/c++/v1/__iterator/move_iterator.h:282:6: note: candidate template ignored: could not match 'move_iterator' against 'std::map'
/usr/include/c++/v1/variant:1659:1: note: candidate template ignored: could not match 'variant' against 'std::map'
 1659 | operator<=>(const variant<_Types...>& __lhs, const variant<_Types...>& __rhs) {
      | ^
glaze-modularized/tests/beve_test/beve_test.cpp:1443:43: note: replace 'default' with 'delete'
 1443 |       auto operator<=>(const A1&) const = default;
      |                                           ^~~~~~~
      |                                           delete
glaze-modularized/tests/beve_test/beve_test.cpp:1451:12: error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]
 1451 |       auto operator<=>(const B&) const = default;
      |            ^
glaze-modularized/tests/beve_test/beve_test.cpp:1449:10: note: defaulted 'operator<=>' is implicitly deleted because it would invoke a deleted comparison function for member 'a'
 1449 |       A1 a{};
      |          ^
glaze-modularized/tests/beve_test/beve_test.cpp:1443:12: note: explicitly defaulted function was implicitly deleted here
 1443 |       auto operator<=>(const A1&) const = default;
      |            ^
glaze-modularized/tests/beve_test/beve_test.cpp:1441:35: note: defaulted 'operator<=>' is implicitly deleted because there is no viable three-way comparison function for member 'a'
 1441 |       std::map<uint8_t, uint64_t> a{};
      |                                   ^
/usr/include/c++/v1/__threading_support:680:17: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to '__thread_id' for 1st argument
  680 | strong_ordering operator<=>(__thread_id __x, __thread_id __y) noexcept {
      |                 ^           ~~~~~~~~~~~~~~~
/usr/include/c++/v1/system_error:481:1: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'const error_code' for 1st argument
  481 | operator<=>(const error_code& __x, const error_code& __y) noexcept
      | ^           ~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/system_error:489:1: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'const error_condition' for 1st argument
  489 | operator<=>(const error_condition& __x, const error_condition& __y) noexcept
      | ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/c++/v1/__variant/monostate.h:32:49: note: candidate function not viable: no known conversion from 'const std::map<uint8_t, uint64_t>' (aka 'const map<unsigned char, unsigned long>') to 'monostate' for 1st argument
   32 | _LIBCPP_HIDE_FROM_ABI constexpr strong_ordering operator<=>(monostate, monostate) noexcept {
      |                                                 ^           ~~~~~~~~~
/usr/include/c++/v1/__utility/pair.h:444:1: note: candidate template ignored: could not match 'pair' against 'std::map'
  444 | operator<=>(const pair<_T1,_T2>& __x, const pair<_T1,_T2>& __y)
      | ^
/usr/include/c++/v1/__iterator/reverse_iterator.h:288:1: note: candidate template ignored: could not match 'reverse_iterator' against 'std::map'
  288 | operator<=>(const reverse_iterator<_Iter1>& __x, const reverse_iterator<_Iter2>& __y)
      | ^
/usr/include/c++/v1/__iterator/reverse_iterator.h:288:1: note: candidate template ignored: could not match 'reverse_iterator' against 'std::map'
/usr/include/c++/v1/tuple:1572:1: note: candidate template ignored: could not match 'tuple' against 'std::map'
 1572 | operator<=>(const tuple<_Tp...>& __x, const tuple<_Up...>& __y)
      | ^
/usr/include/c++/v1/tuple:1572:1: note: candidate template ignored: could not match 'tuple' against 'std::map'
/usr/include/c++/v1/string_view:784:1: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
  784 | operator<=>(basic_string_view<_CharT, _Traits> __lhs, basic_string_view<_CharT, _Traits> __rhs) noexcept {
      | ^
/usr/include/c++/v1/string_view:797:38: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
  797 | _LIBCPP_HIDE_FROM_ABI constexpr auto operator<=>(
      |                                      ^
/usr/include/c++/v1/string_view:797:38: note: candidate template ignored: could not match 'basic_string_view' against 'std::map'
/usr/include/c++/v1/string:4077:38: note: candidate template ignored: could not match 'basic_string' against 'std::map'
 4077 | _LIBCPP_HIDE_FROM_ABI constexpr auto operator<=>(
      |                                      ^
/usr/include/c++/v1/string:4085:1: note: candidate template ignored: could not match 'basic_string' against 'std::map'
 4085 | operator<=>(const basic_string<_CharT, _Traits, _Allocator>& __lhs, const _CharT* __rhs) {
      | ^
/usr/include/c++/v1/string:4085:1: note: candidate template ignored: could not match 'basic_string' against 'std::map'
/usr/include/c++/v1/__memory/unique_ptr.h:566:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
  566 | operator<=>(const unique_ptr<_T1, _D1>& __x, const unique_ptr<_T2, _D2>& __y) {
      | ^
/usr/include/c++/v1/__memory/unique_ptr.h:566:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
/usr/include/c++/v1/__memory/unique_ptr.h:658:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
  658 | operator<=>(const unique_ptr<_T1, _D1>& __x, nullptr_t) {
      | ^
/usr/include/c++/v1/__memory/unique_ptr.h:658:1: note: candidate template ignored: could not match 'unique_ptr' against 'std::map'
/usr/include/c++/v1/__memory/shared_ptr.h:1374:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
 1374 | operator<=>(shared_ptr<_Tp> const& __x, shared_ptr<_Up> const& __y) noexcept
      | ^
/usr/include/c++/v1/__memory/shared_ptr.h:1374:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
/usr/include/c++/v1/__memory/shared_ptr.h:1483:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
 1483 | operator<=>(shared_ptr<_Tp> const& __x, nullptr_t) noexcept
      | ^
/usr/include/c++/v1/__memory/shared_ptr.h:1483:1: note: candidate template ignored: could not match 'shared_ptr' against 'std::map'
/usr/include/c++/v1/__iterator/move_iterator.h:282:6: note: candidate template ignored: could not match 'move_iterator' against 'std::map'
  282 | auto operator<=>(const move_iterator<_Iter1>& __x, const move_iterator<_Iter2>& __y)
      |      ^
/usr/include/c++/v1/__iterator/move_iterator.h:282:6: note: candidate template ignored: could not match 'move_iterator' against 'std::map'
/usr/include/c++/v1/variant:1659:1: note: candidate template ignored: could not match 'variant' against 'std::map'
 1659 | operator<=>(const variant<_Types...>& __lhs, const variant<_Types...>& __rhs) {
      | ^
glaze-modularized/tests/beve_test/beve_test.cpp:1451:42: note: replace 'default' with 'delete'
 1451 |       auto operator<=>(const B&) const = default;
      |                                          ^~~~~~~
      |                                          delete
2 errors generated.
ninja: build stopped: subcommand failed.
```
In tests/beve_test/beve_test.cpp, just literally relace `default` with `delete` and it will work. If you know why, could you tell me? @stephenberry

stephenberry:
@msqr1, this first line tells the error:
`error: explicitly defaulted three-way comparison operator is implicitly deleted [-Werror,-Wdefaulted-function-deleted]`

This would typically be a warning, but warnings are considered errors with `Werror`. It is just letting the programmer know that the default spaceship operator is deleted and therefore should be marked `delete` to be more clear.

msqr1:
@stephenberry Umm, hey, I found some circular dependencies that is disallowed by modules (I had to make a customized tools to find these cycles, it took way too long due to my lack of experience lol). Idk how the header build work, but the import build failed with cyclic dependency on modules. Here are the cycles:
- `glaze/core/meta.hpp` includes `glaze/util/variant.hpp` includes `glaze/util/parse.hpp` looping back to ^^^^^
- `glaze/json/ptr.hpp` and `glaze/json/write.hpp` including each other.

Could you please address these issues? And if possible, could let me know how the header build worked in the first place!

msqr1:
@stephenberry

stephenberry:
@msqr1, I just merged in a fix for the circular includes: #1656. Headers only create circular dependencies if dependencies haven't been reorganized by another include ordering elsewhere, just as in "glaze/json.hpp". This is a big benefit of modules, being able to easily catch circular dependencies.

PR #1602, opened 2025-02-07
Add error_handler to asio_server

stephenberry:


stephenberry:
@gjb1002, What do you think about this solution? It adds a `std::function<void(const std::string&)> error_handler{};` to provide a callback for handling errors. I think it needs to pass a `std::string` through to be generic.

It is important to realize that anything within the callback should be thread safe, as multiple threads could be calling this.

gjb1002:
@stephenberry thanks for fix, looks like it resolves the issues I raised!

Issue #1594, opened 2025-02-05
Request to Add TOML Support to Glaze


Kanerudaisuki:
I hope this message finds you well. I’m writing to propose that Glaze extend its capabilities by adding support for **TOML**. Building on Glaze’s impressive compile-time reflection and high-performance JSON processing, integrating TOML could offer several benefits:

**Benefits:**
- **Improved Configuration Handling:**  
  TOML’s human-friendly syntax makes it perfect for configuration files, often providing better clarity and structure than JSON.
- **Broader Use Cases:**  
  Adding TOML support opens up Glaze to projects that prefer TOML for their configuration needs, expanding its user base and application scope.
- **Leverage Compile-Time Reflection:**  
  Glaze’s existing reflection mechanisms could be adapted to map TOML’s data structures efficiently, ensuring high performance while offering multiple serialization formats.

**Challenges:**
- **Data Structure Differences:**  
  Unlike JSON, TOML supports inline tables, arrays of tables, and datetime types. Integrating these features will require careful mapping between TOML constructs and Glaze’s existing types.
- **Performance Considerations:**  
  Maintaining Glaze’s performance benchmarks will be critical. We’d need to ensure that any additional TOML parsing capabilities do not compromise speed or memory efficiency.
- **Parser Refactoring:**  
  The current architecture is optimized for JSON. Extending it for TOML might involve significant modifications in the parser and validation logic, to properly handle TOML’s unique syntax elements while preserving the “zero-overhead” philosophy.

I deeply appreciate the excellent work behind Glaze, and I believe that adding TOML support could significantly enhance its versatility. If you’re open to exploring this, I’d be happy to contribute or discuss further approaches.

Thank you for considering this suggestion. I look forward to your thoughts.

stephenberry:
Glaze is designed to support various formats, so adding TOML wouldn't be too difficult. I currently don't have a use for TOML, and therefore haven't given it attention. But, if you were interested in adding TOML support to Glaze I would be happy to show you the steps needed.

I could create the initial write/read files and a unit test in Glaze for TOML that you could work off. How does that sound?

stephenberry:
Closing because initial support has been added and users can submit specific pull requests to expand the TOML support.

PR #1603, opened 2025-02-07
Initial TOML write support

stephenberry:
#1594 -> Request for TOML support.


stephenberry:
@Kanerudaisuki

Here is the beginnings of TOML support. I'm just working on writing right now, and I'm figuring out what can be reused as to not duplicate code unnecessarily.

PR #1604, opened 2025-02-07
fix ODR violations

arturbac:
fixes some still left ODR violations


stephenberry:
Thanks!

include/glaze/ext/jsonrpc.hpp
```diff
@@ -49,7 +49,7 @@ namespace glz::rpc
 namespace glz::rpc
 {
    using id_t = std::variant<glz::json_t::null_t, std::string_view, std::int64_t>;
-   static constexpr std::string_view supported_version{"2.0"};
+   inline constexpr std::string_view supported_version{"2.0"};
```
sjanel, line 52: `static constexpr` implies `inline`, so it's better here to stick with `static constexpr`.

include/glaze/json/json_format.hpp
```diff
@@ -25,7 +25,7 @@ namespace glz::detail
       Comment = '/'
    };
 
-   constexpr std::array<json_type, 256> json_types = [] {
+   inline constexpr std::array<json_type, 256> json_types = [] {
```
sjanel, line 28: Why about making sure that all the `constexpr` variables are `static constexpr` instead?

stephenberry: static constexpr implies one definition per compile unit. inline constexpr means one definition across all compile units, merged my the linker.

arturbac: constexpr by default implies static - internal to translation unit linking, so address after including and defining is different for every translation unit. 
inline constexpr - forces linker to make sure all translation units will use only one instance with the same address of given variable/constant.
The result of using (static) constexpr variable is almost harmless only blows You executable as long U don't compare pointers. But when You will pass address of such variable across multiple translation units and compare to itself address will be not equal to.
so in simple words 
A.cpp and B.cpp #include <header> with constexpr std::array<json_type, 256> json_types.
&json_types[0] from A.cpp != &json_types[0] B.cpp

PR #1614, opened 2025-02-14
Removing macros

stephenberry:
Removing macros for the sake of C++20 modules and easier debugging.

### DIFF
--- a/include/glaze/beve/header.hpp
+++ b/include/glaze/beve/header.hpp
@@ -13,11 +13,19 @@
 #include "glaze/core/context.hpp"
 #include "glaze/util/inline.hpp"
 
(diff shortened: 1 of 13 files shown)


msqr1:
Do we have performance tests for  glaze?

stephenberry:
I think it is because the boolean is simpler for the compiler to see through. It is easy for the compiler to see that a boolean is being checked twice (to achieve a single path) and merge it into a single check. It is more complicated if a register is being modified, because there could be other side effects.

stephenberry:
I'm merging in these changes. @msqr1, let me know if there are more macros that I ought to eliminate.

Issue #1619, opened 2025-02-17
Support for custom formats


noahfrn:
Hi there,

Thank you for this great library!

It'd be great to add support for custom, user-defined formats: As far as i can see, these are currently blocked behind the `read|write_<format>_supported` [concept check here](https://github.com/stephenberry/glaze/blob/4a6b016b9a33ed8d3959075ef46b6814dfff9c39/include/glaze/core/opts.hpp#L375). We could extend the series of `if constexpr` checks here to allow for a user-provided format trait.

I'm happy to take on this work myself, but just wanted to check that this would be an acceptable change, and if not, how else I could help to use glaze as a generic reflection library, without necessarily relying on its built-in serialization formats.

stephenberry:
Thanks for bringing this up. As Glaze continues to add more formats this code is problematic, because it is not generic. And, it would be great to allow user formats.

I started a pull request to make this change here: #1622 

Basically, the concepts should simplify to generic versions like:
```c++
template <uint32_t Format, class T>
concept write_supported = requires { detail::to<Format, std::remove_cvref_t<T>>{}; };

template <uint32_t Format, class T>
concept read_supported = requires { detail::from<Format, std::remove_cvref_t<T>>{}; };
```

noahfrn:
Thank you ! I took a look at that change, and that looks like exactly what I wanted

noahfrn:
As a follow-up, out of curiosity:

Formats are currently defined by a uint32. Is there a guaranteed range that I can use to add custom formats? It'd be unfortunate to have a new format in glaze collide with the format integer that users choose

stephenberry:
I agree, there should be a range for custom formats, which should be larger than built in formats to help prevent collisions. I'm thinking built in formats will be less than 65536, and custom formats can occupy 65536 to 4294967296.

stephenberry:
I've merged this into `main`, and I will release this change in `v5.0.0`. I have a few more breaking changes that I intend to make before releasing a new version.

PR #1622, opened 2025-02-18
Generic supported concepts

stephenberry:
The format specific concepts like: `read_json_supported` and `write_beve_supported` added extra boilerplate and did not allow custom formats. This change removes these non-generic functions and simply uses:

```c++
template <uint32_t Format, class T>
concept write_supported = requires { detail::to<Format, std::remove_cvref_t<T>>{}; };

template <uint32_t Format, class T>
concept read_supported = requires { detail::from<Format, std::remove_cvref_t<T>>{}; };
```

This more generic solution simplifies the code and makes adding new formats cleaner and possible for users without needing to modify the main Glaze repository.

Requested in #1619, but also noticed when beginning to add TOML that this approach is bad, especially as the number of formats grows.

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -1546,13 +1546,15 @@ namespace glz
       };
    }
 
-   template <read_beve_supported T, class Buffer>
+   template <class T, class Buffer>
+   requires (read_supported<BEVE, T>)
    [[nodiscard]] inline error_ctx read_beve(T&& value, Buffer&& buffer)
    {
       return read<opts{.format = BEVE}>(value, std::forward<Buffer>(buffer));
    }
 
-   template <read_beve_supported T, class Buffer>
+   template <class T, class Buffer>
+   requires (read_supported<BEVE, T>)
    [[nodiscard]] inline expected<T, error_ctx> read_beve(Buffer&& buffer)
    {
       T value{};
@@ -1563,7 +1565,8 @@ namespace glz
       return value;
    }
 
-   template <opts Opts = opts{}, read_beve_supported T>
+   template <opts Opts = opts{}, class T>
+   requires (read_supported<BEVE, T>)
    [[nodiscard]] inline error_ctx read_file_beve(T& value, const sv file_name, auto&& buffer)
    {
       context ctx{};
@@ -1578,14 +1581,16 @@ namespace glz
       return read<set_beve<Opts>()>(value, buffer, ctx);
    }
 
-   template <read_beve_supported T, class Buffer>
+   template <class T, class Buffer>
+   requires (read_supported<BEVE, T>)
    [[nodiscard]] inline error_ctx read_binary_untagged(T&& value, Buffer&& buffer)
    {
       return read<opts{.format = BEVE, .structs_as_arrays = true}>(std::forward<T>(value),
                                                                    std::forward<Buffer>(buffer));
    }
 
-   template <read_beve_supported T, class Buffer>
+   template <class T, class Buffer>
+   requires (read_supported<BEVE, T>)
    [[nodiscard]] inline expected<T, error_ctx> read_binary_untagged(Buffer&& buffer)
    {
       T value{};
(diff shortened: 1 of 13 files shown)


stephenberry:
@arturbac, This is a breaking change that will effect [simple_enum](https://github.com/arturbac/simple_enum). I intend to release this in v5.0.0.

Do you have additional thoughts on this change?
I plan to the remove the `detail::` in the namespace for `to/from` specializations as well in v5.0.0. I'll do that in a follow on pull request.

arturbac:
I looked at it, I was using those concepts and I will remove them from simple_enum header and move use to unit tests.
So add pls feature test macro only.
As i understand I'll have to add only conditional test on this feature test macro like below unit tests...
Not a problem at all, thx for info

``cpp
#ifndef glz_feature_xxxx
static_assert(glz::read_json_supported<test_enum_e>);
static_assert(glz::write_json_supported<test_enum_e>);
#else
static_assert(glz::read_supported<glz::JSON,test_enum_e>);
static_assert(glz::write_supported<glz::JSON,test_enum_e>);
#endif
```

**Comment by arturbac** [comment]
My taughts:
to/from specializations in simple enum are generic .


```cpp
namespace glz::detail
  {
template<simple_enum::bounded_enum enumeration_type>
#ifdef glaze_v3_5_0_to_from
struct from<glz::JSON, enumeration_type>
#else
struct from_json<enumeration_type>
#endif
  {
  template<auto Opts, typename... Args>
    requires simple_enum::bounded_enum<enumeration_type>
  static void op(enumeration_type & arg, is_context auto && ctx, Args &&... args)
    {
    std::string_view value;
#ifdef glaze_v3_5_0_to_from
    read<glz::JSON>::op<Opts>(value, ctx, args...);
#else
    read<glz::json>::op<Opts>(value, ctx, args...);
#endif
    if(bool(ctx.error)) [[unlikely]]
      return;

    simple_enum::expected<enumeration_type, simple_enum::enum_cast_error> res{simple_enum::enum_cast<enumeration_type>(value
    )};
    if(!res.has_value()) [[unlikely]]
      {
      ctx.error = error_code::syntax_error;
      return;
      }
    arg = res.value();
    }
  };
```
So actually when You plan to remove detail all I have to do is to remove namespace scope and use namespace prefix dependant on feature test  macro
```cpp
#ifdef glz_feature_xxx....
namespace glz_spec_namespace = glz::detail;
#else
namespace glz_spec_namespace = glz;
#endif
template<simple_enum::bounded_enum enumeration_type>
#ifdef glaze_v3_5_0_to_from
struct glz_spec_namespace::from<glz::JSON, enumeration_type>
#else
struct glz_spec_namespace::from_json<enumeration_type>
#endif
  {
```

arturbac:
Give me names of feature test macros and I'll blind implement those changes, for You in advance.

stephenberry:
@arturbac, Sweet! The feature test macro `glaze_v5_0_0_generic_supported` has been added to indicate the generic `write_supported` and `read_supported` concepts.

arturbac:
@stephenberry can You please take a look and make review I tried adding You to review but I can not do that without Your consent.
For 5.x I had to change read/write to direct use from/to::op
Does this look reasonable ? Or I should do it differently ? Did I use proper feature test macros ?

https://github.com/arturbac/simple_enum/pull/26/files
ex:
```cpp
namespace glz_to_from_ns = glz;
struct glz_to_from_ns::from<glz::JSON, enumeration_type>
{
  template<auto Opts, typename... Args>
    requires simple_enum::bounded_enum<enumeration_type>
  static void op(enumeration_type & arg, is_context auto && ctx, Args &&... args)
    {
    std::string_view value;
#ifdef glaze_v3_5_0_to_from
#ifdef glaze_v5_0_0_generic_supported
    glz::from<glz::JSON, std::string_view>::op<Opts>(value, ctx, args...);
```

stephenberry:
@arturbac, this looks good although some compilers require `template` on the `op`:
`glz::from<glz::JSON, std::string_view>::template op<Opts>(value, ctx, args...);`

Glaze was confusing prior to v5 because glz::read/glz::write were different from glz::detail::read/glz::detail::write. The detail structs were what you were calling. These still exist, but are now glz::parse/glz::serialize.

glz::parse adds a bit of const handling logic that you probably want, but shouldn't need for most cases. So, in most cases your change works fine.

To keep the same behavior as before the change would look like:
`parse<glz::JSON>::op<Opts>(value, ctx, args...);`

So, `read` just changes to `parse`.

PR #1656, opened 2025-03-19
Fix circular includes

stephenberry:


stephenberry:
#1599 brought up these circular headers

PR #1664, opened 2025-03-24
Add glz::array_adapter, particularly useful for immutable elements

stephenberry:


stephenberry:
Addresses: #1658

stephenberry:
@joshuamaiche, how does the name `glz::array_adapter` sound to you? This pull request adds the feature.

joshuamaiche:
@stephenberry I defer to you, since you know what sort of language best fits cohesively with the rest of the library.

I like that `array_adapter` has precedence with the Android SDK, so people coming from that will see the parallel. I personally haven't used the Android SDK before, so I would still have to look up what it does. Something like `array_visitor` would make that more clear to me.

A potential issue with both of these names is that they're both objects, which could be inconsistent with the other wrappers (which seem to have more of a "take action on data" type of name, like `quoted` or `bools_as_numbers`. If we were trying to stick more with that theme, maybe something like `transformed_array` or `adapted_array`?

stephenberry:
@joshuamaiche, Thanks for your feedback, it's always good to discuss names before releasing a new tool. I chose `array_adapter` as it relates to JSON array types rather than object types. I also like `array_visitor`, and I think it might be a bit better because it is applying another wrapper to the underlying elements.

I'm thinking one of these three. Which is your preference?
- `array_adapter`
- `array_visitor`
- `array_apply`

I'm leaning towards `array_apply`, because it would look like `array_apply<ImmutableStructInserter<&T::vals>`, which is pretty easy to read that we are applying the template parameter wrapper.

joshuamaiche:
@stephenberry I agree, `array_apply` sounds like the best to me!

PR #1669, opened 2025-03-26
fix static_string

SlaaneshChosen:
`static_string` should use `capacity` to check whether it has enough elements to store the string, rather than `size()` (which is used in `std::array<char, N>`).


stephenberry:
Thanks for this fix!

PR #1673, opened 2025-03-28
Erlang term parse and serialize

fregate:
Add Erlang external term format parser and serializer.
This is "first" release of support with sets of features that enough for my purposes (an even more, but I don't think this is full set)

Use custom options for that format (great addition by the way)

This format supports only with `glaze_ERLANG_FORMAT` CMake options enabled (disabled by default) because I had to include erlang libs to parse this format.

### DIFF
--- a/include/glaze/core/opts.hpp
+++ b/include/glaze/core/opts.hpp
@@ -26,6 +26,10 @@ namespace glz
    inline constexpr uint8_t rowwise = 0;
    inline constexpr uint8_t colwise = 1;
 
+   // layout erlang term
+   inline constexpr uint8_t map = 0;
+   inline constexpr uint8_t proplist = 1;
+
    enum struct float_precision : uint8_t { //
       full, //
       float32 = 4, //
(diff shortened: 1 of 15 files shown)


fregate:
And of course at this moment there is no checks in CI because there is no erlang support and pipeline doesn't know anything about ERLANG_FORMAT option.

fregate:
Add erlang workflows to actions

include/glaze/core/opts.hpp
```diff
@@ -20,6 +20,7 @@ namespace glz
    inline constexpr uint32_t TOML = 400;
    inline constexpr uint32_t STENCIL = 500;
    inline constexpr uint32_t CSV = 10000;
+   inline constexpr uint32_t ERLANG = 20000;
```
qudix, line 23: shouldn't this be named `EETF`?

fregate: Ok, I will rename erlang->eetf everywhere

stephenberry:
This looks good to me. I'm fine with using the Erlang library with the long term goal of replacing it. Since Erlang support is optional in Glaze with `glaze_EETF_FORMAT` I'm fine with this approach. Ideally Glaze does not have additional dependencies other than the standard template library, but that can be accomplished over time.

stephenberry:
I'm merging this!

PR #1691, opened 2025-04-10
Fix Erlang format workflow

stephenberry:
The wrong CMake option name was being using, so eetf_tests was not being run. And, this update also restricts the EETF workflows to only testing the EETF format.


stephenberry:
@fregate, I fixed the workflows so that they actually run the eetf_tests and restrict these workflows to only run these tests so that they'll complete faster.

fregate:
Yes, my fault. Thanks!

PR #1699, opened 2025-04-17
REST API for Glaze

stephenberry:
Glaze REST and HTTP support will grow over time, but this reworks the registry to be more generally useful and helps future-proof registry design to support various protocols.

This pull request also adds `glz::mustache`, which escapes HTML when using double braces and does not escape with triple braces. `glz::stencil` still exists for contexts that don't care about HTML.


tests/rest_test/rest_registry_test.cpp
```diff
@@ -0,0 +1,358 @@
...
+   UserService userService;
+   
+   glz::rest_registry registry;
+   
+   registry.on(userService);
+   
+   server.mount("/api", registry.endpoints);
+   
+   // Serve the frontend files
+   server.get("/", [](const glz::Request& /*req*/, glz::Response& res) {
```
qudix, line 97: Any reason not have these (`glz::Request` and `glz::Response`) be snake case like the rest?

stephenberry: I do want them snake case. Thanks for reminding me. I was afraid of a collision with `glz::repe::request`, but there are different namespaces. Will update.

Issue #1675, opened 2025-03-31
Build is broken on Big-endian platforms due to `hash_map.hpp`


barracuda156:
One of the headers explicitly requires an LE platform: https://github.com/stephenberry/glaze/blob/fafa42a345752fd54ff33e6b461909d198e20c2d/include/glaze/util/hash_map.hpp#L86

Subsequently, it cannot be built on any Big-endian platform. It means that dependencies cannot be built, such as `hyprland`, which depend on `glaze`: https://github.com/freebsd/freebsd-ports/blob/e032f88df168456dbf6ad0b9af9a56f4cd75668e/x11-wm/hyprland/Makefile#L14

@stephenberry It is possible to fix hash implementation here?

barracuda156:
@stephenberry Thank you for responding!
AFAIK, CI can be set for BE Linux, I will check for a reference.

Overall, of the currently developed OS all major BSDs support BE in their releases and snapshots (FreeBSD, NetBSD, OpenBSD) and at least some Linux distros. This is true both for 64- and 32-bit BE archs.
I am building on macOS on PowerPC, but there is no need to be concerned with any macOS specifics, that I can take care of on my end.

hyperair:
Heya, this is also causing trouble for Debian on s390x. Here's a sample build failure from hyprland: https://buildd.debian.org/status/fetch.php?pkg=hyprland&arch=s390x&ver=0.51.1%2Bds-1&stamp=1762980407&file=log

stephenberry:
I'm working through all the active issues, so I hope to get to this one soon, but it will take some time. Thanks for giving more feedback and impetus to support big endian.

stephenberry:
Big endian support has been merged! #2141 

The CI now cross compiles to big endian with GCC and unit tests via QEMU.

These changes should have no runtime effect on little endian performance. Relies on compile time checking: `if constexpr (std::endian::native == std::endian::big)`

barracuda156:
@stephenberry Great, thank you!

PR #1727, opened 2025-05-03
Add CI jobs for big-endian builds

silvanshade:
Related: https://github.com/stephenberry/glaze/issues/1675

This PR adds CI jobs for big-endian builds to help development of big-endian support.

The build target is `s390x-unknown-linux-gnu` since this is the only big-endian target easy to set up on GitHub actions without resorting to containers or other virtualization options. Although Ubuntu lists `powerpc` (32-bit) as a supported big-endian target, it is apparently no longer an officially maintained platform and has no recent releases.

The `clang-linux-big-endian` job uses `clang` and `libc++` and targets `s390x-unknown-linux-gnu`. It uses `mold` as the linker because `lld` does not support `s390x` and using GNU `ld` would require setting up a different compiler toolchain (which we do in the other job). This job also installs the latest CMake since version `>= 3.29` is needed (with support for `CMAKE_LINKER_TYPE`) to configure the linker easily for cross-compilation.

The `gcc-linux-big-endian` job uses a standard GNU cross-compiling toolchain configuration with the `s390x-linux-gnu-` prefix.

Both jobs also install `qemu-user-static` which automatically configures `binfmt` so any `s390x` binaries can be run from the shell without manually invoking qemu. This can also be configured separately through CMake for CTest via the `CMAKE_CROSSCOMPILING_EMULATOR` variable.

Both jobs successfully begin compilation but fail as expected on the little-endianness assertions.

### DIFF
--- a/.github/workflows/clang-linux-big-endian.yml
+++ b/.github/workflows/clang-linux-big-endian.yml
@@ -1,0 +1,110 @@
+name: clang-linux-big-endian
+
+on:
+  push:
+    branches:
+      - main
+      - feature/*
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        clang: [17, 18]
+        build_type: [Debug]
+        std: [23]
+        target: [
+          { arch: s390x, triple: s390x-unknown-linux-gnu }
+        ]
+
+    env:
+      CC: clang-${{matrix.clang}}
+      CXX: clang++-${{matrix.clang}}
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Discover Ubuntu release codename
+      id: discover-ubuntu-codename
+      run: echo "codename=$(lsb_release -cs)" >> $GITHUB_OUTPUT
+
+    - name: Add target architecture with dpkg
+      run: sudo dpkg --add-architecture ${{ matrix.target.arch }}
+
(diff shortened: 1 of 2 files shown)


stephenberry:
This is fantastic! Thanks so much, I'll now be able to start working through fixes to support big endian.

silvanshade:
I think I'll go ahead and close this just so it doesn't show as an open PR for merge. Of course it's only meant for demonstration purposes for now.

Out of curiosity, how much effort do you think it would be to get a build working for big endian even if it wasn't optimal and maybe just replaced some of the existing algorithms with simpler but portable alternatives? Are there any easy tasks one could help with in that space?

The reason I ask is because I'm working on a large PR for `ccache` that adds support for caching C++20 module artifacts and after considering several alternatives decided to use glaze for parsing the p1689 dynamic dependency JSON files.

So far glaze seems great and the only concern I have is portability.

I suspect the intersection of users on big-endian that would also care about caching C++20 module compilation is probably not very large, but it would be nice if there was at least some way to allow it to work there without having to fall back to an alternative JSON library (or just disable the feature entirely).

stephenberry:
@silvanshade, thanks again for this demonstration. I would like to support big endian in the long run, but right now it's going to be a slow transition because I have limited time to devote to this task. There are a few core improvements I want to make to the library that will remove code and reduce the complexity of supporting big endian (one of those is removing the old hash-map system, which has already been superseded in most places in the codebase).

I don't actually know how much work it would take to support big endian, as I haven't thought through the transition. I hope to do so soon to get an estimate of the effort. But, I want to be honest with you and say that I can't set a date by which big endian would be supported, and therefore this might be a reason to avoid Glaze for now.

stephenberry:
I would recommend simdjson as long as your JSON document typically parses all keys and in the order that they are defined. You'll pay penalties if keys are missing or out of order with simdjson. I think it supports big endian, but you'll have to ensure that it does. It supports parsing with std::string_view, avoiding allocations. And, it is quite mature.

Issue #1676, opened 2025-04-02
[Question] Possible to support types like `std::inplace_vector`?


lord-ne:
Hello. I was wondering, how does the library interact with container types like [`std::inplace_vector`](https://en.cppreference.com/w/cpp/container/inplace_vector) (or its [beman implementation](https://github.com/bemanproject/inplace_vector)) that have a known maximum capacity? Is it possible for a user to define custom formatting for these types so that glaze will return an error code if attempting to read a JSON array into them that is too large (instead of throwing `std::bad_alloc`)? I'm not seeing a way for a user to do that except by completely redoing all of the code parsing JSON arrays in their own `from` method, but I'm happy to be shown otherwise.

stephenberry:
Nice, I hadn't realized `std::inplace_vector` was accepted to C++26. This is great!

A bit of work would be needed to return a nice error code if attempting to read out of the array bounds. But, it shouldn't be too difficult.

Do you have any additional thoughts of what a C++20 concept for `std::inplace_vector` would look like to differentiate it from `std::vector` and similar interfaces? I think we should be able to use the fact that it takes a `size_t` as its second template parameter. And, `capacity()` should be able to distinguish it from a `std::span` and the like.

I'll keep this issue alive until support is added. I'll probably use the [beman implementation](https://github.com/bemanproject/inplace_vector) for testing. Thanks for pointing that out.

lord-ne:
I think a `try_emplace_backable` concept makes sense: For classes that have a [`try_emplace_back()`](https://en.cppreference.com/w/cpp/container/inplace_vector/try_emplace_back) method, use that instead of `emplace_back`, and return an error code if it fails

stephenberry:
Nice. I think checking for `try_emplace_back` is a good solution. I probably won't get to this right away, so feel free to submit a pull request if you need support sooner, otherwise I'll get this done when I'm able.

PR #1729, opened 2025-05-05
Add support for inplace_vector

20162026:
#1676


20162026:
It works, but I'm not sure if it is worthy to be merged becouse:
1) only json read is implemented ( beve dose not use emplace_back and would require more modifications, didn't check others)
2) serializing into inplace_vector without exceptions is tricky becouse of multiple resize calls
3) beman::inplace_vector is not fully implemented (e.g. <=> is broken, no proper constexpr handling, init list only works with exceptions enabled)

stephenberry:
Thanks for making this merge request! I do like the idea of inplace_vector support and this gives a starting point for experimenting with that support. I'll look into them when I get some time. I wouldn't mind merging in a solution that is 90% there and improving it over time.

20162026:
I will try to bring beman.inplace_vector into a somewhat usable state before continuing with this PR.

If anyone knows of a good inplace_vector implementation that doesn’t require a full custom STL, please let me know :)

stephenberry:
@20162026, I walked AI through generating an `inplace_vector` implementation based only the STL and using C++23. I've added it to Glaze: `glaze/containers/inplace_vector.hpp`. Will need more testing, but I think this should be a good approach for implementing it until new get it in C++26. The goal will be to make it efficient while complying with the future standard.

stephenberry:
@20162026, I think `glz::inplace_vector` is looking pretty good now. It could probably be optimized in a few cases, but it should be high performance for trivial types. Let me know what you think.

include/glaze/containers/inplace_vector.hpp
```diff
@@ -0,0 +1,1070 @@
...
+            try_emplace_back(*it);
+         }
+         return it;
+      }
+
+      // Unchecked APIs
+      template <class... Args>
+      constexpr reference unchecked_emplace_back(Args&&... args)
+      {
+         return *try_emplace_back(std::forward<Args>(args)...);
```
20162026, line 502: potential nullptr dereference, it should be done the other way around - `try_emplace_back` calls `unchecked_emplace_back` .

I will go trough the whole source later this week and most likely add some more test, although I personally would rather use beman implementation

stephenberry: @20162026, thanks for looking at this! My thinking is that we can keep this implementation available in Glaze for the sake of testing, experimenting with optimizations, and avoiding another external dependency. But, since C++20 concepts are used for Glaze to support `inplace_vector` you can personally keep using the beman implementation, and anyone can bring whatever implementation complies with the concepts.

20162026:
@stephenberry also would you be interested in supporting `freestanding` inplace_vector in glaze, that would `=delete` all of the potentaly throwing functions?  According to [spec](https://eel.is/c++draft/inplace.vector.overview) all throwing functions are marked with 'freestanding-deleted` which means its up to implementer if they want to leave them or remove.

deleting throwing/aborting functions is extremely convenient for embedded, as you get compiler warning instead of runtime error when trying to use potentially throwing functions. But it would require at least some minor concept changes as it will no longer be direct vector replacement.


also s a sidenot, would you like to add [pre-commit](https://pre-commit.com/) for formatting?

stephenberry:
@20162026, yes I think 'freestanding-deleted' makes sense. Go ahead.

And, per your side note about formatting, Glaze is actually setup to use clang-format (just upgraded to v20) automatically when the code gets merged, so don't worry about formatting this merge request.

stephenberry:
@20162026, CI passed.

20162026:
hi @stephenberry, don't want to rush you or sound rude, but what is the state of this PR? would you like to see some changes or is it ok as is?

stephenberry:
I think it is in a good place. If you agree, I'll merge it in.

20162026:
ok, you can merge it

include/glaze/containers/inplace_vector.hpp
```diff
@@ -0,0 +1,1111 @@
...
+         }
+         else {
+            for (size_type i = 0; i < n; ++i) {
+               std::construct_at(this->data_ptr() + this->storage_size(), value);
+               this->set_storage_size(this->storage_size() + 1);
+            }
+         }
+      }
+
+      // TODO add assign(initializer_list<T>)
```
schaumb, line 616: Missing

include/glaze/containers/inplace_vector.hpp
```diff
@@ -0,0 +1,1111 @@
...
+      }
+
+      constexpr void resize(size_type sz, const T& c) // freestanding-deleted
+      {
+         if (sz > N) GLZ_THROW_OR_ABORT(std::bad_alloc());
+
+         if (sz > this->storage_size()) {
+            if constexpr (std::is_trivially_copyable_v<T>) {
+               // For trivially copyable types, construct one and copy the rest
+               if (this->storage_size() < sz) {
```
schaumb, line 643: Double check. 640 already the same condition

include/glaze/containers/inplace_vector.hpp
```diff
@@ -0,0 +1,1111 @@
...
+                  if (this->storage_size() == 0 && sz > 0) {
+                     std::construct_at(this->data_ptr(), c);
```
schaumb, line 645: I didn't get it. At 652 it is OK to memcpy the &c, then here why constructed the first element and then memcpy that (isn't it easier to memcpy &C here too)?

include/glaze/containers/inplace_vector.hpp
```diff
@@ -0,0 +1,1111 @@
...
+               }
+
+               this->set_storage_size(this->storage_size() + count);
+            }
+         }
+         else {
+            // For non-random access iterators, copy to temporary buffer first
+            inplace_vector<T, N> temp(first, last);
+
+            if (this->storage_size() + temp.size() > N) GLZ_THROW_OR_ABORT(std::bad_alloc());
```
schaumb, line 893: Insert() already checks this, not necessary

stephenberry:
@schaumb, thanks so much for taking the time to review this implementation of inplace_vector. Would you be willing to submit a pull request with fixes? Otherwise, I'll get to this as soon as I'm able.

schaumb:
I won't be able to handle the implementation myself, but I'm happy to review it once it's ready.

stephenberry:
No, problem. Thanks!

PR #1744, opened 2025-05-17
Update fast_float from 7.0.0 to 8.0.2

musicinmybrain:
There’s nothing earthshaking in the upstream release notes, but it’s nice to keep up to date.

- https://github.com/fastfloat/fast_float/releases/tag/v8.0.2
- https://github.com/fastfloat/fast_float/releases/tag/v8.0.1
- https://github.com/fastfloat/fast_float/releases/tag/v8.0.0

I used https://github.com/fastfloat/fast_float/blob/v8.0.2/script/amalgamate.py, then hand-edited to prefix `FASTFLOAT_` macros with `GLZ_` and replace `namespace fast_float` with `glz::fast_float`. (I also did the namespace fix in the namespace-closing comments, which wasn’t the case before.)

### DIFF
--- a/include/glaze/util/fast_float.hpp
+++ b/include/glaze/util/fast_float.hpp
@@ -1,4 +1,4 @@
-// fast_float version 7.0.0, https://github.com/fastfloat/fast_float
+// fast_float version 8.0.2, https://github.com/fastfloat/fast_float
 
 // clang-format off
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks so much for your work on this update!

PR #1783, opened 2025-06-03
Move jump table parameter pack expansion into helper for MSVC

neilharan:
Fixes MSVC error


stephenberry:
Thanks for this fix! Do you think you can put the code in an MSVC macro check? This will make the hack more clear and I'll raise an issue with MSVC at some point.
I'd like to keep the Clang/GCC code as intended.

stephenberry:
I filed an issue with MSVC: https://developercommunity.visualstudio.com/t/MSVC-fails-to-compile-visit-pattern-but/10916572

But, for now, your workaround makes sense.

stephenberry:
@neilharan, thanks again for this fix! I added a macro for MSVC. Now merging.

PR #1787, opened 2025-06-04
Fix packit specfile_path

stephenberry:


stephenberry:
@bkmgit, okay just fedora build failed. Any ideas?

bkmgit:
Thanks for fixing.

PR #1789, opened 2025-06-05
Build and test on ix86

bkmgit:


bkmgit:
Clang build failure seems unrelated.

stephenberry:
@bkmgit, thanks for this fix! The Clang failure is a corner case threading issue from a unit test that I'm now fixing.

PR #1808, opened 2025-06-14
Add static_cast to remove warning

Frafjord:
When compiling with VS 2022 (17.14.5) I get the warning:

glaze\include\glaze\stencil\stencil.hpp(16,35): warning C4244: 'argument': conversion from 'double' to 'const unsigned __int64', possible loss of data

I added a static_cast<size_t>.


stephenberry:
Thanks!

PR #1815, opened 2025-06-17
add fuzzer for json jmespath

pauldreik:
this also adds a testcase for an out of bounds read the fuzzer found


stephenberry:
Thanks for these pull requests! I will fix the issues and merge as soon as I'm able.

stephenberry:
The fuzz test needed the option `null_terminated = false`, because a non-null terminated buffer was being used.
The unit test that was added was illogical and passed a mutable buffer while passing a string_view as the layout. The mutable buffer could be changed, so this was a broken test. The reason why the fuzz failed was because it was missing the option `null_terminated = false` or a null terminated buffer.

pauldreik:
there are problems even with null_terminated set to false, pushed a slightly modified version.

stephenberry:
Yeah, I'm seeing these issues. I'll try to tackle them soon. Thanks!

PR #1827, opened 2025-06-18
extend csv fuzzer to cover also rowvise operations

pauldreik:


stephenberry:
Thanks for this. I'll be out of pocket for a few weeks, but will get this fixed/merged when I'm able.

PR #1835, opened 2025-06-23
(feat): enable avx2 for mac arm chipset

jayakasadev:
* Add `simde` for Mac Arm to take advantage of `AVX2`


stephenberry:
Can you explain more of your motivation for using AVX2 via `simde` on Arm? The Glaze SWAR code for ARM is faster than NEON SIMD instructions for Mac, so I would expect this change to make performance worse on Arm Mac.

jayakasadev:
I hadn't seen the benchmarks for SWAR vs NEON. If NEON is negatively impacting perf, I'm fine closing this.

stephenberry:
Yeah, I did some focused benchmarking a year or more ago and NEON struggles because of the complexity of handling unicode code points. If we could assume no unicode code points or escaped characters then Neon would be faster, but this isn't generally true. And, SWAR is also faster for short strings.

Thanks for this pull request, but I'll close it due to these reasons.

PR #1850, opened 2025-07-09
Add title and required to generated JSON schema

oleid:
This PR adds two  quality-of-life changes (IMHO) to the generated JSON schema.

* Set the name of the type as title
 
    Some JSON schema code generators like oxidecomputer/typify use the title  of a schema as the name of the generated type. Thus, this change sets a sane default for the title.

* non-nullable members as required
    
    Currently, when generating code based on the generated schema files, all members become optional with typical JSON schema code generators and languages like C++ or rust as output target. To better reflect the semantics of the code, this change marks non-nullable object members as required.

### DIFF
--- a/docs/json-schema.md
+++ b/docs/json-schema.md
@@ -55,3 +55,63 @@ struct local_schema_t
 };
 ```
 
+## Required Fields
+
+Glaze can automatically mark fields as required in the generated JSON schema based on their nullability and compile options.
+
+### Automatic Required Fields
+
+By default, when using the `error_on_missing_keys` option, non-nullable fields will be marked as required in the schema:
+
+```c++
+struct user_config
+{
+   std::string username{};  // required (non-nullable)
+   std::optional<int> age{}; // optional (nullable)
+   int id{};                 // required (non-nullable)
+};
+
+// Generate schema with required fields
+auto schema = glz::write_json_schema<user_config, glz::opts{.error_on_missing_keys = true}>();
+```
+
+This will generate a schema with `"required": ["username", "id"]`.
+
+### Custom Required Field Logic
+
+You can override the default behavior by providing a custom `requires_key` function in your type's `glz::meta` specialization:
+
+```c++
+struct api_request
+{
+   std::string endpoint{};
+   std::string api_key{};
+   std::optional<std::string> optional_param{};
+   std::string reserved_field{};  // internal use only
+};
+
+template <>
+struct glz::meta<api_request>
+{
+   // Custom logic to determine which fields are required
+   static constexpr bool requires_key(std::string_view key, bool is_nullable)
+   {
+      // Don't require fields starting with "reserved"
+      if (key.starts_with("reserved")) {
+         return false;
+      }
+      // All non-nullable fields are required
+      return !is_nullable;
+   }
+};
+```
+
+The `requires_key` function receives:
+- `key`: The name of the field being checked
+- `is_nullable`: Whether the field type is nullable (e.g., `std::optional`, pointers)
+
+This allows fine-grained control over which fields are marked as required in the generated JSON schema, useful for:
+- Excluding internal/reserved fields from requirements
+- Making certain nullable fields required based on business logic
+- Implementing complex validation rules
+


oleid:
@stephenberry The tests fail, since the schema has changed. Having had a look at `json_test.cpp` nothing looks fundamentally broken to me. 
If you're okay with this, I'll gladly fix the tests in an additional commit.

stephenberry:
I like this. However, I think there needs to be a compile time option to turn on and off whether all non-nullable fields are required. Most of the use cases I deal with have non-nullable fields that are not required because they have default values for configuration. Often there are hundreds of fields and the user will only specify a few of them, which makes requiring all fields an immense pain for users.

I think that it would be best to also plan to relate this option for requiring all non-nullable keys with the `error_on_missing_keys` option. This doesn't have to be done now, but should be considered for the future.

I think that this compile time option should be handled like the new `rename_key` function that can be defined in `glz::meta`: https://stephenberry.github.io/glaze/rename-keys/

Here is an example of what a new require_key compile time function could look like, which the JSON schema, and later the `error_on_missing_keys` could use:

```c++
struct person {
    std::string first_name{};
    std::string last_name{};
};

template <>
struct glz::meta<person> {
    static constexpr bool require_key(std::string_view key, bool nullable) { 
        if (not nullable) {
            return true;
        }
        return false;
    }
};
```

This would make all non-nullable keys required, but you'd then be able to customize at compile time which specific keys you want to require, or you could require all keys by just returning true. This allows a lot of flexibility and will fit well into the new design of compile time functions within `glz::meta` for customization.

Let me know your thoughts and if you need help implementing this. If you look at how `rename_key` is handled, it should help guide the approach.

oleid:
> I like this. However, I think there needs to be a compile time option to turn on and off whether all non-nullable fields are required. 

That's fine with me.

> Most of the use cases I deal with have non-nullable fields that are not required because they have default values for configuration. 

Indeed, if there are sane default values, this makes sense.
Usually, I need to know if we are dealing with a default value or a user specified value, because I don't want to serialize the default values to disk, but keep them in the binary. Thus, I'm typically making  the values explicitly `optional` and then pass the default to the `value_or()` function of the optional. But most certainly, use cases may be different. 

> I think that it would be best to also plan to relate this option for requiring all non-nullable keys with the `error_on_missing_keys` option. This doesn't have to be done now, but should be considered for the future.

I don't know the implementation details, but semantically they seem quite related. I now check for the existing `error_on_missing_keys` option. And  if it is set, make the member `required`. 


> I think that this compile time option should be handled like the new `rename_key` function that can be defined in `glz::meta`: https://stephenberry.github.io/glaze/rename-keys/
> 
> Here is an example of what a new require_key compile time function could look like, which the JSON schema, and later the `error_on_missing_keys` could use:
> 
> ```c++
> struct person {
>     std::string first_name{};
>     std::string last_name{};
> };
> 
> template <>
> struct glz::meta<person> {
>     static constexpr bool require_key(std::string_view key, bool nullable) { 
>         if (not nullable) {
>             return true;
>         }
>         return false;
>     }
> };
> ```
> 
> This would make all non-nullable keys required, but you'd then be able to customize at compile time which specific keys you want to require, or you could require all keys by just returning true. This allows a lot of flexibility and will fit well into the new design of compile time functions within `glz::meta` for customization.
> 
> Let me know your thoughts and if you need help implementing this. If you look at how `rename_key` is handled, it should help guide the approach.

I'll look into this. For now, I've tied it to `error_on_missing_keys`.

oleid:
@stephenberry Is the most recent commit what you had in mind?

Since I figured the logic if a certain member "is required" could be used outside the schema generation, I created a freestanding function that uses both `Opts` and `meta<T>`. Originally, I wanted it to place it into a `detail` namespace right next to the `concept` I added, but sadly, I need the`nullable_like` concept, which is only defined in the `common` header. Thus, I placed the function there for now and kindly ask for comments on what I came up with :)

oleid:
@stephenberry I know you're probably busy. But in case you just forgot (happens to me all the time) Do you have any thoughts on this?

stephenberry:
Sorry, I have been busy. But, I'll try to look over this in depth next week. Thanks for your patience and work on this!

stephenberry:
@oleid, thanks for your patience. I finally got to look over these small changes and they look great! So, I'm merging it in. Thanks!

PR #1853, opened 2025-07-10
meta_context and skip documentation

stephenberry:
@Yaraslaut, when creating documentation for this new skip functionality, I realized that what we implemented only currently affects writing.

Skipping can also apply to reading. And, I think in some cases we want to only skip writing or only skip reading.

Here are the two approaches I'm thinking of taking.

First option is to simply name the function `skip_write` and a `skip_read` could be added later.
```c++
static constexpr bool skip_write(const std::string_view key) {
    return key == "internal_member";
}
```

> The downside of this approach is that if we wanted to skip both reading and writing then we'd need to implement both functions or have a third function just called `skip`, but having three separate functions complicates the internals of Glaze.

The second option is to make the skip function templated on whether it is a read or write:
```c++
template <bool IsRead>
static constexpr bool skip(const std::string_view key) {
    if (not IsRead) {
       return key == "internal_member";
    }
    return false;
}
```

I'm leaning towards this second option. What are your thoughts?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ Thumbs.db
 # Allow mock test data
 !tests/mock_json_test/json/*.json
 alabastar.beve
+*.jsonc
--- a/docs/skip-keys.md
+++ b/docs/skip-keys.md
@@ -18,7 +18,7 @@ struct my_struct_t {
 
 template <>
 struct glz::meta<my_struct_t> {
-    static constexpr bool skip(const std::string_view key) {
+    static constexpr bool skip(const std::string_view key, const meta_context&) {
         // Return true to skip the key, false to include it
         return key == "internal_member";
     }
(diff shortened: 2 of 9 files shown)


Yaraslaut:
I think having one function is better, but the issue with the template arguments that I see is that user has to see it all the time, I personally would try to add an additional argument to the function aka "context" which the user can use it to figure out if this is a read/write or something else, furthermore the context can be easily extendable

stephenberry:
@Yaraslaut, I implemented a new `meta_context` that is passed to the `skip` function. Right now it just provides information about whether the operation is parsing or serializing, but more meta information could be added in the future, as per your suggestion. Thanks for recommending this, I think it's a great approach! I also implemented support for skipping when reading.

Yaraslaut:
Thanks for the documentation, i also think that this is a great approach :)

PR #1861, opened 2025-07-17
Fix usage of removed function in newer asio versions

azais-corentin:
Hi,

This pull request removes the use of asio::buffer_cast which was deprecated then removed in recent asio versions.

Here's the relevant commit: https://github.com/chriskohlhoff/asio/commit/03f2451af055e5550d04763705dd9dd608f8d1cf

[Boost documentation](https://beta.boost.org/doc/libs/1_82_0/doc/html/boost_asio/reference/buffer_cast.html).


stephenberry:
Thanks!

PR #1863, opened 2025-07-17
Add support for the Boost version of Asio

azais-corentin:
Hi,

This PR adds support for the Boost version of Asio.

`asio::error_code` is is the main difference between the two versions for our usage.

In standalone, it's just a typedef to `std::error_code` but in the Boost version they use `boost::system::error_code` which is it's own class. 
It's convertible to/from std::error_code but since they use reference the conversion can't work without changing our code.

I have verified these changes compile with Asio 1.34.2 and with Boost.Asio 1.88.0.


stephenberry:
Thanks for this!

PR #1867, opened 2025-07-19
Fix incorrect error message on missing required key 

### LINKED ISSUE #1866: Incorrect error message on missing required key when std::optional is present and error_on_missing_keys = true
### Actual 
When using glz::opts{.error_on_missing_keys = true}, Glaze emits an error pointing to the optional key, even when a required (non-optional) key is actually missing.

[Example](https://godbolt.org/z/771YxxEEY)

### Expected:
Error should reference the actual missing required key parentName, not the optional excludeFromCodex.

oleid:
Check if key is actually required when generating error message.

What puzzles me a little is that I couldn't reproduce the issue by simply adding an optional one of the existing test cases. Probably has to do with the hashing of the keys and thus the resulting order. Hence, I added the test case from the issue.

Fixes #1866

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -2243,7 +2243,7 @@ namespace glz
                         constexpr auto req_fields = required_fields<T, Opts>();
                         if ((req_fields & fields) != req_fields) {
                            for (size_t i = 0; i < num_members; ++i) {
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this fix!

PR #1876, opened 2025-07-28
Read toml tests fixes

StormLord07:
### Summary

This PR adds initial support for standard TOML tables (`[table]`) and dotted key access resolution. #1865

Introduced support for parsing and resolving `[table]` headers.
Implemented dotted key resolution during parsing, enabling nested object assignment.
Created a new `toml/opts.hpp` file to define custom parsing options. This allows toggling behavior based on context (e.g., whether parsing occurs in the root or within a table). I'm still not sure if that's the way...
```cpp
            else if (*it == '[') { // Normal table
               std::vector<std::string> path;

               if constexpr (toml::check_is_internal(Opts)) {
                  return; // since we are in internal table we need to return to root.
               }
               else {
                  ++it;
                  ...
```
* When encountering a `[table]`, the parser now returns root context, as required by the TOML spec.
* The function `resolve_nested` is currently used for path resolution. It returns a `bool` for now, though I originally intended it to return a reference to the resolved object but, couldn't figure it out, that'd be better and would remove unnecessary logic for this function when we'll need to add arrays, since for array table defining it like
`[able]`
is not right

### Notes

* This does **not** implement support for table arrays (`[[table_array]]`) yet.
* The ordering of values is still not as per the TOML spec. when writing and relies on struct layout
* Inline table handling (`{ key = value }`) is still based on the old logic and hasn't been refactored.

* The current implementation is functional, but likely not fully optimized.
* This also does not use the fixed size arrays for key length like I wanted [here](https://github.com/stephenberry/glaze/discussions/1865#discussioncomment-13819991)

### DIFF
--- a/include/glaze/toml/opts.hpp
+++ b/include/glaze/toml/opts.hpp
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <glaze/core/opts.hpp>
+
+namespace glz::toml
+{
+
+   enum struct opts_internal : uint32_t {
+      none = 0,
+      internal_struct = 1 << 0, // Currently in an inner struct
+   };
+
+   struct toml_opts
+   {
+      uint32_t format = TOML;
+      bool error_on_unknown_keys{true};
+
+      uint32_t internal{}; // default to 0
+   };
+
+   consteval bool check_is_internal(auto&& o) { return o.internal & uint32_t(opts_internal::internal_struct); }
+
+   template <auto Opts>
+   constexpr auto is_internal_on()
+   {
+      auto ret = Opts;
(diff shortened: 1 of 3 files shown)


stephenberry:
Awesome! Thanks for this work, and even if we just slowly move towards full TOML support forward progress is great.
Are you wanting this to be merged as-is, or do you plan to make more changes to the pull request?

StormLord07:
I think merging as is is ok, I dont know if I'll be able to get back to it for the time being, But i do plan on revisiting it later on, if noone does that beforehand.

stephenberry:
Cool, I'll merge it in!

PR #1887, opened 2025-08-08
Add Boost.Asio workflow

### LINKED ISSUE #1883: Standalone asio include breaks Boost.Asio compatibility
Hi,

Commit 56619189fe077218487bcfaaf38f13e2f5367497 breaks compatibility with Boost.Asio by adding a standalone asio include: 
https://github.com/stephenberry/glaze/blob/5c24f146326580da68047c2180e9570b9c04b82e/include/glaze/net/http_server.hpp#L6

I think this could be fixed by including `"glaze/ext/glaze_asio.hpp"` instead?

azais-corentin:
Fixes #1883

### DIFF
--- a/.github/workflows/boost-asio.yml
+++ b/.github/workflows/boost-asio.yml
@@ -44,7 +44,12 @@ jobs:
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
           echo "deb http://apt.llvm.org/noble/ llvm-toolchain-noble-${{ matrix.compiler.version }} main" | sudo tee /etc/apt/sources.list.d/llvm.list
           sudo apt-get update
-          sudo apt-get install -y --fix-missing clang-${{ matrix.compiler.version }} libc++-${{ matrix.compiler.version }}-dev libc++abi-${{ matrix.compiler.version }}-dev
+          # Clang 17's libc++ does not compile, so we use the default libc++-dev and libc++abi-dev
+          if [ "${{ matrix.compiler.version }}" = "17" ]; then
+            sudo apt-get install -y --fix-missing clang-${{ matrix.compiler.version }} libc++-dev libc++abi-dev
+          else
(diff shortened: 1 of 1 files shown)


azais-corentin:
One of the tests failed and fuzzing failed to link but I'm not sure why.
Could you take a look @stephenberry ?

DockedFerret800:
Why are you only testing with CLang?

azais-corentin:
> Why are you only testing with CLang?

Two reasons mainly:

- I thought the workflows were a bit of mess so I didn't want to touch them too much 😄
- Since we're only testing standalone vs boost asio my reasoning is that it's enough to make sure all the functions/includes work on a single compiler

It wouldn't be too much work to add gcc.

DockedFerret800:
> > Why are you only testing with CLang?
> 
> Two reasons mainly:
> 
> * I thought the workflows were a bit of mess so I didn't want to touch them too much 😄
> * Since we're only testing standalone vs boost asio my reasoning is that it's enough to make sure all the functions/includes work on a single compiler
> 
> It wouldn't be too much work to add gcc.

Well, I believe this should be tested against everything Glaze is currently tested against - so GCC, clang and MSVC. I can submit a commit if you wish.

stephenberry:
Thanks for this pull request! This will allow us to validate that boost-asio is still supported. Others or myself can make the required changes to get everything passing.

The threading test occasionally fails with GitHub Actions. I need to spend more time to figure out why, but on my local machine I could not replicate the issue, so I think it is affected by the virtual environment. I'm going to have to figure out a debug workflow via codespaces I think. But, for now it can be ignored.

azais-corentin:
I'll work on this again soon.
I just wanted to note that we should probably test with a few different Asio/Boost.Asio versions because they've recently removed a few deprecated functions.

azais-corentin:
I have updated the workflow to test with multiple versions of clang (17, 18, 20) and gcc (12, 13, 14), each against two different Boost versions (1.88.0 and 1.73.0).

stephenberry:
This looks good, thanks for the CMake cleanup! Give me the thumbs up and I'll merge it in.

azais-corentin:
You can merge this :)

stephenberry:
Thanks so much for your work on this!

PR #1893, opened 2025-08-15
Remove unneeded CMake VS Code extension from devcontainer

kaladron:
This PR removes the deprecated/unnecessary `twxs.cmake` VS Code extension from `.devcontainer/devcontainer.json`.

Rationale:
- We already use `ms-vscode.cmake-tools` and `clangd`, which cover CMake support and language features.
- Keeping extensions minimal reduces container build time and avoids conflicts.

Change details:
- Delete `twxs.cmake` from the `customizations.vscode.extensions` list in the devcontainer.

No functional changes to the codebase; devcontainer image only.

Testing:
- Container still provisions and relevant tools versions print via `postCreateCommand`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -59,7 +59,6 @@
          "ms-vscode.cpptools",
          "ms-vscode.cpptools-themes",
          "ms-vscode.cmake-tools",
-         "twxs.cmake",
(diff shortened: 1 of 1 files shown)


kaladron:
Specifically, vscode gives a warning at creation time saying that this is deprecated and recommends uninstalling it.

stephenberry:
Thanks!

PR #1894, opened 2025-08-15
devcontainer version updates

kaladron:
Update devcontainer for newer Ubuntu, LLVM and GCC releases.  This prepares for future work to add C++ module support to the build which needs a newer ninja and newer clang.

This also forces the use of libc++ on clang.  The libstdc++ version linked from the apt.llvm.org sources doesn't have the needed support for C++23.

Tested:
Rebuilt container.
Built with clang-debug
Ran ctest with clang-debug


stephenberry:
This looks good, thanks!

PR #1899, opened 2025-08-16
Update CMakePresets.json to use system GCC and G++ compilers

### LINKED ISSUE #1898: CMake GCC Preconfigured options don't work with Codespaces
The CMake presets file defines the GCC paths like this:

                "CMAKE_C_COMPILER": "/opt/homebrew/bin/gcc",
                "CMAKE_CXX_COMPILER": "/opt/homebrew/bin/g++",

However, the CodeSpaces is using a Launchpad PPA to install the specific version of GCC instead of Homebrew, so it's using Ubuntu's alternatives system to put the compiler at /usr/bin/{gcc,g++}

I can adjust the Codespace container to use homebrew for GCC instead of the PPA, or I can update the presets to use the system compiler.  What preference do you have?

(I was just sending another patch to fix up the devcontainer and found this.  All my changes work when this path is fixed)

kaladron:
Closes #1898

Tested:
In Fresh codespace:
cmake --preset gcc-debug
cmake --build --preset gcc-debug
ctest --preset gcc-debug

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -54,8 +54,8 @@
             "generator": "Ninja",
             "cacheVariables": {
                 "CMAKE_BUILD_TYPE": "Debug",
-                "CMAKE_C_COMPILER": "/opt/homebrew/bin/gcc",
-                "CMAKE_CXX_COMPILER": "/opt/homebrew/bin/g++",
+                "CMAKE_C_COMPILER": "/usr/bin/gcc",
+                "CMAKE_CXX_COMPILER": "/usr/bin/g++",
                 "CMAKE_EXPORT_COMPILE_COMMANDS": true,
                 "glaze_DEVELOPER_MODE": true,
                 "BUILD_TESTING": true,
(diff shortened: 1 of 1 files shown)


kaladron:
Eh - this might not actually be the right thing in the end.  In trying to make my modules patch work with GCC as well, I'm getting weird error messages with gcc 14, and the PPA doesn't have GCC 15 yet.  It might make sense to just convert the base image to use homebrew.  I don't love that answer, though. =/

stephenberry:
Should I merge this change or hold off?

kaladron:
Eh, let's merge it.  Homebrew locations are different on Mac and Linux anyway, so I think it's better to just point to the system paths and if people need a newer compiler they can sort it out with a local path change.

I have more running commentary in https://github.com/stephenberry/glaze/issues/1586 of me trying to get homebrew working, and I'm mostly throwing my hands up at this point.  This patch makes the current target of GCC 14 work with the devcontainer config so that way the repository itself is consistent.

PR #1919, opened 2025-08-28
Variant tag validation for auto-deduced structs

stephenberry:
Validates that the type selected via auto-deduction matches any tag that is also included in the input. If no compile time tag is provided the same short-circuiting behavior is applied. If a tag is provided, the entire object must be parsed in order to determine if a tag exists and if it matches the deduced type.


stephenberry:
@GTruf, this adds validation that the tag value matches the selected type for the variant.

PR #1947, opened 2025-09-11
made stencil spacing preservation more consistent inside section iteration ({{#items}} ... {{/items}})

StormLord07:
As i mentioned in #1936 
Currently, glz::stencil does not consistently preserve whitespace when rendering inside section iterations.

Now the test like this:

```cpp
"todo_list"_test = [] {
    std::string_view layout =
        R"({{#items}} {{text}} {{#completed}}✓ {{/completed}}{{^completed}}○ {{/completed}} {{/items}})";

    TodoList list{
        "Mixed Tasks",
        {{"Task 1", false, "high", "home", 1, 0},
         {"Task 2", true, "low", "home", 1, 0}},
        true, 2};

    auto result = glz::stencil(layout, list);
    expect(result == " Task 1 ○   Task 2 ✓  ");
};
```

now passes, with spaces preserved exactly as written in the template.

This may break existing templates that implicitly relied on the old whitespace-collapsing behavior.
For example, templates written like:

```
R"({{#items}}{{text}}, {{/items}})"
```

previously yielded "a,b,c,". After this change, if authors included spaces/newlines around {{/items}}, those will be preserved.

If this doesn't work for you, please tell me, I can introduce `opts` like `preserve_whitespace`. For backwards compatability.


stephenberry:
Thanks for this merge request! I think this makes sense and I'll merge it in.

PR #1949, opened 2025-09-12
fixed compile error when including both toml and json

StormLord07:
A small fix to where char_escape_table is defined for both json and toml, before including both `glaze.hpp` and `toml.hpp` would not compile because of double declaration.

I removed it from `write.hpp` from both `json/` and `toml/` put it into a `util/parse.hpp`, it seemed as best fit there.


stephenberry:
Thanks!

Issue #1873, opened 2025-07-25
Templating http_router to allow broader re-use


nlescoua:
Glaze is really a great library, very versatile for json and other uses. The http_router is also very promising. Currently, the http_router assumes that handler = void(request&, response&). This is perfect for use in the context of glaze::http_server, and works for use with std::future. However, we believe that templating http_router
 
template<typename H = std::function<void(request&, response&)>
class http_router {
   using handler = typename H;
   ...
}
 
would be a great way to allow the use of http_router in other contexts (other futures implementations, coroutine-based servers, or even callback-based servers). Just get/put would make sense in this case, so  get_async/put_async could be enabled/disabled using concepts so that they are present iff handler == std::function<void(request&, response&)>.

stephenberry:
I'm all for making the http_router more generic. Do you want to submit a pull request?

PR #1955, opened 2025-09-16
Templated http_router for broader reuse and using boost-ext/ut instead of openalgz/ut

arun5309:
I have made http_router more generic by templating the handler type to address issue #1873. 

But I had to use [boost-ext/ut](https://github.com/boost-ext/ut) instead of [openalgz/ut](https://github.com/openalgz/ut) as the latter lacked support for [parameterized tests](https://github.com/boost-ext/ut/blob/master/example/parameterized.cpp) to make testing it easier.

I will clean up the commits and possibly move back to openalgz/ut depending on your feedback.


stephenberry:
@arun5309, this look excellent! Yeah, I should add parameterized tests to [openalgz/ut](https://github.com/openalgz/ut). The reason we can't use [boost-ext/ut](https://github.com/boost-ext/ut) is that there are critical multi-threading bugs that cause programs to crash by simply including boost-ext/ut: https://github.com/boost-ext/ut/issues/618

arun5309:
One thing I forgot to mention is that std::move_only_function did not work as a handler (type) that is why I created alt_handler type. Looking into if the handler type needs to be copyable might be of interest.

stephenberry:
Based on this pull request I implemented this in #2151. So, we can close this now. Thanks for contributing this approach and idea!

PR #1962, opened 2025-09-21
TOML Specification conformance for writing, reading arrays [[key]]

StormLord07:
TOML writing/reading more spec-conformant:

* [x] Basic (scalar) types are emitted before tables/structs.
* [x] Partial support added for `[[array-of-tables]]` on read.
* [ ] dotted keys support for AoT (not tested)
* [ ] Multi dimensional array of tables support 

* **Write ordering.** Emit key/value scalars first to avoid interleaving tables and scalars that can confuse TOML readers; tables/structs follow.
* **`[[array-of-tables]]` (read).** Introduces a minimal pass that recognizes and aggregates repeated table headers into arrays. Dotted keys inside these arrays are not yet expanded/tested correctly. 
* **N-dim arrays.** In case of AoT is not yet supported currently i see two approaches.
1. Append "items" to a key
```
[[vector.items.items....]]
key = "value"
```

But i see confusion might arise, if we have a vector called items inside of the struct itself.

2. just inline it.

I want a review or direction of how i could approach this better, currently i put functionality where it seemed best fit, but i dont exactly like it, repetition of code, and it seems logical to put AoT handling for reading in the `reflectable || glz_object_t` op to avoid introducing, completely separate constexpr loop in the vector reading, but you might have a different idea. Also read.hpp line 794-822 behaves completely differently from 892-943 and I completely dont understand why... 

Also coud you please point me to a place where I can see how to itroduce meta specifier so I can mark table as inline.


nickdademo:
Looking forward to this! 👍

StormLord07:
> Looking forward to this! 👍

I'm sorry I'm currently busy with my job, I intent to return to this as soon as I can. possibly need a bit of rewrite of what i already did, I think i jumbled responsibilities of reads/writes quite a lot... So a little bit of a push into right direction would be appreciated

I also really need that in my project. :)

stephenberry:
Thanks for this work! I started from it and implemented #2216, so this work should now be complete.

StormLord07:
> Thanks for this work! I started from it and implemented #2216, so this work should now be complete.

Thank you so much, sorry i couldnt return to this and finish it myself, life and mainly work-related stuff got in a way. Hope my work even in this state was at least a little helpfull.

Also Happy New Year

stephenberry:
It gave me a good idea of what you were going after, so thanks!

PR #1963, opened 2025-09-22
Construct string_view explicitly

CrackedMatter:
This makes writing work for types that have an explicit conversion operator to std::string_view.


stephenberry:
Thanks for this! Could you add a unit test that would fail if this change were not in place? I'm just afraid of your change getting removed at some point because the codebase doesn't see it as required.

stephenberry:
@CrackedMatter, I added unit tests and plan to merge once everything passes.

PR #1964, opened 2025-09-22
Fix interop build

kaladron:
In commit 879857f a number of interop tests were added to the build.  However, they don't seem to build.  This patches fixes it:

 * The tests are built with -fno-exception, but have try and catch statements.
  * Variable types were defined as iglaze and ivalue instead of i_glaze and i_value.

With this patch, the file now builds, although the following tests fail:
 * interop_complex_test
 * interop_member_function_test
 * interop_zero_copy_test

I don't know enough about what's happening to explore these.  I found these when working on something unrelated.


stephenberry:
Yeah, this was my bad, I thought the CI was running tests and changed naming and it wasn't. Thanks for your help on this. I'll explore in more detail.

stephenberry:
I've decided to remove the Interop code in v6. I ended up going down the route of a pure C interface for broader use and I don't want to manage all this code. It could be broken into another repository if someone needed it, but I'm removing it from Glaze, so this pull request can close.

PR #1970, opened 2025-09-23
More detailed status for http errors

stephenberry:


klaussilveira:
This looks good! :+1:

klaussilveira:
I know this is an edge case, but would it be possible to customize what is considered an error code via a callback or an argument? The reason behind this is that Typesense has a very unusual API:

https://typesense.org/docs/29.0/api/documents.html#index-multiple-documents

It returns 200 in a chunked response, with each line of the response being a JSON document containing true/false inside success. I've been trying to handle this with glaze and assuming error > 400 would get in the way.

stephenberry:
@klaussilveira, thanks for the feedback. I'll allow error code customization before merging this.

stephenberry:
@klaussilveira, I added an optional `status_is_error` customizable function, so that you can change what status codes are considered errors. Does this look good to you?

klaussilveira:
That's awesome! Definitely works.

stephenberry:
Thanks! Merged

PR #1991, opened 2025-09-29
BEVE support for simple value types in map keys

polyxemus:
Adds support for simple value types with explicit glaze meta to be serialized in BEVE. Use case is serialization of a map which has "strong" IDs as keys that are specifically _not_ arithmetic.

# before
I have employed a workaround for this which I don't particularly like.

```c++
namespace glz {

template <>
struct from<BEVE, ModuleID> {
    template <auto Opts>
        requires(not glz::check_no_header(Opts))
    static void op(ModuleID& value, is_context auto&& ctx, auto&& it, auto&& end) {
        parse<BEVE>::op<Opts>(value.value, ctx, it, end);
    }

    template <auto Opts>
        requires(glz::check_no_header(Opts))
    static void op(
        ModuleID& value, const uint8_t tag, is_context auto&& ctx, auto&& it, auto&& end
    ) {
        // The BEVE map code calls us with tag::string because ModuleID is not arithmetic.
        // But our writer (no_header) wrote the raw uint64_t bytes directly.
        // So we just need to read those 8 raw bytes.
        (void)tag;

        if ((it + sizeof(value.value)) > end) [[unlikely]] {
            return;
        }

        // Read the raw uint64_t value that was written by no_header
        std::memcpy(&value.value, it, sizeof(value.value));
        it += sizeof(value.value);
    }
};

template <>
struct to<BEVE, ModuleID> {
    template <auto Opts>
    static void op(const ModuleID& value, is_context auto&& ctx, auto&& b, auto&& ix) {
        serialize<BEVE>::op<Opts>(value.value, ctx, b, ix);
    }

    template <auto Opts>
    static void no_header(
        const ModuleID& value, is_context auto&& ctx, auto&& b, auto&& ix
    ) {
        serialize<BEVE>::no_header<Opts>(value.value, ctx, b, ix);
    }
};
```

# after
With this proposed change, it would instead require:

```c++
template <>
struct meta<ModuleID> {
    static constexpr auto value{&ModuleID::value};
};
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ include(cmake/prelude.cmake)
 
 project(
     glaze
-    VERSION 5.7.2
+    VERSION 6.0.0
     LANGUAGES CXX
 )
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Glaze also supports:
 
 > [!IMPORTANT]
 >
-> ## Breaking changes coming in version 6.0.0
+> ## Breaking v6.0.0 changes
 >
 > - `glz::json_t` has been renamed to `glz::generic` and will be deprecated in v6.0.0. Update your code to include `glaze/json/generic.hpp` and prefer `glz::generic` to stay aligned with the upcoming release.
 > - Removed `v5.6.0` Glaze C FFI interop. This was a significant experiment that looked like would take off and be extremely useful, but after attempting to use it in production it became clear that the developers wouldn't use this feature and instead create a low-level C API. Someone could create a third party library with the code, but it has been removed from Glaze to focus on more critical features.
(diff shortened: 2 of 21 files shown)


stephenberry:
There is already a `glz::cast`, which could probably be used here:

```c++
template <>
struct glz::meta<ModuleID> {
    static constexpr auto value = glz::cast<[](ModuleID& s) -> auto& { return s.value; }, uint64_t>;
};
```

This works as long as `static_cast<uint64_t>` can be used on your `.value`.

What are your thoughts on this other approach?

polyxemus:
When I try with:

```c++
template <>
struct glz::meta<ModuleID> {
    static constexpr auto value = glz::cast<[](ModuleID& s) -> auto& { return s.value; }, uint64_t>;
};
```

The first issue I hit is:

```
glaze-5.7.2/include/glaze/beve/write.hpp:33:53: error: no member named 'no_header' in 'glz::to<1, ModuleID>'
   33 |          to<BEVE, std::remove_cvref_t<T>>::template no_header<Opts>(std::forward<T>(value), std::forward<Ctx>(ctx),
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~         ^
```

I'm probably missing something, but in order to get it to compile I ended in the same spot with `from/to` specializations for `ModuleID`, and then back to the beve read/write mismatch.

I'm still not convinced this isn't user error on my part, but I can't see how:

```c++
   template <writable_map_t T>
   struct to<BEVE, T> final
   {
      template <auto Opts, class... Args>
      static auto op(auto&& value, is_context auto&& ctx, Args&&... args)
      {
         using Key = typename T::key_type;

         constexpr uint8_t type = str_t<Key> ? 0 : (std::is_signed_v<Key> ? 0b000'01'000 : 0b000'10'000);
         constexpr uint8_t byte_cnt = str_t<Key> ? 0 : byte_count<Key>;
         constexpr uint8_t tag = tag::object | type | (byte_cnt << 5);
         dump_type(tag, args...);

         dump_compressed_int<Opts>(value.size(), args...);
         for (auto&& [k, v] : value) {
            serialize<BEVE>::no_header<Opts>(k, ctx, args...);
            serialize<BEVE>::op<Opts>(v, ctx, args...);
         }
      }
   };
```

given a type of `std::map<ModuleID, string>` there's a place for the cast logic to intercept the type handling of the key.

polyxemus:
Superseded by #2002

PR #1997, opened 2025-10-06
CORS preflight requests succeed w/o user stubs

stephenberry:
- Implicitly handle preflight OPTIONS requests inside http_server, discovering sibling route methods and constructing the Allow header on the fly so that browsers receive a consistent view of which verbs are valid without developers wiring up placeholder handlers.
- Run the full middleware chain during these synthetic responses so CORS and other filters can stamp the preflight exchange, and tighten the logic to only return success when the requested verb is actually registered—mismatched methods now yield a proper 405.
- Grow the CORS feature surface with dynamic origin validation (callbacks and glob/regex helpers), allow-all conveniences for headers/methods, configurable success status and private-network support, credential safety warnings, and additive helpers for header exposure lists.
- Extend the HTTP client test harness to issue OPTIONS requests and add regression coverage for both the happy-path 2xx preflight and the new 405 safeguard, keeping the whole suite green.

### DIFF
--- a/include/glaze/net/http_server.hpp
+++ b/include/glaze/net/http_server.hpp
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <atomic>
 #include <charconv>
 #include <chrono>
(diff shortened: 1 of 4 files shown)


stephenberry:
@azais-corentin, I worked with AI on these CORS preflight improvements. Could you look over the changes or test with this `pre-flight-cors` branch to see if this will satisfy your needs?

azais-corentin:
I belive this should be kept as simple as possible.
I have cleaned up the code and removed a few options I deemed unnecessary.

Could you please pull my changes at https://github.com/azais-corentin/glaze/tree/pre-flight-cors ? Alternatively, I can open another pull request.

azais-corentin:
I have tested my branch and it seems to be working on a few basic cases.

stephenberry:
@azais-corentin, can you submit a pull request with your changes? That will let me easily merge them into here.

azais-corentin:
Here it is: https://github.com/stephenberry/glaze/pull/1998

stephenberry:
@azais-corentin, thanks so much for your cleanup! I do think simpler is better as we can carefully add more features if they are needed in the future. I merged in your changes and I think it should be good to go if everything passes.

include/glaze/net/http_server.hpp
```diff
@@ -1158,17 +1160,119 @@ namespace glz
...
+
+               try_method(http_method::GET);
+               try_method(http_method::POST);
+               try_method(http_method::PUT);
+               try_method(http_method::DELETE);
+               try_method(http_method::PATCH);
+               try_method(http_method::HEAD);
+
+               if (!allowed_methods.empty()) {
+                  request.params = std::move(preflight_params);
```
azais-corentin, line 1194: I'm not sure what this achieves and why we need it

PR #2002, opened 2025-10-13
Strong ID and deeper cast support for use with map keys with BEVE

stephenberry:


stephenberry:
@polyxemus, this pull request is much like your #1991 pull request, but also adds deeper support for `glz::cast`. So, both your preferred syntax works and `glz::cast` works for strong ID types. Let me know if you think this looks good. Thanks for pointing a way forward with your pull request.

polyxemus:
Perfect! Verified this works on my end.

PR #2007, opened 2025-10-16
Improve parsing for TOML integers

### LINKED ISSUE #2003: Support for non-base 10 numbers in the TOML format
Beyond the typical base ten, the TOML spec supports specifying numbers in hexadecimal, binary, and octal format, with the `0x`, `0b`, and `0o` prefixes respectively. Currently, it seems that Glaze's TOML implementation reads integers with `glz::atoi` which parses numbers according to the JSON specification, which is more limited. Adding this support would likely be useful for other formats as well. The "libconfig" format for instance allows similar prefixes, with the addition of being able to use `0q` as well as `0o` to specify octal.

Specifically, my use-case is that I would like to use TOML to specify properties of a theme, and for color properties to be able to be specified like `0xAABBCC` instead of like `11189196`.

(Thank you as well for creating and maintaining Glaze, it is very impressive and the ergonomics are astounding.)

friedkeenan:
Closes #2003.

I feel like the code could likely be condensed, but I didn't feel like I could do much more without adding too much convolution. Open to suggestions, if you think it needs some work that way.

### DIFF
--- a/include/glaze/toml/read.hpp
+++ b/include/glaze/toml/read.hpp
@@ -302,35 +302,6 @@ namespace glz
          return is_any_toml_digit(c);
       }
 
-      constexpr bool is_end_of_toml_decimal_zero(auto&& it, auto&& end) noexcept
-      {
-         // The character before 'it' should be checked to be '0'.
-
-         if (it == end) {
-            return true;
-         }
-
-         // Underscore after the zero indicates either
-         // that the zero was a leading zero, or that
-         // the underscore is a trailing underscore.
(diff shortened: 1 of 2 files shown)


stephenberry:
@friedkeenan, thanks for this pull request! I just pushed a significant cleanup to your code that also adds a bunch more unit tests. Let me know what you think.

friedkeenan:
Tests look great to me, and your changes look fine to me as well. If you're happy with it, I'm happy with it. Thanks for your help.

PR #2051, opened 2025-11-09
MSVC fix

stephenberry:
Use subtraction from zero instead of unary minus to avoid MSVC C4146 error


pps83:
that's a proper fix. Shouldn't it be the same code for all compilers? Shouldn't it generate identical code?

PR #2063, opened 2025-11-14
Allow http_server to run on external io_context.

gobdevel:
1. **Add an additional `http_server` constructor to support external `io_context` instances.**
2. **Support both internally owned and externally provided `io_context` objects.**
3. **Update `start()` API to allow running without creating worker threads.**
4. **Add test cases to verify the new constructor, `io_context` handling, and `start()` behavior.**

### DIFF
--- a/docs/networking/advanced-networking.md
+++ b/docs/networking/advanced-networking.md
@@ -263,33 +263,102 @@ void run_dual_servers() {
     http_server.get("/redirect", [](const glz::request& req, glz::response& res) {
         res.status(301).header("Location", "https://localhost:8443" + req.target);
     });
-    
+
     // HTTPS server
     glz::https_server https_server;
     https_server.load_certificate("cert.pem", "key.pem");
     setup_secure_routes(https_server);
-    
+
     // Start both servers
     auto http_future = std::async([&]() {
         http_server.bind(8080);
         http_server.start();
     });
-    
+
     auto https_future = std::async([&]() {
         https_server.bind(8443);
         https_server.start();
     });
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this pull request and I think the idea is sound. But, this implementation seems a bit more complex than it needs to be.

I think the io_context could just be a shared_ptr. And, the user could provide a custom destructor if they don't want the asio_server deleting the context on destruction.

unique_ptr needs to go away. But, a shared_ptr would allow an outside context to be passed in or allocated internally by default.

This would eliminate the additional optional and reference_wrapper.

You can make this change or I'll try to get to it next week. Thanks!

gobdevel:
Thanks, I will do this.
I wanted to know about the testing infra little more. It looks like the test cases are not fully validating the end to end flow. How can i test after making changes that every things works as expected. Also how to run all test cases in one go ?
A developer doc would be highly recommended. let me know if its already there.

stephenberry:
I need to add more developer documentation. Anything using the unit test library is run in the CI. There are a few examples that are just built (tested for build passing) and ought to be unit tested.

There are CI tests that execute the asio_server. So, this change should be tested. But, I might add some more next week to this pull request.

stephenberry:
I tweaked the start input to take in an optional of the size, so that default using the hardware concurrency still works. But, a size of `0` works great with external io_context

stephenberry:
I'm merging this in, because I've added documentation and I think it's ready. Feel free to submit more pull requests to tweak things.

PR #2079, opened 2025-11-22
Add glz::escape_bytes_t

stephenberry:
By default Glaze treats char arrays as null-terminated C-strings during JSON serialization. This behavior will cause data truncation when char arrays contained null bytes (`\0`), causing incorrect output (e.g., char arr[4] = {0, 0, 1, 0}; would serialize to "").

# New glz::escape_bytes_t wrapper:
1. `glz::escape_bytes_t` (Runtime Wrapper): A new class template that allows direct wrapping of char arrays or std::vector<char> to explicitly signal that their content should be serialized and deserialized as a fully escaped JSON string. This ensures all bytes, including nulls and control characters, are preserved.
2. `glz::escape_bytes` (Meta Wrapper): A corresponding variable template for use with glz::meta (reflection), enabling specific struct members to be declared as escaped byte sequences.

**Local Usage (`glz::escape_bytes_t{value}`):**

```c++
char local_data[4] = {0, 'A', 0, 'B'};
std::string out;
glz::write_json(glz::escape_bytes_t{local_data}, out);
expect(out == R"("\u0000A\u0000B")");

char read_back_data[4];
glz::read_json(glz::escape_bytes_t{read_back_data}, out);
expect(read_back_data[0] == 0);
expect(read_back_data[1] == 'A');
expect(read_back_data[2] == 0);
expect(read_back_data[3] == 'B');
```

### DIFF
--- a/tests/json_test/byte_array_test.cpp
+++ b/tests/json_test/byte_array_test.cpp
@@ -28,7 +28,7 @@ suite byte_array_tests = [] {
     "default_char_array_behavior"_test = [] {
         char arr[4] = {0, 0, 1, 0};
         std::string json;
-        glz::write_json(arr, json);
+        expect(not glz::write_json(arr, json));
         // arr is treated as null-terminated string, so it stops at first null.
         expect(json == "\"\"") << "Default char arrays stop at null";
     };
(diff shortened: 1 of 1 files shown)


kaladron:
Thank you!  This got buried in my stack, I appreciate it!

stephenberry:
You're very welcome!

PR #2101, opened 2025-11-30
Fix clang build with structs embedding vector of incomparable variants

makekryl:
Hello

Those `tuple_elem` comparison operators specializations make clang build fail. I guess the two previous default operators already cover such cases so the specializations are not needed.

See https://godbolt.org/z/3z5zrTs3K

If I understand correctly, it is actually a GCC oversight that such code compiles with it. Clang eagerly instantiates template, but instead of expected SFINAE behavior, we are getting a compilation error as it believes that the program is ill-formed.

### DIFF
--- a/include/glaze/tuplet/tuple.hpp
+++ b/include/glaze/tuplet/tuple.hpp
@@ -113,17 +113,6 @@ namespace glz
          constexpr decltype(auto) operator[](tag<I>) && { return (std::move(*this).value); }
          auto operator<=>(const tuple_elem&) const = default;
          bool operator==(const tuple_elem&) const = default;
-         // Implements comparison for tuples containing reference types
-         constexpr auto operator<=>(const tuple_elem& other) const noexcept(noexcept(value <=> other.value))
-            requires(std::is_reference_v<T> && ordered<T>)
-         {
-            return value <=> other.value;
-         }
-         constexpr bool operator==(const tuple_elem& other) const noexcept(noexcept(value == other.value))
-            requires(std::is_reference_v<T> && equality_comparable<T>)
-         {
-            return value == other.value;
-         }
       };
    } // namespace tuplet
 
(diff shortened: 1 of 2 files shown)


stephenberry:
@makekryl, this actually seems like a `libc++` (standard library) bug. Note that if you use Clang 21 your code builds and works: https://godbolt.org/z/38z9P33vr

---

> I guess the two previous default operators already cover such cases so the specializations are not needed.

No, we lose comparison with reference values if we remove our special cases, because if structs contain references their default comparison is removed. The compiler cannot automatically generate them because the semantics of comparing references (address vs. value) are not strictly defined for defaulted operators in this context, or rather, the standard decided not to support it implicitly.

---

Could you update the pull request to avoid the `<=>` operator specialization for Clang versions before 21? Would this solve the problem for you?

makekryl:
Thank you for the explanation, I did not think about that

> Could you update the pull request to avoid the <=> operator specialization for Clang versions before 21? Would this solve the problem for you?

I'm afraid putting this under ifdef would result in inconsistent behavior and confuse people; I'd update to clang 21 to solve this problem for me. I hope if anyone runs into this problem, Google will lead them to this thread

On the other hand, compilation succeeds if I comment out the `ordered<T>` , what do you think about putting it under ifdef?

stephenberry:
Yeah, the `ordered<T>` could be put under an ifdef

makekryl:
Sorry, unfortunately, it did not work with more complicated case, failing with the same error on the `operator<=>` body - turns out when I was initially testing the proposed fix within my project, I did not comment out `ordered<T>`, but added `false &&` to the begin of the constraints list, efficiently removing the overload, i.e. it would never compile on clang < 21 :(

stephenberry:
But, moving to Clang 21 works for you?

makekryl:
Haven't tested yet as it is not available in my package registry, I'll try to update soon and respond

Issue #1837, opened 2025-06-28
Fix cmake plz


Arniiiii:
I'm a maintainer of gentoo's recipe to build glaze: https://github.com/Arniiiii/ex_repo/tree/master/dev-cpp/glaze

Could you fix some parts of cmake plz? 

Here's patches: https://github.com/Arniiiii/ex_repo/tree/master/dev-cpp/glaze/files 

[`0003_no_to_quiet_find_package.patch`](https://github.com/Arniiiii/ex_repo/blob/master/dev-cpp/glaze/files/0003_no_to_quiet_find_package.patch)
Just don't be quiet. Logs are great.

[`0004_fix_bundling_asio_and_add_tests_execution.patch`](https://github.com/Arniiiii/ex_repo/blob/master/dev-cpp/glaze/files/0004_fix_bundling_asio_and_add_tests_execution.patch)
I've changed from `FetchContent` to `find_package` , but that's not the main issue: the cmake for bundling asio was all over networking tests.
Also some networking tests were not enabled because of absence of `add_test(...)`

[`0005_fix_erland_cmake_test_option_name.patch`](https://github.com/Arniiiii/ex_repo/blob/master/dev-cpp/glaze/files/0005_fix_erland_cmake_test_option_name.patch)
Just consistency with main option for erlang

[`0007_fix_using_installed_glaze_with_eetf-format_enabled.patch`](https://github.com/Arniiiii/ex_repo/blob/master/dev-cpp/glaze/files/0007_fix_using_installed_glaze_with_eetf-format_enabled.patch)
You see, if your target has dependency, it is preferable to have scripts for finding dependencies inside of i.e. `glazeConfig.cmake` . I've just fixed it in this way, but this can be fixed using [PackageProject.cmake](https://github.com/TheLartians/PackageProject.cmake) or some techniques from it.

stephenberry:
Thanks for these notes. I'm currently occupied for a few weeks, so I won't get to this immediately. But, I'll try to address this as soon as I'm able.

carlo711:
Hi, would also appreciate if all tests deps would be declared in tests/CMakeLists.txt. As @Arniiiii said if you remove the network tests subdir it will fail to build as it can not link against glz_asio.

PR #2112, opened 2025-12-05
CMake Improvements

stephenberry:

PR #2114, opened 2025-12-08
Fix:CMake:looking for Erlang if `glace_EETF_FORMAT` during `find_package`

Arniiiii:
#1837


Arniiiii:
I haven't seen work at #2112 , sorry, going to close.

PR #2157, opened 2025-12-18
JSON Schema default value deduction

### LINKED ISSUE #1296: Can JSON schema defaults pull from the defaults on the struct if not otherwise specified?

stephenberry:
## Summary

Adds opt-in automatic default extraction for JSON Schema generation. When enabled, Glaze reads each primitive member's default-constructed value and emits it as the schema's `"default"` keyword. **Off by default** — C++ default-initialization and JSON Schema's `"default"` keyword have overlapping but not identical semantics, so making this implicit for every reflectable struct would silently change the meaning of every generated schema.

## Usage

Opt in via a custom opts struct:

```cpp
struct my_opts : glz::opts { bool schema_auto_defaults = true; };
auto schema = glz::write_json_schema<MyType, my_opts{}>();
```

Without the flag, schema output is unchanged from main.

## What changes from main

- **`schema_auto_defaults` opt**: new `check_schema_auto_defaults(Opts)` helper detects the bool on the caller's opts. Default is `false`, so existing call sites (and snapshot tests) are unaffected.
- **Primitive extraction when enabled**: constructs `T{}` at compile time and reads each member's value. Fields of type `bool`, integer, floating-point, and `std::monostate` are emitted as `"default"`. `std::string`, containers, and other non-representable types are skipped — their storage would not outlive the transient consteval `T{}`.
- **Value-init filter**: a member is emitted only if its value differs from `val_t{}`. `int x{}`, `bool enabled{false}`, and `double x{0.0}` produce no `"default"` (they are sentinel-init, not deliberate recommendations); `flag{true}`, `count{42}`, `ratio{3.14}` still do. Users who specifically want `"default":0` can set it explicitly via `glz::json_schema<T>`.
- **Single extraction per type**: `T{}` is constructed once; all member defaults are pulled in one index-pack pass and stored in an `inline constexpr auto cached_defaults<T>` variable template, so repeated accesses are free.
- **Explicit overrides always win**: an explicit `glz::json_schema<T>` with `.defaultValue = ...` takes precedence regardless of the opt-in flag.
- **Integrates with main's schema refactor**: plugs into main's post-refactor code paths (inline `bool`/`string` primitives, `anyOf` for nullables, `prefixItems` for tuples, reference-counted single-use inlining).
- **Docs**: `docs/json-schema.md` gets a new "Automatic Default Extraction (opt-in)" section describing the flag, the extraction scope, the value-init filter, and the explicit-override precedence.

## Test plan

- [x] New `auto_defaults`, `mixed_defaults`, `explicit_override`, `nested_defaults`, `value_init_defaults` tests in `jsonschema_test.cpp` cover extraction, primitive filtering, explicit-override precedence, defaults on inlined single-use definitions, and the value-init filter — all go through the opt-in opts struct
- [x] Default-off test locks in the "no `"default"` emitted without opt-in" contract
- [x] Pre-existing schema tests across `example_json`, `exceptions_test`, `json_reflection_test`, `json_test`, and `jsonschema_test` produce identical output to main
- [x] Full local test suite passes (92 tests) on AppleClang


wilkolbrzym-coder (changes requested):
Review – PR #2157

Not ready. Branch has unresolved conflicts with main; rebase first.

Concept is fine. Opt-in, default off is correct.

Bugs to fix:

1. uint64 defaults above INT64_MAX

to_schema_default stores unsigned as uint64_t.

Schema parser reads positive integers as int64_t.

Example: 18446744073709551615 will not round-trip.

Filter to int64_t max or skip.

2. Non-finite float defaults

double d = infinity; passes the zero-filter because infinity != 0.

Would emit invalid JSON Schema.

Filter with std::isfinite.

3. Character-like types

Only char is excluded.

wchar_t, char8_t, char16_t, char32_t remain integral and may emit large integers as defaults.

Exclude all character types.

4. Negative zero

-0.0 == 0.0, so it is filtered.

If intentional negative zero is desired, it requires explicit json_schema override.

Document this.

5. Non-constexpr destructor path

can_extract_defaults correctly falls back.

But no test covers default-constructible type with non-constexpr destructor.

Add compile-only test.

6. Rebase must preserve recent schema refactor

19-commit old branch.

Rerun schema-related tests after rebasing, especially GCC 13.

Verdict

Request changes.

Fix conflict, uint64 range, non-finite floats, and character-type handling.

cc @stephenberry

PR #2169, opened 2025-12-21
fix invalid control code in string

huangminghuang:
Fixes parsing of strings that contain an invalid control character sequence. Currently, only the test with long tail will succeed while the tests with short/middle tail fail.


stephenberry:
Thanks so much for this fix! I added the fix to the `std::u8string` code as well.

Issue #2190, opened 2026-01-02
std::string Memory Bomb (DoS) vulnerability


ElevationsRPG:
Description:
Currently, when decoding BEVE (Binary) data into a std::string or std::vector container with strings or custom struct (with std::string inside it), the decoder reads the integer length header and immediately calls .resize() on the target container.

In a server-side context, this creates a Memory Bomb (DoS) vulnerability. A malicious client can send a valid BEVE header specifying a massive length (e.g., 4GB), causing the server to crash with std::bad_alloc or be terminated by the OOM killer before any validation constraints (like read_constraint) can be executed.

The Problem:
While glz::read_constraint exists, it is evaluated after the field has been materialized (allocated and read). For binary formats where the length is known upfront, we need a way to validate that length header before the allocation happens.

Current Behavior (Simplified):

```
1. Read length L from buffer.

2. call str.resize(L); // Crash occurs here if L is malicious.

3. Read L bytes into str.

4. Run read_constraint (Too late).
```

Proposed Solution:
Introduce a way to define a maximum length at the metadata level that the BEVE decoder respects before calling resize().

Potential Syntax Ideas:

New Wrapper:

C++

```
//manually wrap your strings in glz::max_len

struct my_struct {
   glz::max_len<std::string, 128> username; 
};


or define a metadata Attribute:

template <>
struct glz::meta<my_struct> {
   static constexpr auto value = object(
      "username", [](auto& s) { return s.username; }, 
      glz::opts{.max_len = 128} // Hypothetical opt
   );
};
```

Impact:
Without this, Glaze cannot be safely used for high-performance server-side binary networking without manual "peeking" logic or wrapping every string in a glz::custom lambda to validate the length header manually.

stephenberry:
Having a max length option is good, but I think a more general solution is to check agains the length of the input buffer. The input buffer length can be controlled by the server and be guarded against malicious attacks. We can then check all our resizes for vectors and strings agains the length of the input buffer. If it is beyond this length we can short circuit and error, knowing that we have an erroneous or malicious value. If it is a malicious size that is within the buffer length there is no threat of a resize bomb, and a parse failure would occur further down the line, which would short circuit after the safe memcpy.

In short, we can check the number of bytes needed for the resize against the length of the input buffer and ensure that only as much data as could possibly be in the message will be allocated. This lets the server limit the allocation size and avoids this threat.

[...]

Issue #2187, opened 2025-12-31
BEVE Parsing Security (Verification/Validation)


ElevationsRPG:
Currently i do not see a way where we can validate what we are about to parse for example lets say we are a malicious user spoofing

```
std::vector<int> example;
example.push_back(1000000000000);//some ridiculous amount of elements
```

//The normal code to read_beve //We just parsed a giant amount of ints with no validation before the parse?

```
std::string buffer = /* delimited BEVE data */;
size_t offset = 0;

while (offset < buffer.size()) {
   std::vector<int> obj;
   auto result = glz::read_beve_at(obj, buffer, offset);
   if (!result) {
      break;  // Error or end of data
   }
   offset += *result;  // Advance by bytes consumed

   // Process obj...
}
```

stephenberry:
You make a good point that there should be a configurable option for the maximum number of elements allowed.

I think you meant to call `resize` in your example rather than `push_back`. This resize is what needs to be guarded. If the size is ridiculous then the allocator will typically throw.

Thanks for pointing this out.

ElevationsRPG:
> You make a good point that there should be a configurable option for the maximum number of elements allowed.
> 
> I think you meant to call `resize` in your example rather than `push_back`. This resize is what needs to be guarded. If the size is ridiculous then the allocator will typically throw.
> 
> Thanks for pointing this out.

Yes i some what mitigated this with the partial_read option which did work but i had to manually resize my vector

```
std::vector<int> example; 
example.resize(100); //this forces 100 elements when doing with option partial_read
```

 **However!** it does not seem to force limits on std::string (on its own) so there's a problem with partial_read now too!


```
            std::string buffer;
            std::string example = "hello world test test test test test test and this is the end of the string";
            glz::write_beve(example, buffer);

            std::string example2; 
            example2.resize(5);
            glz::read_beve<glz::opts{.partial_read = true}>(example2, buffer); //THIS gives us the ENTIRE string back! UH OH

           //INCOMING 10GB std::strings! ! !
```

We definitely need another option or some more specifically "beve_read_header" functions will be nice so that can inspect the header before deciding to parse it or not.

I also like using std::read_beve_at so even just a glz::opts::size option to force a container/string limit would be ideal. I'm currently intending to use the BEVE format for network packets which is why this type of validation is important to me.

Oh no i also just realized that if we did:

```
std::vector<std::string> example; //How do we detect what the size of the std::strings are inside the vector? (They could be 20 GB)
```

[...]

PR #2194, opened 2026-01-03
Secure BEVE from DoS attacks with invalid lengths

stephenberry:
## BEVE Security: Pre-allocation Bounds Checking

Addresses #2187 and #2190 - prevents memory bomb (DoS) attacks in BEVE binary parsing.

### Problem

When parsing BEVE data, length headers were used to `resize()` containers before validating that the buffer contained sufficient data. A malicious actor could send a buffer claiming billions of elements but containing only a few bytes, causing memory exhaustion before validation could occur.

### Solution

Added pre-allocation bounds checks that validate length headers against remaining buffer size **before** any memory allocation:

- **Boolean arrays**: Validates `(count + 7) / 8` bytes available
- **String arrays**: Validates at least `count` bytes available (1 byte minimum per string header)
- **Generic arrays**: Validates at least `count` bytes available (1 byte minimum per element tag)

### Changes

**`include/glaze/beve/read.hpp`**
- Added bounds checks before `resize()` for boolean, string, and generic arrays

**`include/glaze/beve/header.hpp`**
- Fixed `int_from_compressed()` to return errors instead of silent failures:
  - 32-bit systems receiving 8-byte length encoding: `invalid_length`
  - Invalid config values: `syntax_error`

**`include/glaze/core/context.hpp`**
- Added `invalid_length` error code for length validation failures

**`include/glaze/core/error_category.hpp`**
- Added `buffer_overflow` and `invalid_length` to error code mappings

**`docs/security.md`**
- New documentation covering DoS protection and security best practices

### DIFF
--- a/tests/beve_test/beve_test.cpp
+++ b/tests/beve_test/beve_test.cpp
@@ -4166,42 +4166,6 @@ suite dos_prevention = [] {
       expect(bool(ec)) << "Should reject truncated numeric array buffer";
    };
 
-   "valid data still parses after security checks"_test = [] {
-      // Verify that legitimate data still works correctly
-
-      // String array
-      std::vector<std::string> str_original = {"hello", "world", "test"};
-      std::string str_buffer;
-      expect(not glz::write_beve(str_original, str_buffer));
-      std::vector<std::string> str_result;
-      expect(!glz::read_beve(str_result, str_buffer));
-      expect(str_result == str_original);
-
-      // Bool array
-      std::vector<bool> bool_original = {true, false, true, true, false};
-      std::string bool_buffer;
-      expect(not glz::write_beve(bool_original, bool_buffer));
-      std::vector<bool> bool_result;
-      expect(!glz::read_beve(bool_result, bool_buffer));
-      expect(bool_result == bool_original);
-
-      // Numeric array
-      std::vector<int> int_original = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
-      std::string int_buffer;
-      expect(not glz::write_beve(int_original, int_buffer));
-      std::vector<int> int_result;
-      expect(!glz::read_beve(int_result, int_buffer));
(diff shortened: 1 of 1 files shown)


ElevationsRPG:
Hi Stephen,

I appreciate you implementing the checks for spoofed/truncated length headers. However, I want to emphasize that a global MAX_MESSAGE_SIZE check is still insufficient for complex or high-scale applications.

While it stops truncated headers, it doesn't protect against a 'Real Bomb' a payload where the headers and data are both perfectly valid and match the 1GB size, but the content is malicious.

The 'Blind Limit' Problem: If my application requires a 1GB limit to support legitimate world saves or terrain data, I have to open my network gate to 1GB. But a global buffer limit cannot distinguish between Volume and Complexity:

Legit Data: 1GB of flat terrain data = 1 allocation/operation (Safe).

Attack Data: 1GB of 50 million map entries = 50 million allocations/insertions (Server Crash).

A global limit is too blunt. It protects the buffer, but it doesn't protect the CPU and RAM from being exhausted by the sheer number of operations triggered inside that buffer. We need a Peek API that acts as a 'Verifier.'

By having a verification step, we can inspect the internal size/count headers of containers before Glaze begins the actual deserialization work. This provides the granularity needed to enforce 'Semantic Limits'—for example, allowing a 1GB buffer for raw binary data while simultaneously rejecting a 1GB buffer if it contains more than 10,000 map entries.

The invalid length check is a step in the right direction, but without the ability to verify and enforce our own container-level limits, the system remains vulnerable to resource exhaustion.

A simple solution to fix this would be a glaz::Verifier implementation before the glz::read step.

stephenberry:
@ElevationsRPG, yeah I understand the need for more precise limits and validation. I'm working on another pull request that implements this finer control: https://github.com/stephenberry/glaze/pull/2195. This initial PR just provides a default protection for users who don't define limits.

PR #2195, opened 2026-01-03
max_length wrapper and options

stephenberry:
## User-Configurable Allocation Limits for Binary Formats

Adds extendable compile-time options and a per-field wrapper for limiting string lengths, array sizes, and map sizes when reading BEVE and CBOR formats, providing additional protection against memory exhaustion attacks.

### Extendable Options

Options are extendable - inherit from `glz::opts` and add the fields you need:

```cpp
struct secure_opts : glz::opts
{
   uint32_t format = glz::BEVE;  // or glz::CBOR
   size_t max_string_length = 1024;    // Max 1KB per string
   size_t max_array_size = 10000;      // Max 10,000 elements per array
   size_t max_map_size = 1000;         // Max 1,000 entries per map
};

auto ec = glz::read<secure_opts{}>(obj, buffer);
```

This follows the existing Glaze pattern for optional options - only users who need these features pay for them.

### Separate Limits for Arrays vs Maps

Arrays and maps have separate limits for maximum flexibility:

```cpp
struct data_processing_opts : glz::opts
{
   uint32_t format = glz::BEVE;
   size_t max_array_size = 1000000;  // Allow large numeric data arrays
   size_t max_map_size = 100;        // But limit configuration object complexity
};
```

### Per-Field Wrapper

New `glz::max_length` wrapper for fine-grained control on individual fields:

```cpp
template <>
struct glz::meta<UserInput>
{
   using T = UserInput;
      "username", glz::max_length<&T::username, 64>,  // Max 64 chars
      "scores", glz::max_length<&T::scores, 100>      // Max 100 elements
   );
};
```

Works with strings and arrays. Supported for both BEVE and CBOR formats.

### Supported Container Types

**`max_array_size` applies to:**
- `std::vector<T>`, `std::array<T, N>`, `std::deque<T>`
- Other sequence containers

**`max_map_size` applies to:**
- `std::map<K, V>`, `std::unordered_map<K, V>`
- Other associative containers

**`max_string_length` applies to:**
- `std::string` and string-like types

### Nested Limits

Limits propagate through nested structures:
```cpp
struct nested_opts : glz::opts
{
   uint32_t format = glz::BEVE;
   size_t max_array_size = 100;
   size_t max_string_length = 1024;
};

std::vector<Example> example; // Example contains std::string name
auto ec = glz::read<nested_opts{}>(example, buffer);
// Limits vector to 100 elements AND limits each name string to 1024 chars
```

### Memory Bomb Protection

Added buffer bounds checking before allocation in both BEVE and CBOR to prevent memory exhaustion from malicious length headers.

### Error Handling

When limits are exceeded, parsing fails with `glz::error_code::invalid_length`.

**Documentation:**
- `docs/wrappers.md` - Documented the `max_length` wrapper
- `docs/security.md` - Updated security documentation with per-field limits

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -1298,6 +1298,21 @@ namespace glz
          if constexpr (Opts.partial_read) {
             n = value.size();
          }
+         else {
+            // Validate count against remaining buffer size (minimum 1 byte per key-value pair)
+            if (n > size_t(end - it)) [[unlikely]] {
+               ctx.error = error_code::unexpected_end;
+               return;
+            }
+
+            // Check user-configured map size limit
+            if constexpr (check_max_map_size(Opts) > 0) {
+               if (n > check_max_map_size(Opts)) [[unlikely]] {
+                  ctx.error = error_code::invalid_length;
+                  return;
+               }
+            }
+         }
 
          constexpr uint8_t key_tag = beve_key_traits<Key>::key_tag;
 
--- a/include/glaze/cbor/read.hpp
+++ b/include/glaze/cbor/read.hpp
@@ -1194,6 +1194,20 @@ namespace glz
             if (bool(ctx.error)) [[unlikely]]
                return;
 
+            // Validate count against remaining buffer size (minimum 2 bytes per key-value pair)
+            if (count * 2 > static_cast<uint64_t>(end - it)) [[unlikely]] {
+               ctx.error = error_code::unexpected_end;
+               return;
+            }
+
+            // Check user-configured map size limit
+            if constexpr (check_max_map_size(Opts) > 0) {
+               if (count > check_max_map_size(Opts)) [[unlikely]] {
+                  ctx.error = error_code::invalid_length;
+                  return;
+               }
+            }
+
             for (size_t i = 0; i < count; ++i) {
                Key key{};
                parse<CBOR>::op<Opts>(key, ctx, it, end);
--- a/include/glaze/core/opts.hpp
+++ b/include/glaze/core/opts.hpp
@@ -471,6 +471,16 @@ namespace glz
       }
    }
 
+   consteval size_t check_max_map_size(auto&& Opts)
+   {
+      if constexpr (requires { Opts.max_map_size; }) {
+         return Opts.max_map_size;
+      }
+      else {
+         return 0; // 0 means no limit
+      }
+   }
+
    consteval bool check_opening_handled(auto&& o) { return o.internal & uint32_t(opts_internal::opening_handled); }
 
    consteval bool check_closing_handled(auto&& o) { return o.internal & uint32_t(opts_internal::closing_handled); }
--- a/tests/beve_test/beve_test.cpp
+++ b/tests/beve_test/beve_test.cpp
@@ -4363,6 +4363,117 @@ suite allocation_limits = [] {
       expect(!ec) << "Default opts should allow any array size";
       expect(arr_result == large_array);
    };
+
+   "max_map_size applies to std::map"_test = [] {
+      std::map<std::string, int> large_map;
+      for (int i = 0; i < 100; ++i) {
+         large_map["key" + std::to_string(i)] = i;
+      }
+      std::string buffer;
+      expect(not glz::write_beve(large_map, buffer));
+
+      // Try to read with a limit of 50 entries
+      struct map_limited_opts : glz::opts
+      {
+         uint32_t format = glz::BEVE;
+         size_t max_map_size = 50;
+      };
+
+      std::map<std::string, int> result;
+      auto ec = glz::read<map_limited_opts{}>(result, buffer);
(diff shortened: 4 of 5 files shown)


stephenberry:
@ElevationsRPG, any additional thoughts on this finer control? Anything missing?

ElevationsRPG:
> @ElevationsRPG, any additional thoughts on this finer control? Anything missing?

So far everything looks great! but i do have a few questions/suggestions!

Does it also support std::unordered_map and std::map containers ?

and could we support max_length for unwrapped structs? Currently, I use static constexpr auto value = &T::value; to treat a struct as its underlying type. It would be great if I could do this aswell (But not absolutely necessary).

```
template <>
struct glz::meta<Key> {
   using T = Key;
   static constexpr auto value = glz::max_length<&T::value, 64>; 
};
```

and if the max_length works on integer variables as well that would also be great where the limit is also the max value limit of the integer.

But so far everything looks great! here is how i would like to use it:

```
struct Example
{
std::string name;
int id;
struct glaze
{
using T = Example;
      "name", glz::max_length<&T::name, 64>,  // Max 64 chars
      "id", glz::max_length<&T::id, 100>      // Max value is 100
   );
}
}

and:

std::string buffer;
std::vector<int> example;
auto ec = glz::read_beve<glz::opts{ max_array_size = 100 }>(example, buffer); //this should throw error if it exceeds 100
if (ec)
{
//etc
}

and:

std::string buffer;
std::vector<std::string> example;
auto ec = glz::read_beve<glz::opts{ max_array_size = 100, max_string_length = 1024 }>(example, buffer); //this should throw error if it exceeds 100 elements and an error because a string was above 1024
if (ec)
{
//etc
}

and:

std::string buffer;
std::vector<Example> example; //using our custom struct that contains a std::string and id
auto ec = glz::read_beve<glz::opts{ max_array_size = 100, max_string_length = 1024 }>(example, buffer); //this should throw error if it exceeds 100 elements and an error because a string was above 1024
if (ec)
{
//etc
}
```

Other then that the only other option i would love to see is the .allocate_null_pointers! and a total_allocation_limit could be useful aswell.

ElevationsRPG:
Oh yes i also wanted to confirm that max_string_length (glz::option) will also apply to std::strings that are inside a std::vector or inside a struct, and or inside a map, list, deque, etc globally ?

stephenberry:
@ElevationsRPG, I added a `max_map_size` option and the `max_length` wrapper sets it for maps, so it also works for maps now. This will apply to nested strings, maps, vectors, so you can nest things as much as you like.

I'm going to merge this in.

I haven't added `max_length` support for integers. Integer limits can be set via constraints already, so I'm not sure about duplication this feature, and `max_length` isn't really the appropriate word for a limit on the integer value.

Here's an example of an integer constraint with a custom error message:
```c++
struct age_custom_error_obj
{
   int age{};
};

template <>
struct glz::meta<age_custom_error_obj>
{
   using T = age_custom_error_obj;
   static constexpr auto read_x = [](T& s, int age, glz::context& ctx) {
      if (age < 21) {
         ctx.error = glz::error_code::constraint_violated;
         ctx.custom_error_message = "age too young";
      }
      else {
         s.age = age;
      }
   };
   static constexpr auto value = object("age", glz::custom<read_x, &T::age>);
};
```

Note that constraints read into temporaries, so they don't apply the change to the underlying object until the constraint is satisfied.

ElevationsRPG:
> @ElevationsRPG, I added a `max_map_size` option and the `max_length` wrapper sets it for maps, so it also works for maps now. This will apply to nested strings, maps, vectors, so you can nest things as much as you like.
> 
> I'm going to merge this in.
> 
> I haven't added `max_length` support for integers. Integer limits can be set via constraints already, so I'm not sure about duplication this feature, and `max_length` isn't really the appropriate word for a limit on the integer value.
> 
> Here's an example of an integer constraint with a custom error message:
> 
> ```c++
> struct age_custom_error_obj
> {
>    int age{};
> };
> 
> template <>
> struct glz::meta<age_custom_error_obj>
> {
>    using T = age_custom_error_obj;
>    static constexpr auto read_x = [](T& s, int age, glz::context& ctx) {
>       if (age < 21) {
>          ctx.error = glz::error_code::constraint_violated;
>          ctx.custom_error_message = "age too young";
>       }
>       else {
>          s.age = age;
>       }
>    };
>    static constexpr auto value = object("age", glz::custom<read_x, &T::age>);
> };
> ```
> 
> Note that constraints read into temporaries, so they don't apply the change to the underlying object until the constraint is satisfied.

Yes i understand the reasoning not using max_length for integers, plus most of the fixed types are pretty safe against memory exhaustion e.g int32_t is 4 bytes int64_t 8 bytes and so on ah yes i saw the "propagate" through nested things that is excellent!

the only other option i would like is the `.allocate_raw_pointers` and making sure it will work on all container types aswell e.g

`std::vector<Example*>`
`std::unordered_map<int, Example*>`

PR #2199, opened 2026-01-03
Runtime max length limits

### LINKED ISSUE #2198: Runtime BEVE constraints
> The `allocate_raw_pointers` option has been added. Thanks for making this request!
> 
> Hey stephen, so i just tested the latest build 6.5.1 and everything is working great BUT! theres a big problem.. These are compile-time constants which means i cannot change my sizes at run-time which is very limiting in how i wish to use these e.g through a template at run-time
> 
> ```
> struct secure_opts : glz::opts
> {
>     uint32_t format = glz::BEVE;
>     bool allocate_raw_pointers = true;
>     size_t max_array_size = 0;
> };
> 
> unsigned int test = 100;
> 
> secure_opts test_opts;
> test_opts.max_array_size = test; //cannot dynamically set size at run-time
> glz::read_beve_at <test_opts> (example2, buffer, 0);
> ```
> 
> Would it also be possible to allow for a run-time switch ? that allows us to change the options at run-time or a support options something like:
> 
> `glz::opts::run_time_option = true;`
> 
> I have many different "limits" that i would like to be able to set at run-time. 

 _Originally posted by @ElevationsRPG in [#2192](https://github.com/stephenberry/glaze/issues/2192#issuecomment-3707319804)_

stephenberry:
# Runtime BEVE/CBOR Constraints

Adds support for runtime constraints when reading BEVE and CBOR formats. This complements the existing compile-time constraints (used with `glz::meta` and `max_length_t`).

### `include/glaze/core/context.hpp`

Added three concepts to detect runtime constraint fields in user-defined contexts:

- `has_runtime_max_string_length`
- `has_runtime_max_array_size`
- `has_runtime_max_map_size`

### `include/glaze/beve/read.hpp`

Added runtime constraint checks at 9 locations (strings, arrays, maps) using `if constexpr` with the new concepts.

### `include/glaze/cbor/read.hpp`

Added runtime constraint checks at 7 locations (strings, arrays, maps) using `if constexpr` with the new concepts.

## Usage

Users inherit from `glz::context` and add constraint fields:

```cpp
struct my_context : glz::context {
   size_t max_string_length = 0;  // 0 = no limit
   size_t max_array_size = 0;
   size_t max_map_size = 0;
};

my_context ctx;
ctx.max_array_size = user_provided_limit;  // Set dynamically at runtime

glz::read<glz::opts{.format = glz::BEVE}>(output, buffer, ctx);
```

## Design

- **Zero binary overhead**: Uses `if constexpr` with concepts, so code is compiled out when using the base `glz::context`
- **Additive**: Runtime constraints work alongside compile-time constraints; either can reject
- **Consistent**: Same field names and semantics as compile-time options (0 = no limit)

Closes #2198

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -584,6 +584,12 @@ namespace glz
                return;
             }
          }
+         if constexpr (has_runtime_max_string_length<std::decay_t<decltype(ctx)>>) {
+            if (ctx.max_string_length > 0 && n > ctx.max_string_length) [[unlikely]] {
+               ctx.error = error_code::invalid_length;
+               return;
+            }
+         }
          value.resize(n);
          std::memcpy(value.data(), it, n);
          it += n;
@@ -615,6 +621,12 @@ namespace glz
          }
          if constexpr (check_max_string_length(Opts) > 0) {
             if (n > check_max_string_length(Opts)) [[unlikely]] {
+               ctx.error = error_code::invalid_length;
+               return;
+            }
+         }
+         if constexpr (has_runtime_max_string_length<std::decay_t<decltype(ctx)>>) {
+            if (ctx.max_string_length > 0 && n > ctx.max_string_length) [[unlikely]] {
                ctx.error = error_code::invalid_length;
                return;
             }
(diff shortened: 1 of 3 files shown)


ElevationsRPG:
Looks very good stephen! One more thing though It will be nice to include the allocate_raw_pointers at runtime as-well as other inheritable compile-time options so it reduces me having to write constexpr for every pointer deserialize that will happen, but also the other options from: https://stephenberry.github.io/glaze/options/ that are already specified in compile-time would be nice to have in addition to this so we can also specify them in the context (ctx) aswell when necessary.

Oh and one other thing does it also work for this?

`glz::read_beve_at<glz::opts{.format = glz::BEVE}>(output, buffer, offset, ctx);`

Issue #1048, opened 2024-06-04
calculate serialize size before serialization and serialize to pre-allocated memory


alan0526:
I am working on inter process communication base on shared memory and Unix Socket Domain recently, I plan to use glaze Beve to serialize my data. 
I want calculate the object serialization size then alloc shared memory and write serialization data into it. 
Is there some existing solution to do this. thanks.

stephenberry:
Serializing to a pre-allocated buffer versus a dynamic buffer like `std::string` typically has less than a 5% difference in performance. Furthermore, in inter-process communication you should typically reuse your serialization buffer. Glaze uses a doubling of the buffer capacity if growth is needed, so it is rare that you will have new allocations if you reuse your serialization buffer.

Is there a reason why you can't reuse a `std::string` buffer? 

A function to calculate the serialization size could be added, but it would need to iterate through your structures. This would actually add more overhead than simply reusing the `std::string` buffer. So you might have a tiny performance boost for the very first message you send, but subsequent messages would run slower needing to compute the size beforehand.

alan0526:
Before if I use string I need to serialize to std::string then write it to my shared memory, there is a extra copy.

stephenberry:
I see. I'll keep this issue alive to add the ability to compute the space required before serialization.

I typically use shared libraries which allow me to share the std::string directly across processes, which allows dynamic buffer sharing and avoids the extra copy. But, I can see how if you are wanting to connect services in other programming languages it is easier to not have to deal with dynamic memory and just compute the size beforehand.

alan0526:
thanks

stephenberry:
Support has been added with `glz::beve_size` in #2206

PR #2206, opened 2026-01-04
glz::beve_size precompute

stephenberry:
# Add `beve_size` function for pre-calculating BEVE serialization size

Addresses #1048

## Summary

This PR adds a new `glz::beve_size(value)` function that calculates the exact number of bytes needed to serialize a value to BEVE format, without actually performing the serialization.

## Use Case

This enables efficient serialization to pre-allocated buffers such as shared memory for IPC:

```cpp
MyStruct obj{...};
size_t size = glz::beve_size(obj);
void* shm = allocate_shared_memory(size);
glz::write_beve(obj, std::span{static_cast<char*>(shm), size});
```

## New API

- `glz::beve_size(value)` - Returns the serialized size in bytes
- `glz::beve_size_untagged(value)` - Size for untagged serialization (`structs_as_arrays = true`)
- `glz::compressed_int_size(n)` - Helper to calculate compressed integer encoding size

## Files Changed

- **New:** `include/glaze/beve/size.hpp` - Size calculation implementation
- **Modified:** `include/glaze/beve.hpp` - Added include for size.hpp
- **Modified:** `tests/beve_test/beve_test.cpp` - Added comprehensive tests

## Supported Types

All types supported by BEVE serialization:
- Primitives (bool, integers, floats, char, enums)
- Strings (std::string, string_view, const char*)
- Containers (vector, array, map, set, list, deque, span)
- Nullable types (optional, shared_ptr, unique_ptr)
- Variants, tuples, pairs
- Complex numbers, bitsets
- Nested structures of arbitrary depth
- Glaze-specific types (glaze_value_t, glaze_array_t, glz::obj, raw_json)

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -29,6 +29,80 @@ if (!ec) {
 >
 > Reading binary is safe for invalid input and does not require null terminated buffers.
 
+## Calculate BEVE Size
+
+The `glz::beve_size` function calculates the exact number of bytes needed to serialize a value to BEVE format, without actually performing the serialization. This is useful when you need to pre-allocate a buffer of the exact size, such as when writing to shared memory for inter-process communication.
+
+**Basic Usage**
+
+```c++
+my_struct s{};
+size_t size = glz::beve_size(s);
+// size contains the exact number of bytes needed to serialize s
+```
+
+**Shared Memory Example**
+
+```c++
+#include "glaze/beve.hpp"
+
+struct SensorData {
+   uint64_t timestamp;
+   double temperature;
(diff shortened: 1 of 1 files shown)


ElevationsRPG:
Hi Stephen, glz::beve_size is great for pre-allocating. It would also be very useful to have a glz::beve_peek_size (or count) that works on the incoming buffer. This would allow us to grab the 'claimed' size/elements from the BEVE header before we actually start the deserialization and allocation process, although im aware that our run-time limits already solve that problem. But it would be nice to have the counter-part to glz::beve_size

PR #2211, opened 2026-01-05
glz::lazy_json

stephenberry:
# Add `glz::lazy_json` - Truly Lazy JSON Parsing

This PR introduces `glz::lazy_json`, a lazy JSON parser that performs **zero upfront processing**. Unlike traditional parsers that build a DOM or index the entire document, `lazy_json` only parses the bytes you actually access.

### Lazy Navigation
```cpp
auto result = glz::lazy_json(json);
auto email = (*result)["user"]["profile"]["email"].get<std::string_view>();
```
- `lazy_json()` is O(1) - just stores a pointer
- Field access scans only the bytes needed to find that field
- Progressive scanning: accessing keys in document order gives O(n) total complexity

### Indexed Views for O(1) Access
```cpp
auto users = (*result)["users"].index();  // Build index once - O(n)
auto user500 = users[500];                 // O(1) random access
size_t count = users.size();               // O(1) size query
```
For scenarios requiring multiple random accesses, build an index for O(1) element access.

### Direct Struct Deserialization
```cpp
User user{};
glz::read_json(user, (*result)["user"]);  // Single-pass, ~49% faster
```
New `read_json` overload accepts `lazy_json_view` directly, avoiding the double-scan that `raw_json()` + `read_json()` requires.

### Full Iterator Support
```cpp
for (auto& item : doc["items"]) {
    auto id = item["id"].get<int64_t>();
}
```

### `lazy_json_view` Methods
- `operator[](key)` / `operator[](index)` - Navigate to field/element
- `get<T>()` - Extract typed value
- `read_into(T&)` - Deserialize into struct (single-pass)
- `raw_json()` - Get raw JSON bytes as `string_view`
- `index()` - Build indexed view for O(1) access
- `is_object()`, `is_array()`, `is_string()`, etc. - Type checking
- `size()`, `empty()`, `contains(key)` - Container methods
- Iterator support for range-based for loops

### `indexed_lazy_view` Methods
- `operator[](index)` - O(1) random access
- `size()` - O(1) element count
- Full random-access iterator support

### Free Functions
- `glz::lazy_json(buffer)` - Create lazy document
- `glz::read_json(T&, lazy_json_view)` - Deserialize view to type
- `glz::write_json(lazy_json_view, output)` - Write view to JSON

### Added
- `include/glaze/json/lazy.hpp` - Lazy JSON implementation (~1400 lines)
- `tests/json_test/lazy_test.cpp` - Comprehensive tests (59 tests)
- `docs/lazy-json.md` - Full documentation

### Modified
- `include/glaze/util/parse.hpp` - Added `lazy_char_type` enum and lookup tables
- `include/glaze/json.hpp` - Include lazy.hpp
- `tests/json_test/CMakeLists.txt` - Add lazy_test target
- `mkdocs.yml` - Add lazy-json to docs navigation

## When to Use

| Use Case | Recommended |
|----------|-------------|
| Extract 1-3 fields from large JSON | `glz::lazy_json` |
| Multiple random array accesses | `lazy_json` with `.index()` |
| Full struct deserialization | `glz::read_json` |
| Unknown/dynamic structure with persistence | `glz::generic` |

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,4 @@ alabastar.beve
 
 # Benchmark comparison directory
 benchmarks/
+pr_description.md
--- a/README.md
+++ b/README.md
@@ -239,7 +239,23 @@ For faster compilation at the cost of peak performance, use `glaze_DISABLE_ALWAY
 set(glaze_DISABLE_ALWAYS_INLINE ON)
 ```
 
-> **Note:** This primarily reduces compilation time, not binary size. For binary size reduction, use the `linear_search` option. See [Optimizing Performance](https://stephenberry.github.io/glaze/optimizing-performance/) for more details.
+> **Note:** This primarily reduces compilation time, not binary size. For binary size reduction, see Optimization Levels below.
+
+### Optimization Levels (Embedded/Size Optimization)
+
+Glaze provides optimization levels to control the trade-off between binary size and runtime performance. This is useful for embedded systems:
+
+```cpp
+auto json = glz::write<glz::opts_size{}>(obj);
+auto ec = glz::read<glz::opts_size{}>(obj, buffer);
+```
+
+| Level | Preset | Description |
+|-------|--------|-------------|
+| `normal` | (default) | Maximum performance with 40KB+ lookup tables |
+| `size` | `opts_size` | Minimal binary using `std::to_chars`, linear search |
+
+See [Optimization Levels](https://stephenberry.github.io/glaze/optimization-levels/) for full details.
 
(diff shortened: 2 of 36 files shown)


ElevationsRPG:
@stephenberry So this actually lets you access fields directly in the serialized data? like for example:
```

auto result = glz::lazy_json(json);

if (result["username"] == "mytestusername") // ??
```

If so this would be incredibly useful for beve aswell like a glz::lazy_beve since im sure when beve is using tags its the roughly the same as json metadata no?

Issue #2224, opened 2026-01-07
(6.5.1) Multi object variant reading results in compile error on msvc v145


Clyraz:
Seems to be a v145 only issue, 6.5.0 works fine though.

<img width="764" height="769" alt="Image" src="https://github.com/user-attachments/assets/7c4935f6-18b5-4e74-9072-c7ab9d455155" />

stephenberry:
Ah, I recently removed the `tuple_cat` approach, but didn't test the change on MSVC. I currently can run MSVC on CI because I'm waiting for them to release some compiler fixes, although most of Glaze runs on MSVC. Apparently this parameter pack approach reveals another compiler bug. I'll probably just go back to the working `tuple_cat` approach, even if it is slower on other compilers.

stephenberry:
Thanks for reporting this bug! The fix was merged and will be in the next release of Glaze.

PR #2228, opened 2026-01-07
Using helper struct instead of IIFE for MSVC compatibility

stephenberry:
Issue: GitHub issue #2224 - Multi object variant reading results in compile error on MSVC v145 (VS2022)

Root Cause: The commit 4a6393cb ("remove tuple_cat use") introduced immediately-invoked lambda expressions (IIFEs) with template parameter packs in variable template definitions. While this pattern compiles on Clang and GCC, MSVC v145 has known issues with:
1. Template parameter packs in lambda template-heads
2. IIFEs used in variable template definitions
3. Multiple template parameter packs in a single lambda

Fix: Replaced the three IIFE patterns in include/glaze/json/read.hpp with explicit helper struct specializations:

1. variant_count_v → variant_count_impl<Variant, Trait> helper struct
2. variant_first_index_v → variant_first_index_impl<Variant, Trait> helper struct
3. variant_filtered_count_v → variant_filtered_count_impl<Variant, Trait, IsConst> helper struct

The public interface remains identical - only the implementation changed from:
```c++
template <class Variant, template <class> class Trait>
constexpr size_t variant_count_v = []<class... Ts>(std::variant<Ts...>*) {
   return (size_t(Trait<Ts>::value) + ... + 0);
}(static_cast<Variant*>(nullptr));
```

To:
```c++
template <template <class> class Trait, class... Ts>
struct variant_count_impl<std::variant<Ts...>, Trait> {
   static constexpr size_t value = (size_t(Trait<Ts>::value) + ... + 0);
};
```

template <class Variant, template <class> class Trait>
constexpr size_t variant_count_v = variant_count_impl<Variant, Trait>::value;

This helper struct pattern is more portable across compilers while maintaining the same compile-time performance benefits (fold expressions, no tuple_cat).

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -3378,26 +3378,54 @@ namespace glz
    {};
 
    // Count types in variant matching a trait (fold expression, very fast to compile)
+   // Using helper struct instead of IIFE for MSVC compatibility
    template <class Variant, template <class> class Trait>
-   constexpr size_t variant_count_v = []<class... Ts>(std::variant<Ts...>*) {
-      return (size_t(Trait<Ts>::value) + ... + 0);
-   }(static_cast<Variant*>(nullptr));
+   struct variant_count_impl;
+
+   template <template <class> class Trait, class... Ts>
+   struct variant_count_impl<std::variant<Ts...>, Trait>
+   {
+      static constexpr size_t value = (size_t(Trait<Ts>::value) + ... + 0);
+   };
+
+   template <class Variant, template <class> class Trait>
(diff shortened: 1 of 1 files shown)


stephenberry:
@Clyraz, could you test this branch and see if it fixes the multi object variant reading? If you can't get to it I can test myself.

Clyraz:
> @Clyraz, could you test this branch and see if it fixes the multi object variant reading? If you can't get to it I can test myself.

This branch fixes the issue 👍. (Also v145 is VS2026 not VS2022)

PR #2233, opened 2026-01-08
Query Parameter and URL Encoding Support

stephenberry:
# Query Parameter and URL Encoding Support

Adds query parameter support to the HTTP router and URL encoding utilities for parsing form data.

### New Header: `glaze/net/url.hpp`

Standalone URL utilities:

- `url_decode()` - Decode percent-encoded strings (`%20` → space, `+` → space)
- `url_encode()` - Encode strings for safe use in URLs (space → `+`, special chars → `%XX`)
- `parse_urlencoded()` - Parse `key=value&key2=value2` format into a map
- `split_target()` - Split URL into path and query string components

All functions support buffer-passing overloads for zero-allocation parsing in hot paths.

### Updated: `glaze/net/http_router.hpp`

- Added `path` and `query` fields to `request` struct
- `match()` now strips query strings before route matching
- Path parameters (`:id`) and wildcard captures (`*path`) are now URL-decoded

### Updated: `glaze/net/http_server.hpp`

- Automatically parses query parameters into `request.query`
- Populates `request.path` with the path component (without query string)

## Usage

```cpp
server.get("/api/users", [](const glz::request& req, glz::response& res) {
    // GET /api/users?limit=10&offset=20

    auto limit = req.query.find("limit");
    if (limit != req.query.end()) {
        int n = std::stoi(limit->second);
    }

    // req.path == "/api/users"
    // req.target == "/api/users?limit=10&offset=20"
});

// Parse form POST body
auto form = glz::parse_urlencoded(req.body);

// Build URL with encoded query parameters
std::string url = "/api/search?q=" + glz::url_encode(user_input);
```

## Documentation

- New: `docs/networking/url.md`
- Updated: `docs/networking/http-router.md` (query parameters section)
- Updated: `docs/networking/http-server.md` (query parameters section)

## Tests

- New: `tests/networking_tests/url_test/` - URL utility tests
- Updated: `tests/networking_tests/http_router_test/` - Query string routing tests

### DIFF
--- a/docs/networking/url.md
+++ b/docs/networking/url.md
@@ -51,6 +51,59 @@ glz::url_decode("foo%2Fbar", buffer);
 std::cout << buffer << std::endl;  // "foo/bar"
 ```
 
+## URL Encoding
+
+### Basic Usage
+
+```cpp
+// Encode strings for safe use in URLs
+std::string encoded = glz::url_encode("hello world");     // "hello+world"
+std::string path = glz::url_encode("path/to/file");       // "path%2Fto%2Ffile"
+std::string special = glz::url_encode("a=b&c=d");         // "a%3Db%26c%3Dd"
+```
+
+### Encoding Rules
+
+- **Unreserved characters** pass through unchanged: `A-Z`, `a-z`, `0-9`, `-`, `.`, `_`, `~`
+- **Space** is encoded as `+` (per `application/x-www-form-urlencoded`)
+- **All other characters** are percent-encoded (`%XX`)
+
+### Building Query Strings
+
+```cpp
+// Build a safe query string
+std::string query = "q=" + glz::url_encode(user_search) +
+                    "&category=" + glz::url_encode(category);
+
+// Example: user_search = "C++ templates", category = "programming/advanced"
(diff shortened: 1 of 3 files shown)


ElevationsRPG:
@stephenberry Hey stephen! im not sure this resolves my issue of parsing raw arrays only for example:

in json i wanted to output things like this using write_json:

`[1,2,3,4,5,"examplestring",6,7,"anotherstring"]`

but to get this format it has to be a very specific format such as 

`std::vector<std::variant<int, std::string>> //this will produce a raw array`

but i wanted to be able to get this format even if i use:

`std::unordered_map<std::pair<int, int>, std::string> //this will not produce the raw array`

if that makes sense, im not sure how this url_encode and url_decode will fix that feature although it is a nice thing to have.
basically my idea of "parse raw arrays" was it to be a compile-time option where no matter what container type you specify it will parse it as a raw array in this format:

`[1,2,3,4,5,"examplestring",6,7,"anotherstring"]`

stephenberry:
@ElevationsRPG, sorry about this mistake, I accidentally tagged your issue when I meant to tag another. Your issue is still open.

PR #2235, opened 2026-01-10
Fix http post body addional read

clagix:
Unfortunately, I discovered a bug where post() handlers are not invoked for HTTP requests that have a non-empty body and a large number of additional HTTP header lines. In this situation, the http_server attempts to read the remaining body content using asio::async_read().

In my setup (using standalone Asio), the associated completion handler is never called. I suspect this is caused by an issue with reusing the asio::streambuf object.

To address this, I changed the code that reads the remaining body content to use zero-copy reads directly into the body string. This also allows us to eliminate several additional allocations and copy operations required when using asio::streambuf. Since the required read size is known in advance from the Content-Length header, dynamic allocation is unnecessary.


stephenberry:
Thanks so much for this fix!

Note: Passing in a char* to the `replace` function for the cstr causes a `length` call underneath, which essentially does `strlen`. This is an issue if any null characters exist in the string. Reference: https://en.cppreference.com/w/cpp/string/basic_string/replace.html

I've replaced this with a `std::memcpy`, which is also more explicit about what is happening.

Added some more unit tests as well.

clagix:
Thanks for taking the time to review my PR. I’m happy to see the bug fixed upstream, and I’ll keep using Glaze’s http_server in my project.

While std::copy() would be the more C++-ish choice, memcpy still seems to perform better, so I’m fine sticking with it.

I’ll keep my fork for now since I still have a few ideas to improve http_server.

Issue #2250, opened 2026-01-20
Add support for std::expected<void, E>


oleid:
It would seem that one cannot currently serialize/deserialize `std::expected<void, E>` using glaze.
This type is often used as return value for fail-able functions, which don't have any return value.

When using e.g. JSON serialization, the compilation fails, since `*value` (which obviously doesn't exist for `void`) is an invalid expression.

Tested with glaze version 7.0.1.

oleid:
@stephenberry: Do you know if other formats already support std::expected<void, E>? It is quite possible that they don't.

PR #2251, opened 2026-01-20
FEATURE: [json] Add special handling for std::expected<void,E>

oleid:
This appears to work for JSON serialization. I don't really like the else case in write.hpp, though. Maybe you have a better idea?

See also #2250


stephenberry:
This looks good! I just made some minor tweaks, but I'm merging it. Thanks for the contribution.

PR #2257, opened 2026-01-21
Use std::source_location for get_name

AmmoniumX:
Simple change to use the C++20 `std::source_location` class to get the name of struct fields instead of the `GLZ_PRETTY_FUNCTION`. Uses the feature-test macro `__cpp_lib_source_location` to check if source_location is available and fully implemented with the existing `GLZ_PRETTY_FUNCTION` as a fallback, although since the project targets C++23 as a minimum this might not be necessary, and removing the macro test entirely is an option.

Tests result:
```

99% tests passed, 1 tests failed out of 80

Label Time Summary:
format_BEVE    =   0.01 sec*proc (1 test)
format_JSON    =   0.01 sec*proc (1 test)

Total Test time (real) =   9.84 sec

The following tests FAILED:
	70 - openapi_test (Failed)
```

Full version at: https://pastebin.com/rLF7r8Zt

Note: test `70 - openapi_test` is also failing in main branch as of the time of this PR, so the fail was not introduced due to this.

### DIFF
--- a/.github/workflows/armv7.yml
+++ b/.github/workflows/armv7.yml
@@ -1,0 +1,87 @@
+name: armv7
+
+on:
+  push:
+    branches:
+      - main
+      - feature/*
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - '**/*.md'
+      - 'docs/**'
+  workflow_dispatch:
+
+jobs:
+  build-armv7:
+    runs-on: ubuntu-24.04
+    name: GCC armv7 (NEON, cross-compiled)
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Install cross-compiler and QEMU
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y \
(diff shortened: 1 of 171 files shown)


stephenberry:
I'm been planning to move to `std::source_location`, but I need to benchmark compilation time. And, we're about to get C++26 support (PR is open here: #2217), which will bypass this code altogether.

stephenberry:
@AmmoniumX, thanks for putting this together. While std::source_location is a cleaner mechanism, I’m hesitant to merge it here.
 
If Glaze still needs to keep the GLZ_PRETTY_FUNCTION path for compiler/version edge cases, then this adds a second implementation path that we need to maintain and test across compilers. The existing GLZ_PRETTY_FUNCTION approach has also been exercised by users for a long time, so I’d rather avoid introducing new variability in this area unless there is a clear long-term benefit.

Also, Glaze now has C++26 reflection support, so this would be an interim solution rather than the direction we expect to keep investing in. If this were the long-term approach, I’d be more open to the new approach, but for a temporary path I don’t think the tradeoff is worth it.

But, thanks so much for making this PR!

PR #2258, opened 2026-01-21
Allow to construct a glz::http_server instance using an asio::io_executor

clagix:
**Allow constructing a `glz::http_server` using an `asio::io_executor` instead of requiring direct access to an `asio::io_context`.**

Motivation
In some scenarios, an application uses a shared `io_context` event loop (for example, in a single-threaded architecture), but does not own or have direct access to the underlying io_context instance, nor is it managed via a shared_ptr.

In these cases, the only available handle is an `asio::io_executor`, which Asio provides as an abstraction for posting work to the `io_context`’s event loop without exposing the `io_context` itself.

Requiring a concrete `io_context` makes it difficult or impossible to integrate `glz::http_server` into such setups. By allowing construction from an `asio::io_executor`, the server can be safely and flexibly integrated into applications that intentionally restrict ownership or visibility of the `io_context`.

Benefits
- Improves compatibility with shared or externally-managed event loops
- Aligns with Asio’s executor-based design philosophy
- Enables cleaner integration in single-threaded and modular architectures

### DIFF
--- a/include/glaze/net/http_server.hpp
+++ b/include/glaze/net/http_server.hpp
@@ -1388,6 +1388,9 @@ namespace glz
          threads.clear();
       }
 
+      // Get executor for async operations (timers, etc.)
+      asio::any_io_executor get_executor() const { return io_executor; }
+
      private:
       asio::any_io_executor io_executor;
       std::shared_ptr<asio::io_context> io_context;


stephenberry:
This looks great! I made one change to store the `context` if it is provided. I think this is safer, but please let me know if I misunderstood:

```c++
inline http_server(std::shared_ptr<asio::io_context> context,
                         glz::error_handler custom_error_handler = {})
         : http_server(context ? context->get_executor() : asio::any_io_executor(),
                       custom_error_handler)
      {
         // Preserve shared ownership to prevent use-after-free. The delegated constructor
         // only stores io_context when no executor is provided, but the executor alone
         // doesn't keep the io_context alive.
         if (context) {
            io_context = context;
         }
      }
```

clagix:
Hi,

Thank you for considering the PR. You can implement it that way. It more closely matches the previous behavior. If you prefer to use a shared `io_context` and manage its lifecycle yourself, you can now use the new constructor and supply the `io_executor`.

Naturally, the lifetime of the `io_context` must exceed that of the`http_server` instance, but this is consistent with the requirements for other Asio objects such as sockets or timers.

stephenberry:
Cool, I think it's ready to be merged, do you agree?

clagix:
> Cool, I think it's ready to be merged, do you agree?

If you’re okay with it, I’d like to apply the same change to the `http_client` class by passing an io_executor to its constructor. But we can also do this in another PR?

stephenberry:
> If you’re okay with it, I’d like to apply the same change to the http_client class by passing an io_executor to its constructor. But we can also do this in another PR?

You can make the change in this PR if you like. Whatever is easier.

clagix:
I’ve added the changes to the http_client in the PR. Please take a look and let me know your thoughts.

stephenberry:
Looks good to me, I'll merge.

PR #2268, opened 2026-01-26
SIMD optimization for whitespace skipping

wilkolbrzym-coder:
I’ve been playing around with the whitespace skipping in Glaze and managed to get a massive boost for prettified JSONs using SIMD. On my machine, throughput jumped from about 410 MB/s to over 630 MB/s when dealing with formatted data.
​I have to be honest — there’s a tiny hit on minified JSONs (about 3% drop, from 453 MB/s to 440 MB/s). It’s likely just the extra instruction overhead and a bit of branch pressure when there’s nothing to skip.
​In my opinion, the trade-off is totally worth it. Most real-world formatted logs or configs get a 1.5x speedup, while the "loss" on minified files is basically noise.
​All tests passed, and the logic handles the tail end safely. Let me know what you think about the trade-off!

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -185,6 +185,12 @@ namespace glz
       return false;
    }
 
+   template <auto Opts>
+   GLZ_ALWAYS_INLINE bool parse_ws_colon(is_context auto& ctx, auto&& it, auto end) noexcept
+   {
+      return parse_ws_char<':', Opts>(ctx, it, end);
+   }
+
    // Extracted from decode_index to reduce code duplication across instantiations.
    // This path handles unknown keys and doesn't depend on the field index I.
    template <auto Opts, class T, class Value>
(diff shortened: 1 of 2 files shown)


stephenberry:
@wilkolbrzym-coder, I know lots of benchmarks like to include whitespace in datasets. It makes benchmarks look faster. But, any use of JSON that is serious about performance won't have any whitespace, because it is a waste of bytes and conveys no additional meaning. This is true when working with JSON datasets or server-client use cases where size is critical.

So, any change that slows down the minified case is going in the wrong direction. It's great to speed up prettified JSON parsing, but anyone concerned with performance would just make their code use minified JSON without whitespace.

So, for Glaze, we only want optimizations for whitespace parsing if they have no affect or also speed up minified JSON. Does this make sense?

wilkolbrzym-coder:
Makes total sense. I didn't think about it from the "zero waste" perspective, but you are right—if someone cares about speed, they should just minify their JSON first. No point in slowing down the optimal case just to handle bad data better.
​I'm gonna close this PR for now. I'll look around for other "holes" or bottlenecks in Glaze where I can find some gains without affecting the minified performance.
​Thanks for the explanation!

PR #2270, opened 2026-01-26
SIMD Optimizations and can_error checks

stephenberry:
## SIMD String Escape Refactoring

- Extracted SIMD detection macros into `simd/simd.hpp`, with per-ISA headers: `avx.hpp`, `sse.hpp`, `neon.hpp`.
- Added SSE2 (16-byte) and NEON (16-byte + 64-byte wide) string escape paths alongside the existing AVX2 (32-byte) path.
- Rewrote AVX2 escape detection to use direct `cmpeq`/`movemask` instead of the previous SWAR-in-YMM approach.
- Deduplicated the escape-handling logic into a single `write_escape` lambda shared across all SIMD and SWAR paths.

## Write Optimization (`can_error`)

- Added `static constexpr bool can_error` to `to<JSON, T>` specializations for types that cannot fail during writing (bool, numbers, strings, enums, bitset, flags, null, nullable).
- Object serialization now skips `ctx.error` checks and `ensure_space` calls when all fields are known to be error-free at compile time.
- Extended `array_padding_known` to also cover arrays of objects with computable `fixed_padding`.

## New Utility Header

- Created `util/bit.hpp` consolidating `countr_zero`, `countl_zero`, and `int_log2` (previously scattered across `parse.hpp` and `dtoa.hpp`).
- SIMD headers now include `bit.hpp` directly rather than relying on transitive includes through `write.hpp`.
- Fixed `__has_builtin` checks to match the actual builtin being called.

## Documentation

- Updated README, `installation.md`, and `optimizing-performance.md` to document SSE2, NEON, and the `GLZ_DISABLE_SIMD` option.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -43,3 +43,37 @@ jobs:
           echo "$output" >> "$GITHUB_STEP_SUMMARY"
           echo '```' >> "$GITHUB_STEP_SUMMARY"
         fi
+
+  string_write_benchmark:
+    runs-on: ubuntu-24.04
+
+    env:
+      CC: gcc-14
+      CXX: g++-14
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Verify AVX2 support
+      run: grep -q avx2 /proc/cpuinfo && echo "AVX2 is available"
+
+    - name: Configure CMake
+      run: |
+        cmake -S "$GITHUB_WORKSPACE/benchmarks" -B "$GITHUB_WORKSPACE/build" \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_C_COMPILER=${{env.CC}} \
+          -DCMAKE_CXX_COMPILER=${{env.CXX}}
+
+    - name: Build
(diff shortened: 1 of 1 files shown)


wilkolbrzym-coder:
I’ve finished the portable SWAR implementation for skip_ws including comprehensive benchmarks (which show great results).
I see you’re moving bit utilities to util/bit.hpp. My implementation currently uses a custom countr_zero in parse.hpp, so it will be very easy to swap it for your new centralized version once your PR lands.
Everything is tested and ready on my side. Let me know if you want me to rebase on top of your changes once you're done, or if you'd like to take a look at the SWAR logic now.

stephenberry:
@wilkolbrzym-coder, I'm going to benchmark your pull request as soon as I'm able. And, I'll make sure I get things merged together. I'm still making adjustments and doing more benchmarking for these changes.

wilkolbrzym-coder:
Impressive work on the SIMD refactoring, @stephenberry! The new structure is really clean. I was curious—is AVX-512 support on your roadmap, or are you sticking with AVX2 for now to avoid the complexity/throttling trade-offs? 64-byte processing would be a beastly addition!

PR #2275, opened 2026-01-27
optimize prettified write with space pre-filling (~340ns per op)

wilkolbrzym-coder:
Following up on our discussion about buffer initialization:
​This PR implements Space Pre-filling during buffer allocation, which makes JSON indentation practically free. Instead of double-writing (zero-init then memset spaces), we now leverage the initial resize to fill the buffer with the required padding character.
​Key Changes:
​dump.hpp: Refactored maybe_pad and dumpn_unchecked to handle PaddingChar via template metaprogramming.
​write.hpp: Switched from std::memset to dumpn_unchecked for indentation. For prettified output, this translates to a simple pointer increment (ix += n).
​Zero Overhead: Used if constexpr to ensure that standard minified writes (no padding) are unaffected and maintain their current performance (~628 MB/s).
​Performance Results:
​Prettified Write: ~340 ns for a complex MyStruct (measured via bencher and std::chrono).
​Regression Test: Confirmed no impact on minified JSON throughput.
​Included a new benchmark file: space_prefill_benchmark.cpp to verify these gains and prevent future regressions.


stephenberry:
I just realized that there is a major issue in this proposed optimization. Resizing to another character then skipping works the first time writing to a buffer, but it doesn't work if the buffer is reused. In most real world cases we want to reuse buffers, because this is faster in the long run, but it breaks this optimization.

The sad thing is that being faster in the real world makes benchmarks not look as good for Glaze. But, that's where we need to be careful with how we build benchmarks and we should benchmark buffer reuse.

PR #2280, opened 2026-01-29
Adding glz::pmr::generic

### LINKED ISSUE #2277: glz::generic - ability to customize type and/or allocator
I would like to be able to use PMR (polymorphic memory resources / allocators) with `glz::generic`, but unlike previous versions where I (think I) could achieve this with a `glz::basic_json_t< ... >`, I think this is no longer possible...?

I need to change the types for `array_t` and `object_ t` to something allocator aware, like `std::pmr::vector` and `std::pmr::map` (or, more likely, a pmr-enabled `boost::unordered_flat_map`).

I also need to add allocator-awareness to `generic_json`, I don't know the best way to achieve this. Perhaps inherit from `generic_json` in a custom type?

```cpp
   // Generic json type.
   // Template parameter controls number storage mode for precise integer handling
   template <num_mode Mode = num_mode::f64>
   struct generic_json
   {
      virtual ~generic_json() {}

      using array_t = std::vector<generic_json<Mode>>;
      using object_t = std::map<std::string, generic_json<Mode>, std::less<>>;
      using null_t = std::nullptr_t;
   
      // ...
```

stephenberry:
This is currently a no-go without a polymorphic variant.

STL proposal:
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3153r0.html

### DIFF
--- a/include/glaze/json/generic.hpp
+++ b/include/glaze/json/generic.hpp
@@ -15,6 +15,8 @@
 
 #if __cpp_exceptions
 #include <stdexcept>
+#else
+#include <cstdlib>
 #endif
 
 namespace glz
--- a/include/glaze/json/generic_pmr.hpp
+++ b/include/glaze/json/generic_pmr.hpp
@@ -14,6 +14,9 @@
 #include <variant>
 #include <vector>
 
+// std::pmr::string is not available with the old GCC ABI
+#if !defined(_GLIBCXX_USE_CXX11_ABI) || _GLIBCXX_USE_CXX11_ABI != 0
+
 #include "glaze/json/generic.hpp"
 
 namespace glz::pmr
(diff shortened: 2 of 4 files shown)


wilkolbrzym-coder:
PMR support for generic would be a massive win for performance! It's a shame std::variant makes allocator propagation such a pain.
One small thing I noticed: in the move constructor generic(generic&& other, const allocator_type& alloc), we're asserting that allocators match. In release builds, if they don't match, we might end up with memory corruption since the internal containers won't re-bind to the new resource. Would it be safer to fallback to a deep copy if alloc != other.alloc?

stephenberry:
@wilkolbrzym-coder, yeah we should probably throw an error if on a move allocators don't match. But, I don't think this code will be merged as it is now, because we need a polymorphic allocator aware version of a variant.

PR #2292, opened 2026-02-01
Remove move from conn->request_

stephenberry:


stephenberry:
@clagix, I'm removing the std::move, as you're right that this is safer and not a performance bottleneck.

wilkolbrzym-coder:
It seems send_error_response is gone (which is good), but the raw send_response(std::shared_ptr<socket_type> socket, ...) method is still present at the bottom of the http_server struct. Since it appears unused, it can be removed too.

clagix:
> It seems send_error_response is gone (which is good), but the raw send_response(std::shared_ptr<socket_type> socket, ...) method is still present at the bottom of the http_server struct. Since it appears unused, it can be removed too.

@wilkolbrzym-coder I don't see this method anymore. I think it has somehow resolved itself.

wilkolbrzym-coder:
> > It seems send_error_response is gone (which is good), but the raw send_response(std::shared_ptr<socket_type> socket, ...) method is still present at the bottom of the http_server struct. Since it appears unused, it can be removed too.
> 
> @wilkolbrzym-coder I don't see this method anymore. I think it has somehow resolved itself.

@clagix yes the problem has already been solved

PR #2296, opened 2026-02-03
ZMEM Format

stephenberry:


wilkolbrzym-coder (changes requested):
Review – PR #2296

Not approved. This PR must not land in main in its current state.

1. Critical OOB from unvalidated vector offsets

read_vector_payload and from<ZMEM, std::vector<T>> for variable element types compute:

const char* elem_it = data_start + offsets[i];
const char* elem_end = data_start + offsets[i + 1];

without checking:

- offsets[i] <= offsets[i + 1]
- offsets[i] <= offsets[count]
- offsets[count] <= available_size

A malicious buffer with descending offsets makes elem_end - elem_it negative, which becomes a huge size_t. resize and memcpy then cause OOB read/write.

Security bug.

2. zmem::optional<T> broken for non-integral T

Reader calls zmem::byteswap_le(value.value), which uses std::byteswap only valid for integral types.

For zmem::optional<MyStruct> this does not compile.

3. optional_alignment_v wrong for over-aligned T

optional_alignment_v = sizeof(T) >= 8 ? 8 : alignof(T)

If alignof(T) > 8, the struct is forced to align 8 while containing an over-aligned member. UB.

Tests only cover uint8_t..uint64_t.

4. unordered_map support is fake

make_sorted_entries sorts with a.first < b.first.

unordered_map keys without operator< fail to compile, despite documented support.

Missing concept/requires.

5. Lazy view is not bounds-safe

as_fixed() reinterprets data_ without checking size or alignment.

compute_field_ptr() returns offsets without range validation.

Truncated buffer gives OOB read.

6. CI red

9 failing jobs, mostly clang and gcc 13.

Verdict

Request changes. The format needs robust offset validation, correct optional handling, real unordered_map support, and bounds-safe lazy views before it can be considered for merge.

cc @stephenberry

PR #2325, opened 2026-02-19
glz::generic now preserves insertion order -- glz::ordered_map & glz::ordered_small_map

stephenberry:
### Breaking Change

`glz::generic` now preserves insertion order of JSON keys. Previously keys were lexicographically sorted via `std::map`. The new default backend is `glz::ordered_small_map`, a compact ordered map optimized for typical JSON objects.

Users who depend on sorted-key output can restore the old behavior:

```cpp
template <class T>
using sorted_map = std::map<std::string, T, std::less<>>;

using sorted_generic = glz::generic_json<glz::num_mode::f64, sorted_map>;
```

### What changed

- **`glz::generic_json` is now a template** accepting a configurable map type (`template <num_mode, template<class> class MapType>`). The default `glz::generic` alias uses `glz::ordered_small_map`.
- **`glz::ordered_small_map<T>`** — new 24-byte ordered map for string-keyed JSON objects. Preserves insertion order, uses contiguous storage, linear scan for ≤8 keys, and a lazily-built sorted hash index with bloom filter for larger maps.
- **`glz::ordered_map<T>`** — new general-purpose ordered map with hash index, supporting both string and non-string keys.
- **`glz::sweethash`** — new lightweight hash functions (`sweet32`, `sweet64`) used by the ordered maps.

### Performance vs `std::map` (cold-cache, Apple M-series)

| Benchmark | `ordered_small_map` | `std::map` | Delta |
|---|---|---|---|
| read small (8 keys) | 389 MB/s | 259 MB/s | **+50%** |
| read medium (nested) | 342 MB/s | 292 MB/s | **+17%** |
| read large (100 obj) | 267 MB/s | 205 MB/s | **+30%** |
| write medium (reused buf) | 821 MB/s | 823 MB/s | ~tie |
| key lookup (8 keys) | 727 MB/s | 363 MB/s | **+100%** |
| roundtrip medium | 430 MB/s | 388 MB/s | **+11%** |

### New files

- `include/glaze/containers/ordered_small_map.hpp`
- `include/glaze/containers/ordered_map.hpp`
- `include/glaze/hash/sweethash.hpp`
- `tests/ordered_small_map_test/` (18 tests, 2603 assertions)
- `tests/ordered_map_test/` (44 tests, 40380 assertions)
- `benchmarks/generic_benchmark.cpp`

### DIFF
--- a/docs/generic-json.md
+++ b/docs/generic-json.md
@@ -4,6 +4,15 @@ While Glaze is focused on strongly typed data, there is basic support for comple
 
 If nothing is known about the JSON structure, then [glz::generic](https://github.com/stephenberry/glaze/blob/main/include/glaze/json/generic.hpp) may be helpful, but it comes at a performance cost due to the use of dynamic memory allocations. The previous `glz::json_t` name remains as an alias for backwards compatibility.
 
+> **Note:** `glz::generic` preserves the insertion order of JSON keys. When you parse a JSON object and re-serialize it, the keys will appear in the same order as the original document. If you need lexicographically sorted keys, you can use `std::map` as the map backend:
+>
+> ```cpp
+> template <class T>
+> using sorted_map = std::map<std::string, T, std::less<>>;
+>
+> using sorted_generic = glz::generic_json<glz::num_mode::f64, sorted_map>;
+> ```
+
 ## Generic Type Variants
 
 Glaze provides three generic JSON types with different number storage strategies:
@@ -104,7 +113,7 @@ glz::generic json = {
 };
 std::string buffer{};
 glz::write_json(json, buffer);
-expect(buffer == R"({"answer":{"everything":42},"happy":true,"list":[1,0,2],"name":"Stephen","object":{"currency":"USD","value":42.99},"pi":3.141})");
+expect(buffer == R"({"pi":3.141,"happy":true,"name":"Stephen","nothing":null,"answer":{"everything":42},"list":[1,0,2],"object":{"currency":"USD","value":42.99}})");
 ```
 
 ## get() vs as()
@@ -289,7 +298,7 @@ auto names_list = glz::get<std::list<std::string>>(json, "/names");
 auto names_array = glz::get<std::array<std::string, 3>>(json, "/names");
 ```
 
-This works because `glz::generic` stores arrays as `std::vector<glz::generic>` and objects as `std::map<std::string, glz::generic>`. When you request a specific container type, Glaze deserializes the generic representation into your desired type.
+This works because `glz::generic` stores arrays as `std::vector<glz::generic>` and objects as `glz::ordered_small_map<glz::generic>`. When you request a specific container type, Glaze deserializes the generic representation into your desired type.
 
 ## See Also
 
@@ -297,3 +306,4 @@ This works because `glz::generic` stores arrays as `std::vector<glz::generic>` a
 - [JSON Patch (RFC 6902)](./json-patch.md) - Apply structured patches to `glz::generic` documents
 - [JSON Merge Patch (RFC 7386)](./json-merge-patch.md) - Apply partial updates to `glz::generic` documents
 - [JSON Pointer Syntax](./json-pointer-syntax.md) - Path syntax for navigating JSON documents
+- [Ordered Maps](./ordered-maps.md) - Details on `glz::ordered_small_map` and `glz::ordered_map`
--- a/docs/ordered-maps.md
+++ b/docs/ordered-maps.md
@@ -1,0 +1,180 @@
+# Ordered Maps
+
+Glaze provides two insertion-order-preserving map containers. Both maintain the order in which keys are inserted while providing fast lookups.
+
+## glz::ordered_small_map
+
+`#include "glaze/containers/ordered_small_map.hpp"`
+
+A compact, string-keyed ordered map optimized for small to medium JSON objects. This is the default map backend for [`glz::generic`](./generic-json.md).
+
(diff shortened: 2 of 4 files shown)


ElevationsRPG:
Hey stephen! nice work on this PR! just wondering though is it possible to use glz::read and glz:write with this ? i would of kind of proffered if the sort functionality was just a simple compile-time option that i can pass to either:

e.g

sortmapsbyinsert = true // false = legicograpphically like before

Or if you can give me more examples of how it might be used with glz::generic (The one above is a bit confusing to me)

stephenberry:
@ElevationsRPG, this order only pertains to `glz::generic`, and this change preserves insertion order so that JSON properly roundtrips, rather than potentially getting sorted.

I just merged in a `glz::generic_sorted` alias that uses a std::map underneath and will sort your keys. But, generally keeping things in the order the original JSON was in is better.

The reason this can't be a compile time option for `glz::read`/`glz::write` is that the order data is stored in is built into the `glz::generic_json` specialization. This means that it isn't different parsing logic, instead it is a different container.

I hope this clarifies things. If you want the old sorted approach you can just use `glz::generic_sorted`.

ElevationsRPG:
> @ElevationsRPG, this order only pertains to `glz::generic`, and this change preserves insertion order so that JSON properly roundtrips, rather than potentially getting sorted.
> 
> I just merged in a `glz::generic_sorted` alias that uses a std::map underneath and will sort your keys. But, generally keeping things in the order the original JSON was in is better.
> 
> The reason this can't be a compile time option for `glz::read`/`glz::write` is that the order data is stored in is built into the `glz::generic_json` specialization. This means that it isn't different parsing logic, instead it is a different container.
> 
> I hope this clarifies things. If you want the old sorted approach you can just use `glz::generic_sorted`.

Ohhh so the read simply goes based on the order the json was saved in ?? Then i guess what im asking for is the ability to sort the write because im pretty sure its not the same order that the elements are stored in the map in memory so:

```
std::string myjsonsavedmap;
std::unordered_map<int, int> map1;
example[0] = 1;
example[1] = 2;
example[2] = 3;
glz::write_json(example, myjsonsavedmap);

for (auto [id, value] : map1)
{
std::cout << id << "," << value << std::endl; // print out the elements of map 1 (you will see the order)
}

std::unordered_map<int, int> map2;
glz::read_json(example2, myjsonsavedmap); //read the json back into map 2

for (auto [id, value] : map2)
{
std::cout << id << "," << value << std::endl; // print out the elements of map 2 the order is different from map 1
}

Pretty sure the order is not the same if you compare iteration output.
```

stephenberry:
@ElevationsRPG, if you replace `std::unordered_map<int, int>` with `glz::ordered_map<int, int>`, then your data will be kept in order. Available in glaze/containers/ordered_map.hpp

ElevationsRPG:
> @ElevationsRPG, if you replace `std::unordered_map<int, int>` with `glz::ordered_map<int, int>`, then your data will be kept in order. Available in glaze/containers/ordered_map.hpp

ah but what if i want to use boost::unordered_flat_map ? and still preserve order of elements ? the reason why is because there are scenarios where i may iterate the map and thats where i like the data to be consistent with what the original map order was lol if that makes sense. i understand it seems odd why want ordererd map but thats not really it, its more that writing then reading the map back is where the sort problem occurs.

So what im trying to say is the original unordred map in memory is still ordered in the sense that it doesnt change the element order for the lifetime of the object so when you iterate the map every object is always in the same order.

So my problem is that when we write that map using glaze to the std::string buffer; and then read it back into a duplicate map, the original map order is now lost. The elements are inserted which is good but the original order was lost if that makes sense.

Heres best way i can describe it:

You have two maps:

```
boost::unordered_flat_map<int, int> map1;
boost::unordered_flat_map<int, int> map2;

//now you go to insert data in both maps:

map1[0] = 1;
map1[1] = 2;
map1[2] = 3;

map2[0] = 1;
map2[1] = 2;
map2[2] = 3;

Now when you iterate both these maps the order is ALWAYS the same even though the container is unordered, their order doesnt change its identical
```

stephenberry:
boost::unordered_flat_map has the same limitation as std::unordered_map: iteration order is not part of the container’s guaranteed state. Glaze writes keys in whatever order the map currently iterates, and on read it inserts keys back, but unordered containers are free to lay buckets out differently, so map2 iteration can differ from map1 even with identical key/value data.

So this roundtrip can preserve values, but not a stable unordered-map iteration order. This is what is meant by `unordered`. If you need deterministic order after write/read, use glz::ordered_map (or std::map) for that field, or keep a side std::vector<Key> to define iteration order.

ElevationsRPG:
> unordered containers are free to lay buckets out differently

Ah yes i understand that this is the intended behavior, but its a little bit strange because i am 99% sure the same thing happens even with std::map ? where the order doesnt match the original container version (after glaze parses it).

So im not asking you to fix it but rather im asking for an option of some kind that will enable glaze to preserve the order.

(it could even have something to do with me using my own hasher ?) I only ever noticed the problem with maps but vector seemed to work fine

stephenberry:
@ElevationsRPG, `std::map` doesn't preserve the input order, but it sorts it in lexicographical order. But, once data is in the `std::map` it will round trip in the same order with Glaze every time. It's just if the input JSON is not in alphabetical order then `std::map` will reorder the data.

If you want to preserve the order of the input JSON `glz::ordered_map` will work. But, `std::map` does preserve its own order, just not the order of arbitrary input that might not be sorted.

Does this make sense?

PR #2339, opened 2026-02-28
Implement `skip_read_constraint` option

### LINKED ISSUE #2338: skip_self_constraint does not work
skip_self_constraint does not work, I'm not able to bypass checks.

To reproduce

```
#include <glaze/glaze.hpp>
#include <iostream>
#include <string>

struct person {
   int age{};
};

template <>
struct glz::meta<person> {
   using T = person;

   static constexpr auto limit_age = [](const T&, int age) { return age >= 18; };

      "age", read_constraint<&T::age, limit_age, "You must be at least 18 years old">
   );
};

int main() {
   person p{};
   
    struct skip_constraint_opts : glz::opts
    {
        bool skip_self_constraint = true;
    };

    constexpr skip_constraint_opts opts{};

   std::string json_data = R"({"age": 15})";

   // Parsing con le opzioni specificate
   auto err = glz::read<opts>(p, json_data);

   if (err) {
      std::cout << "Validatione error: " << glz::format_error(err, json_data) << std::endl;
   } else {
      std::cout << "Age is valid: " << p.age << std::endl;
   }

   return 0;
}
```

I expect age is valid, I get validation error.
I'm using version 7.0.2.

eoan-ermine:
Closes #2338

### DIFF
--- a/include/glaze/core/constraint.hpp
+++ b/include/glaze/core/constraint.hpp
@@ -72,7 +72,7 @@ namespace glz
                      parse<Format>::template op<Opts>(input, ctx, it, end);
                      if (bool(ctx.error)) [[unlikely]]
                         return;
-                     if (!check_skip_read_constraint(Opts)) {
+                     if constexpr (!check_skip_read_constraint(Opts)) {
                         auto success = (value.val.*(value.constraint))(input);
                         if (not success) {
                            ctx.error = error_code::constraint_violated;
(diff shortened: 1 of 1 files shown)


stephenberry:
@eoan-ermine, thanks for this pull request. I made some tweaks based on the copilot feedback. Looks good and I'm merging it in!

PR #2345, opened 2026-03-04
Fix enum hashed-lookup bounds check on truncated JSON strings

huangminghuang:
## Summary
This PR fixes a bounds issue in enum reflection parsing when using the sized-hash fast path on malformed/truncated JSON input.

## What changed
- Added an additional guard in `reflect.hpp` to ensure `HashInfo.unique_index` is within the remaining input span before indexing.
- Added a regression test in `json_reflection_test.cpp` that feeds a truncated enum payload and asserts parsing fails cleanly (instead of risking out-of-bounds access).

## Why
A truncated quoted enum value could pass earlier length checks but still allow an invalid index into the remaining buffer. The new condition prevents that read.


stephenberry:
@huangminghuang thanks for this important fix!

PR #2354, opened 2026-03-07
make stream_request buffer size configurable + update docs

frak0d:
I was facing connections drops and slow speeds with default 1 MiB buffer when streaming data from metabase. Increasing buffer size to 8 MiB tripled my throughput and got rid of connection dropouts.

The max_buffer_size parameter was hardcoded earlier, this commit make it user configurable while keeping 1 MiB default. Also updated the out of date docs regarding `stream_request`

---

On the side note i must say, the order of members in `stream_request_params` is quite unfortunate, especially when you are using GCC and hence can't reorder designated initializers. perhaps we could do a `stream_request_v2` that takes params in different order, so that handlers and normal parameters don't interleave.

### DIFF
--- a/docs/networking/http-client.md
+++ b/docs/networking/http-client.md
@@ -182,45 +182,45 @@ void post_json_async(
 The HTTP client supports streaming requests, which allow you to receive data in chunks.
 
 ```cpp
-std::shared_ptr<http_stream_connection> stream_request(const stream_request_params& params);
+std::shared_ptr<http_stream_connection> stream_request_v2(const stream_request_params_v2& params);
 ```
 
-The `stream_request_params` struct contains the following fields:
+The `stream_request_params_v2` struct contains the following fields:
 
 ```cpp
-struct stream_request_params {
-    std::string url;
-    http_data_handler on_data;
-    http_error_handler on_error;
+struct stream_request_params_v2 {
     std::string method{"GET"};
+    std::string url;
+    std::chrono::seconds timeout{30s};
+    stream_read_strategy strategy{stream_read_strategy::bulk_transfer};
+    size_t max_buffer_size{1024 * 1024};
     std::string body;
     std::unordered_map<std::string, std::string> headers;
     http_connect_handler on_connect;
     http_disconnect_handler on_disconnect;
-    std::chrono::seconds timeout{30s};
-    stream_read_strategy strategy{stream_read_strategy::bulk_transfer};
-    std::function<bool(int)> status_is_error;
-    size_t max_buffer_size{1024 * 1024};
+    http_data_handler on_data;
+    http_error_handler on_error;
+    std::function<bool(int)> status_is_error{[](int status){ return status >= 400; }};
 };
 ```
 
--   `url`: The URL to request.
--   `on_data`: A callback that's called when data is received.
--   `on_error`: A callback that's called when an error occurs.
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this PR and benchmarking! I agree with `stream_request_params_v2`, so go head and add it, and I would recommend adding a `[[deprecated(...)]]` attribute on the first version pointing to the new version. This will let us sooner turn `stream_request_params_v2` back into `stream_request_params`.

frak0d:
which order would you prefer for handlers?
```
.on_connect
.on_disconnect
.on_data
.on_error
.status_is_error
```
OR
```
.on_connect
.on_data
.on_disconnect
.status_is_error
.on_error
```

stephenberry:
I think I like this order the best:
.on_connect
.on_disconnect
.on_data
.on_error
.status_is_error

stephenberry:
@frak0d, do you think this is ready to merge?

frak0d:
yes its ready

frak0d:
@stephenberry what is the correct way of doing read_stream_json on data coming from stream_request ? 

As i understand it, there needs to be a custom thread safe buffer derived from std::streambuf into which the on_data callback can push data and read_stream_json will pull data.

Is there such buffer already provided by glaze?
Or Am i missing something?

stephenberry:
@frak0d, Glaze doesn't currently have a production thread-safe bridge buffer for this, but the approach you're describing is right.

The streaming JSON reader you'd want is glz::json_stream_reader<T> (from glaze/json/json_stream.hpp), which takes any std::istream and parses complete JSON values one at a time (supports both JSON arrays and NDJSON).

To bridge stream_request's on_data callback with the reader, you need a thread-safe std::streambuf that the callback writes into and the reader pulls from. There's a pipe_buffer example in the test suite (tests/istream_buffer_test/istream_buffer_test.cpp, line 42) that does exactly this — it uses a mutex + condition_variable to synchronize a writer thread and a reader thread via a std::streambuf interface.

Rough sketch:

```c++
pipe_buffer pipe;
std::istream stream(&pipe);

auto conn = client.stream_request_v2({
    .url = "...",
    .on_data = [&](std::string_view chunk) {
        pipe.write(chunk.data(), chunk.size());
    },
    .on_disconnect = [&]() {
        pipe.close();
    }
});

// On reader thread:
glz::json_stream_reader<MyType> reader(stream);
MyType value;
while (!reader.read_next(value)) {
    process(value);
}
```

I could look into adding a proper `pipe_buffer` to the library if you're interested.

frak0d:
pipe_buffer is trivial to implement so no need to add that. Besides, its only a hack in the first place. json stream reader is pull based but glaze networking is push based.

im hoping this would be more elegantly solved with coroutine based networking in future. something like:
```cpp
auto stream = co_await client.stream_connect(params)

if (stream)
{
    auto reader = async_json_reader<MyType>(stream);
    for (MyType parsed : reader.data())
        process(parsed);
}
```

```cpp
// data() is a coroutine generator
async_json_reader::data() -> std::generator<T>

// somewhere in data() internal loop:
while (stream.active())
{
    ...
    while (not enough buffered data && stream.active()) {
        auto [chunk,err] = co_await stream.next_chunk();
        internal_buffer.append(chunk);
    }
    ...
    do actual parsing
    ...
    co_yeild parsed;
}
```

PR #2359, opened 2026-03-08
Fix YAML double-quoted \xXX being interpreted as a byte instead of a codepoint

nghiaho310pf:
https://yaml.org/spec/1.2.2/#57-escaped-characters:~:text=Escaped%208%2Dbit%20Unicode%20character


stephenberry:
Thanks for this fix!

PR #2377, opened 2026-03-15
Better error messages for unknown variant type

### LINKED ISSUE #2299: Support `glx::context` in `variant` purely structs
Hi, @stephenberry. I am implementing code where there is a type and corresponding structure parameters. In the case of an unknown type, I want to treat it as an error, but in such a way that JSON and a pointer to the location of the incorrect value are printed (as is done in standard error handling using the library using `glz::error_ctx`). I do this as follows: in the constructor of the unknown type, I immediately throw an exception, but in this case, I cannot get any information about JSON.

Therefore, I have the following request for you as a developer: can you do something like glz::context so that, first, parsing can be interrupted without exceptions, and second, so that it is possible to ultimately get the standard `glz::error_ctx` to output JSON with a user message that will be set via `glx::context`. Thank you in advance for your response.

```c++
#include <enchantum/enchantum.hpp>
#include <fmt/ranges.h>
#include <glaze/glaze_exceptions.hpp>

enum class type_e : uint8_t
{
    SOME_TYPE
};

struct some_type
{
   std::string type;
};

struct unknown_type
{
    unknown_type() // <--- here glz::context or something like this
    {
        static auto const errorMessage = fmt::format("Unknown type. Supported types: {}", fmt::join(enchantum::names<type_e>, ", "));
        throw std::runtime_error(errorMessage);
    }
};

using types_variant_t = std::variant<some_type, unknown_type>;

template <>
struct glz::meta<types_variant_t>
{
    static constexpr std::string_view tag = "type";
    static constexpr auto ids = std::array{"SOME_TYPE"};
};

TEST_CASE("Example")
{
    // static constexpr std::string_view config = R"({"type":"SOME_TYPE"})"; // Sunny case
    static constexpr std::string_view config = R"({"type":"UNKNOWN"})";

    types_variant_t typesVariant;
    glz::ex::read<glz::opts{.minified = true, .error_on_missing_keys = true}>(typesVariant, config);
}
```

Output:
```bash
... with exception message:
  Unknown type. Supported types: SOME_TYPE
```

stephenberry:
## Improved error messages for tagged variant type mismatches

Closes #2299

When a tagged `std::variant` encounters an unknown type tag and no default/catch-all handler exists, the error message now lists the supported type IDs. This gives users actionable information without needing to throw exceptions from a catch-all constructor.

### Before

```
1:19: no_matching_variant_type
   {"action":"DELETE","id":"456"}
                     ^
```

### After

```
1:19: no_matching_variant_type
   {"action":"DELETE","id":"456"}
                     ^ supported types: CREATE, UPDATE
```

### Changes

- **`include/glaze/core/meta.hpp`** — Added `variant_ids_string_v<T>`, a compile-time variable template that joins a variant's string type IDs into a comma-separated message (e.g. `"supported types: CREATE, UPDATE"`).
- **`include/glaze/json/read.hpp`** — Set `ctx.custom_error_message` at the two tagged-variant error sites where an unknown tag has no default handler.
- **`tests/json_test/variant_ambiguous_test.cpp`** — Added tests verifying the full `format_error` output for unknown tags (tag-first and tag-last orderings).

### DIFF
--- a/include/glaze/core/meta.hpp
+++ b/include/glaze/core/meta.hpp
@@ -723,6 +723,40 @@ namespace glz
 
    // Compile-time string listing the variant type IDs (e.g. "TYPE_A, TYPE_B, TYPE_C")
    // Used for error messages when no matching variant type is found
+   namespace detail
+   {
+      template <is_variant T>
+      inline constexpr size_t variant_ids_string_len = [] {
+         constexpr auto& ids = ids_v<T>;
+         constexpr auto N = ids.size();
+         size_t len = 17; // "supported types: "
+         for (size_t i = 0; i < N; ++i) {
+            if (i > 0) len += 2; // ", "
+            len += ids[i].size();
+         }
+         return len;
+      }();
+
+      template <is_variant T, size_t Len>
+      inline constexpr auto variant_ids_joined = [] {
+         constexpr auto& ids = ids_v<T>;
+         constexpr auto N = ids.size();
+         std::array<char, Len + 1> arr{};
+         const char* prefix = "supported types: ";
+         size_t pos = 0;
+         for (size_t i = 0; i < 17; ++i) arr[pos++] = prefix[i];
+         for (size_t i = 0; i < N; ++i) {
+            if (i > 0) {
+               arr[pos++] = ',';
+               arr[pos++] = ' ';
+            }
+            for (auto c : ids[i]) arr[pos++] = c;
+         }
+         arr[Len] = '\0';
+         return arr;
+      }();
+   }
+
    template <is_variant T>
    inline constexpr std::string_view variant_ids_string_v = [] {
       constexpr auto& ids = ids_v<T>;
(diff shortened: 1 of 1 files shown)


stephenberry:
@GTruf, how does this solution look?

GTruf:
@stephenberry, overall, everything's fine, but in a real-world scenario, there might be a NONE/UNKNOWN value in the enum, and if that's also displayed as “supported,” it wouldn't be ideal. But otherwise, yes, it's exactly what I needed.

stephenberry:
I think if the user has a NONE type that is unsupported, then they can exclude it from their variant or provide custom handling for the type. Sometimes a NONE might be valid and just an empty object, which this supports. If further customization is needed to handle cases like this we can make another PR. But, thanks for the feedback! I'll merge it in as soon as everything is passing.

GTruf:
Overall, you're right, it's better to focus on real needs, so I'm looking forward to seeing your PR merged. Thank you very much.

Issue #2386, opened 2026-03-20
glz::custom getter returning std::nullopt results in JSON field with null instead of being omitted


Zetonem:
### Description

When using glz::custom with a getter that returns std::optional in a glz::object, the serialization includes the field with value null even if the getter returns std::nullopt. The expected behavior (consistent with how std::optional members are handled) is that the field should be omitted from the output entirely.

### Steps to Reproduce

Define a struct with a glz::meta specialization containing an object that uses glz::custom with a getter returning std::optional<std::string>.

In the getter, return std::nullopt.

Serialize the struct with glz::write_json.

Observe the resulting JSON string.

### Example Code

```
#include <glaze/glaze.hpp>

#include <iostream>

struct my_struct_optional
{
    int value = 42;
};

template<>
struct glz::meta<my_struct_optional>
{
    using T = my_struct_optional;
    static constexpr auto set_status = [](T& s, std::optional<std::string> value)
    {
        if (value.has_value())
        {
            s.value = 50;
        }
        else
        {
            s.value = 0;
        }
    };
    static constexpr auto get_status = [](auto& s) -> std::optional<std::string>
    {
        if (s.value > 50)
        {
            return "high";
        }
        else
        {
            return std::nullopt;
        }
    };
    static constexpr auto value = glz::object("value", &T::value, "status", glz::custom<set_status, get_status>);
};

int main()
{
    my_struct_optional obj{};
    obj.value = 42;

    std::string buffer;
    std::ignore = glz::write_json(obj, buffer);

    if (buffer == R"({"value":42})")
    {
        std::cout << "Test passed" << std::endl;
    }
    else
    {
        std::cout << "Test failed: '" << buffer << "'" << std::endl;
    }
}
```

### Expected Behavior

The output should be {"value":42} (the status field should be omitted because the getter returned std::nullopt).

### Actual Behavior

The output is {"value":42,"status":null}. The field is present with a null value.

### Environment

- Glaze version: 7.2.1
- Compiler: gcc 14.2.1 (C++23)
- OS: Linux

stephenberry:
Thanks for sharing this issue. Should be fixed with the merge of #2390

PR #2388, opened 2026-03-20
Add custom type handling for nullable fields in JSON serialization

Zetonem:
### Summary

When a glz::object field uses glz::custom<setter, getter> and the getter returns std::optional<T> that is std::nullopt, Glaze currently serializes the field as "field": null instead of omitting the field (under skip_null_members=true). This PR adds a regression test and fixes the JSON writer so the field is skipped when the custom getter produces a null/empty optional.

### Changes

1. Fix skip logic for JSON serialization

- Updated the Opts.skip_null_members handling for fields whose value type is glz::custom_t.
- For glz::custom_t, the serializer now invokes the custom getter (supports member-function/member-object pointer and invocable forms), detects whether the getter result is “null” for nullable return types (notably std::optional / optional-like via existing nullable_* concepts) and skips writing the field when the getter result is empty (std::nullopt -> omitted field)

2. Ensure glz::custom_t is treated as a candidate for skipping

Expanded the maybe_skipped condition so that when skip_null_members=true, glz::custom_t fields are considered “potentially skippable”, allowing the runtime logic in json/write.hpp to run.

### Issue
https://github.com/stephenberry/glaze/issues/2386

### DIFF
--- a/include/glaze/core/reflect.hpp
+++ b/include/glaze/core/reflect.hpp
@@ -380,6 +380,9 @@ namespace glz
    template <class T, size_t I>
    using field_t = std::remove_cvref_t<refl_t<T, I>>;
 
+   template <class V>
+   consteval bool custom_to_is_nullable_or_false();
+
    template <auto Opts, class T>
    inline constexpr bool maybe_skipped = [] {
       if constexpr (reflect<T>::size > 0) {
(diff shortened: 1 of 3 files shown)


stephenberry:
@Zetonem, thanks for submitting this. I see a few completeness issues. I'll either contribute to this PR, or open a similar one and close this. But, this is helpful to see an initial solution, thanks!

stephenberry:
I just merged in a similar fix with #2390, so I'm closing this. But, thanks for contributing and feel free to keep posting issues and PRs!

Zetonem:
> I just merged in a similar fix with #2390, so I'm closing this. But, thanks for contributing and feel free to keep posting issues and PRs!

Thank you!

PR #2391, opened 2026-03-21
Fix GNU-style flag passing to MSVC frontend

annihilatorq:
## Issue
When building `glaze` with `clang-cl`, the build fails because some branches in `tests/CMakeLists.txt` and `fuzzing/CMakeLists.txt` select GNU-style flags based only on `CMAKE_<LANG>_COMPILER_ID=Clang`

This is incorrect for `clang-cl`: although the compiler ID is `Clang`, its frontend is MSVC-compatible and therefore expects MSVC-style command-line options. To solve this issue, CMake introduced [CMAKE_LANG_COMPILER_FRONTEND_VARIANT](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_FRONTEND_VARIANT.html) flag in v3.14

## Fix
This PR updates the compiler checks in `tests/CMakeLists.txt` and `fuzzing/CMakeLists.txt` to account for the frontend variant, so `clang-cl` no longer enters the GNU-style flag branch. As a result, the affected targets configure and build correctly with `clang-cl`

gotta see what CI will say, and if there are any issues (which shouldn't be the case, but who knows 😁), I'll take care of them

P.S. I also removed `target_link_options(glz_test_common INTERFACE -ftime-trace)` since `target_link_options()` applies to the linker, while afaik the flag is intended for compile-time tracing


annihilatorq:
I also wanted to ask if there's a reason why `glaze` hasn't added a basic workflow for building and testing with clang-cl?
If it's because of a lack of time or because this task isn't a high priority, I'd be more than happy to help

annihilatorq:
hmm, I guess we can also simplify these checks, so instead of
```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC")
  # Something for compiler with MSVC command line style
endif()
```
we can use
```cmake
if(MSVC)
  # Something for compiler with MSVC command line style
endif()
```
since, as stated in CMake [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html) variable documentation, it is `Set to true when the compiler is some version of Microsoft Visual C++ or another compiler simulating the Visual C++ cl command-line syntax.`

should we change our current approach (`CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC"` to `MSVC`), or would `glaze` prefer to stick to a single style for all compilers?

annihilatorq:
seems like `glaze` already uses `if(MSVC)` 
https://github.com/stephenberry/glaze/blob/6a7d6996f4e1b31c826e52a1b1c66d3cb9a9a020/CMakeLists.txt#L22
so I replaced the long `if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "MSVC")` with `if(MSVC)`

stephenberry:
Thanks for this contribution for clang-cl. I'll look it over in a bit more detail ASAP. But, I'm also open for a PR that tests clang-cl if you'd like to add it.

stephenberry:
Looks good!

PR #2396, opened 2026-03-23
fedora ci test timeout

stephenberry:
Trying to figure out why rpm-build:fedora-rawhide-s390x always hangs in CI. Was having runner issues, but I'm thinking this might be a Glaze issue.


stephenberry:
Oh, I think the runner is just never pulling in the job.

PR #2408, opened 2026-03-29
Modules

### LINKED ISSUE #1586: C++ module interface
I see your project using C++23, it doesn't have a C++ module yet, and I would like to provide one using https://github.com/msqr1/importizer ? I would be happy to do this if you want!

DockedFerret800:
This pull request converts the library to a native modules version.

There are a few compiler errors that I found in MSVC during the conversion that make most targets uncompilable, altough some do compile, e.g. `json_test`. I also found MSVC to have the best support for `import std`. I used Clang-22 on Ubuntu WSL as my second compiler. It's also currently not possible to compile the full project because of certain bugs; however, some targets compile, including `json_test`. I think it's a bit more difficult with `import std` on Clang than it is with MSVC, I'll have a more detailed look soon. 

There is still a lot to do, at least to get it compiling. I'm opening this draft pr so that you can see progress.

What is left to do:
- [x] Fix compilation bugs.
- [x] Achieve full coverage.
- [x] Decide how to handle includes. More research required.
- [ ] Ensure CI workflows are working.
- [ ] Update documentation.

Closes #1586 

UPD 3: The project now fully compiles with Clang-22.


DockedFerret800:
Even though it doesn't fully compile yet, which is something I will be fixing over the next few days, we are still very close to getting this done!

ElevationsRPG:
@DockedFerret800 @stephenberry  Great to see the progress on modernizing the library! Will this native modules support be an optional alternative to the current header-only implementation?

Many of us are on stable toolchains or build systems where C++20/23 module support isn't fully mature yet. I’d personally prefer to keep the current #include workflow as the primary or at least a supported fallback is the plan to eventually replace headers entirely, or to offer a glaze.ixx alongside the existing headers?

DockedFerret800:
@ElevationsRPG Great to see interest in this! 

In discussions, we've reached a consensus to convert the library to native modules; however, it will still be possible to use headers. We are researching how this would be done. The current aim is to generate headers from module files, which is what was proposed by @stephenberry.

ElevationsRPG:
> @ElevationsRPG Great to see interest in this!
> 
> In discussions, we've reached a consensus to convert the library to native modules; however, it will still be possible to use headers. We are researching how this would be done. The current aim is to generate headers from module files, which is what was proposed by @stephenberry.

Ah, that’s a very interesting approach! Generating headers from module files sounds ambitious if it works, it would definitely solve the dual-maintenance headache.

Would it be worth keeping this in a separate feature branch (e.g., feature/native-modules) for now? Given that compiler support (especially with import std in Clang vs MSVC) is still evolving, a separate branch would let the community test the modular version without risking the stability of the main header-only workflow

DockedFerret800:
@ElevationsRPG, in our discussions, we also concluded that the ecosystem support is mature enough.  The aim of Glaze was always to drop older compiler versions as newer are released (see Readme). That said, this will definitely not be merged until we ensure this is stable, but the end goal is to merge this one day.

ElevationsRPG:
> @ElevationsRPG, in our discussions, we also concluded that the ecosystem support is mature enough. The aim of Glaze was always to drop older compiler versions as newer are released (see Readme). That said, this will definitely not be merged until we ensure this is stable, but the end goal is to merge this one day.



> @ElevationsRPG, in our discussions, we also concluded that the ecosystem support is mature enough. The aim of Glaze was always to drop older compiler versions as newer are released (see Readme). That said, this will definitely not be merged until we ensure this is stable, but the end goal is to merge this one day.

That’s a fair point, and it’s great to see Glaze continuing to push the envelope. I'm definitely looking forward to the improved isolation and faster incremental build times that native modules offer. Once the ecosystem and compiler support (especially for import std) settle a bit more across all platforms, I’ll be keen to migrate my own workflows. I appreciate the focus on ensuring stability (especially knowing there are still bugs glaze needs to fix) before the final merge!

stephenberry:
@DockedFerret800, is there a good reason for `std::size_t` and `std::uint64_t` rather than just `size_t` and `uint64_t`? I'd rather not require the `std::` before these basic types. Across all the major compilers we haven't needed it when using `#include <cstddef>` (these are defined without a namespace for C compatibility). If this is required for C++20 module support, then maybe we make `glz::size_t` and `glz::uint64_t` aliases so that the Glaze code doesn't need to namespace these types. This just keeps code cleaner and will reduce a lot off the diffs in this PR.

DockedFerret800:
@stephenberry This is not required at all and we can return to C-types. I initially switched to `std::` types as this is a much cleaner design, because we get these types directly from `import std`. With C-types, we'll have to do `#include <cstddef>` in every file that uses these types, which means the full potential of modules is not utilised. We can return to C-types if you wish.

DockedFerret800:
If the only reason is diffs, we can use `using std::size_t` and similar for other types.

stephenberry:
@DockedFerret800, thanks for the explanation. What do you think about implementing glaze aliases for types like `size_t`? This would avoid `#include <cstddef>` in every file.

```c++
namespace glz
{
   using size_t = std::size_t;
   ...
}
```

DockedFerret800:
@stephenberry Sure, we can do either what you proposed, which would make it available as `glz::size_t` or we can do `using std::size_t;` at the top of the file, which would make `size_t` available globally in that file. This would be available in that module and for module importers if we export it, similar to what we do with `export using sv =` in `glaze.util.string_literal`. Whichever you prefer.

UPD: The example from `glaze.util.string_literal` is not quite correct as it's defined in glz namespace, however I think you got what I mean

stephenberry:
I'm good with either solution, so pick whichever causes the least friction, or just flip a coin :)

DockedFerret800:
Full build was achieved with Clang-22 on Ubuntu WSL. MSVC is currently blocked due to compiler errors. The good thing is that the Insider channel has switched to weekly releases, so fixes should drop soon. There is still work left to do, but this is a good milestone.

DockedFerret800:
Sorry to everyone who gets notification every time I push. I tried stopping packit CI from running using `[skip ci]`, but it appears to ignore it. I'm trying to push as rarely as possible 😅

wilkolbrzym-coder (commented):
# Review — PR #2408: Native C++20 modules conversion

Thank you for this enormous effort — converting Glaze to native modules while preserving header usability is a very exciting direction. I really appreciate the amount of work that went into this.

I don’t think this is ready to merge yet, and I’d like to request a few changes. I’m only flagging things I’m confident about from the PR description, diff, file tree, and CI logs. For anything uncertain, I’ve marked it as such.

---

## 1. CI is currently almost entirely failing

From the checks page: **57 checks are failing, 1 is cancelled, and only 2 are passing.**

That alone means the PR can’t be merged in its current state. I understand many failures come from compilers that don’t yet handle `import std` / C++20 modules well, but from a project-stability perspective we’ll need the normal CI matrix to remain green — possibly by falling back to header-only builds when modules aren’t supported.

I’m 100% sure about the CI numbers; I’m **not** sure what the exact root cause is for each job, but the overall failure rate is not mergeable.

---

## 2. The branch has conflicts with `main`

The PR page says:

> This branch has conflicts that must be resolved

This is factual and blocks any merge until resolved.

---

## 3. `cmake/install-rules.cmake` sets `CMAKE_INSTALL_INCLUDEDIR` to `"modules"`

I can see this in the diff:

```cmake
set(CMAKE_INSTALL_INCLUDEDIR "modules" CACHE PATH "")
```

I’m confident this is present. This is risky because:

- Existing users expect headers under `${CMAKE_INSTALL_INCLUDEDIR}` (normally `include`).
- Changing that to `modules` may break install layouts and `find_package` consumers.
- Headers and module files appear to be installed into the same directory, which could cause collisions or confusing include paths.

I’d suggest keeping standard `include` for headers and using a separate module installation directory instead of overriding the global include dir.

---

## 4. Fuzzing is temporarily disabled

In `cmake/dev-mode.cmake`:

```cmake
#add_subdirectory(fuzzing) # temporary
```

I’m confident this is in the diff. I understand it may be a temporary workaround while modules are being stabilised, but merging a PR with fuzzing disabled would reduce test coverage. It would be great to re-enable fuzzing before merge, or at least mark it as a known TODO that CI enforces.

---

## 5. The modules tree still contains many non-converted `.hpp` files

From the file tree under `modules/glaze`, there are still `.hpp` files, e.g.:

- `eetf/*.hpp`
- `net/*.hpp`
- `yaml/*.hpp`
- `simd/*.hpp`
- some `util/*.hpp`
- `eetf.hpp`
- `yaml.hpp`
- `forward.hpp`

I’m 100% sure these files exist in the PR file tree. This means the conversion is not yet complete, and there are still two source-of-truth areas inside the module tree. That’s not necessarily wrong, but it should be intentional and documented.

---

## 6. Global `using std::size_t;` appears in some files

I see additions like:

```cpp
using std::size_t;
using std::uint64_t;
```

in several benchmark files, e.g. `benchmarks/simple_float_benchmark.cpp`, `benchmarks/skip_ws_benchmark.cpp`, `benchmarks/string_write_benchmark.cpp`.

I’m confident those are present in the diff.

I’m **not sure** whether the same pattern appears in generated headers or module files. If it does, it could leak into user namespaces and cause ambiguity/ODR issues. Even if it’s only in benchmarks right now, I’d prefer using `glz::size_t` / `std::size_t` fully qualified rather than global `using`.

---

## 7. Most tests still appear to be header-based, not module-based

From the file tree, most tests remain `.cpp` files that `#include` Glaze headers. Only a handful of `.ixx` test files are visible, like:

- `tests/json_performance/json_perf_common.ixx`
- `tests/json_test/json_test_shared_types.ixx`
- `tests/lib_test/test_lib.ixx`

I’m **not 100% sure** how these are built or whether CMake treats them as module consumers. But it would be very helpful to have explicit module-based tests that use `import glaze;` and `import std;`, at least for the compilers where modules work.

---

## Summary

This is a really valuable project, and I’m excited about Glaze moving toward modules. But from what I can see in the current PR:

- CI is almost fully red,
- the branch conflicts with `main`,
- install rules look dangerous,
- fuzzing is disabled,
- the module conversion is incomplete,
- module usage is not clearly tested.

I’d love to see this split into smaller, reviewable steps and merged only once the normal CI matrix is green and the install/header-generation story is solid.

Thank you again for all the work — this is a huge undertaking, and I’m looking forward to seeing it progress.

DockedFerret800:
@wilkolbrzym-coder This AI "review" is not necessary here. I am perfectly aware about all these things. This PR is not currently being queued for merge, obviously nobody is going to merge a PR with failing CI and conflicts.

wilkolbrzym-coder:
> @wilkolbrzym-coder This AI "review" is not necessary here. I am perfectly aware about all these things. This PR is not currently being queued for merge, obviously nobody is going to merge a PR with failing CI and conflicts.

@DockedFerret800 The fact that nobody is merging a draft with failing CI and conflicts is not new information — it is exactly why the review exists. A draft is still a technical proposal, and the moment you expose it for review, the structural problems in it are fair game.

What your reply calls “AI review” is a detailed list of blockers that are visible in the diff, file tree, and CI logs. Several of them are not trivial:

- This branch mixes modules with unrelated changes from other PRs.
- Fuzzing is disabled in `cmake/dev-mode.cmake`.
- `CMAKE_INSTALL_INCLUDEDIR` is changed to `modules`, which breaks existing header-only consumers.
- The module tree still contains many plain `.hpp` files.
- Header generation is not verified in CI.
- Module-based consumer tests are effectively missing.

Being aware of those issues is not the same as solving them. Nobody expects a red draft to be merged. What reviewers expect is that when a PR is open for review, it gets reviewed — even when the author already knows the current state is not mergeable.

I will continue the full pass through this draft and post my detailed findings by **7 September, 18:00 UTC**.

cc @stephenberry

DockedFerret800:
@wilkolbrzym-coder We are perfectly aware about all these issues. I appreciate the attempt to help, however this is just spam, so please keep it out of this thread.

wilkolbrzym-coder:
@DockedFerret800 Then keep it in your fork and don’t open a draft in the main repository.

If you cannot handle review on a public PR, that is not a reviewer problem.

You did not refute a single finding. You dismissed the work.

Close it until it is ready, or accept that an open PR gets reviewed.

DockedFerret800:
@wilkolbrzym-coder Nobody asked you to spam in this thread. Your review is a slop output from an AI, that is not necessary here. This PR is open for a reason. You were already asked twice to retain from further spam to this thread - please do so.

PR #2416, opened 2026-03-30
Update deprecated CI dependencies

annihilatorq:
You can read more [here](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)
You will see warnings if you open "Annotations" dropdown in [any glaze workflow run](https://github.com/stephenberry/glaze/actions/runs/23657543310/job/68918793812)

### Fix
This PR resolves this issue by replacing `checkout/v4`, which runs on Node.js 20, with `checkout/v6`, whose runtime uses Node.js 24. `v5` also uses Node.js 24, but I figured that since the opportunity was there and it wouldn’t break anything, why not upgrade to the latest version?

### DIFF
--- a/.github/workflows/armv7.yml
+++ b/.github/workflows/armv7.yml
@@ -22,7 +22,7 @@ jobs:
     name: GCC armv7 (NEON, cross-compiled)
 
     steps:
-    - uses: actions/checkout@v4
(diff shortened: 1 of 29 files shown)


stephenberry:
Thanks for this actions bump!

PR #2433, opened 2026-04-04
fix(yaml): quote strings containing commas in needs_quoting()

psallandre:
`needs_quoting()` only checked the first character against is_yaml_indicator(). Commas appearing later in the string were not checked in the per-character scan, which only looked for `':', '#', '\n', '\r', '\t'`.

Commas are YAML flow indicators and must be quoted in block context plain scalars. Without quoting, strings like "hello, world" are written as unquoted plain scalars, which can cause ambiguous/invalid YAML.

Add `','` to the per-character scan in needs_quoting().

Tested with glaze 7.2.2, `clang++-21 -std=c++23 -I<glaze-include-path>`


stephenberry:
Thanks for this! Merging.

PR #2434, opened 2026-04-04
fix(yaml): custom_t/skip_if support and custom read tail scan

psallandre:
Three related YAML fixes:

1. `write_block_mapping`: honor `meta::skip` and `meta::skip_if` The JSON writer checks meta_has_skip (compile-time) and meta_has_skip_if (runtime) before writing each field. The YAML writer had no such checks — skip/skip_if were silently ignored. Add the same checks to `write_block_mapping()` before the null member check.

2. `write_block_mapping`: write custom_t and custom to<YAML> inline Types with custom to<YAML> specializations and custom_t fell to the catch-all "complex type" branch, which wrote key:\nvalue instead of key: value. Reorder the catch-all so known complex types (containers, objects, maps) go to the newline branch, and everything else writes inline.

3. custom.hpp: use from<> instead of parse<> for reading from<Format, custom_t> called parse<Format>::op() to read the inner value. For YAML, parse<YAML>::op() includes a document-level tail scan that incorrectly flags remaining mapping keys as syntax_error. Use glz::from<Format, T>::op() directly to skip the document-level scanning.

Tested with glaze 7.2.2, clang++-21 -std=c++23 -I<glaze-include-path>
Disclaimer: PR done with Claude Code

```cpp
// glaze_yaml_custom_bug_reproducer.cpp
#include <glaze/glaze.hpp>
#include <glaze/yaml.hpp>
#include <iostream>
#include <ranges>
#include <string>
#include <vector>

struct Param {
    std::string        name;
    std::string        value;
    std::vector<Param> subParams{};

    bool operator==(Param const&) const = default;
};

// Custom serialization: pack name+status+values into a single formatted string
// (This pattern is useful for human-readable config files)
template <>
struct glz::meta<Param> {
    using T = Param;

    static constexpr auto read_fn = [](Param& t, std::string const& str) {
        auto words = str | std::views::split(',') | std::views::transform([](auto&& r) {
                         auto sv = std::string_view(r);
                         // trim
                         auto start = sv.find_first_not_of(' ');
                         auto end   = sv.find_last_not_of(' ');
                         return (start == std::string_view::npos) ? std::string{} : std::string{sv.substr(start, end - start + 1)};
                     }) |
                     std::ranges::to<std::vector>();
        if (words.size() >= 2) {
            t.name  = words[0];
            t.value = words[1];
        }
    };

    // Produces a long string with commas — triggers the bug
    static constexpr auto write_fn = [](auto const& t) -> std::string {
        // Pad fields to fixed widths for readability (produces ~80+ char string)
        char buf[256];
        std::snprintf(buf, sizeof(buf), "%-30s, %15s", t.name.c_str(), t.value.c_str());
        return buf;
    };

    static constexpr auto value = object("paramName", glz::custom<read_fn, write_fn>, "subParams", &T::subParams);
};

int main() {
    Param p{
        .name      = "n1",
        .value     = "v1",
        .subParams = {{
            .name  = "n2",
            .value = "v2",
        }},
    };

    // === JSON roundtrip: works fine ===
    auto json = glz::write_json(p).value_or("error");
    std::cout << "=== JSON output ===\n" << json << "\n\n";

    Param p_json{};
    auto  json_err = glz::read_json(p_json, json);
    std::cout << "JSON read error: " << (bool(json_err) ? "YES" : "none") << "\n";
    std::cout << "JSON roundtrip:  " << (p == p_json ? "OK" : "MISMATCH") << "\n\n";

    // === YAML roundtrip: FAILS ===
    auto yaml = glz::write_yaml(p).value_or("error");
    std::cout << "=== YAML output ===\n" << yaml << "\n";

    Param p_yaml{};
    auto  yaml_err = glz::read_yaml(p_yaml, yaml);
    if (bool(yaml_err))
        std::cout << "YAML read error: " << glz::format_error(yaml_err, yaml) << "\n";
    else
        std::cout << "YAML read error: none\n";
    std::cout << "YAML roundtrip:  " << (p == p_yaml ? "OK" : "MISMATCH") << "\n";

    return bool(yaml_err) ? 1 : 0;
}
```

```cpp
// glaze_yaml_skip_if_bug_reproducer.cpp
#include <glaze/glaze.hpp>
#include <glaze/yaml.hpp>
#include <iostream>
#include <string>
#include <vector>

struct Node {
    std::string       name;
    std::vector<Node> children{};
    bool              operator==(Node const&) const = default;
};

template <>
struct glz::meta<Node> {
    using T = Node;

    // skip_if: omit "children" when the vector is empty
    static bool skip_if(auto const& value, std::string_view key, glz::meta_context const&) {
        if constexpr (requires { value.empty(); })
            return key == "children" && value.empty();
        return false;
    }

    static constexpr auto value = object("name", &T::name, "children", &T::children);
};

int main() {
    Node root{"parent", {{"child1"}, {"child2"}}};

    // JSON: skip_if works — leaf nodes omit empty "children"
    auto json = glz::write_json(root).value_or("error");
    std::cout << "=== JSON (skip_if works) ===\n" << json << "\n\n";
    // Expected: {"name":"parent","children":[{"name":"child1"},{"name":"child2"}]}
    //                                        ^ no "children":[] on leaf nodes

    bool json_has_empty_children = json.find(R"("children":[])") != std::string::npos;

    // YAML: skip_if is IGNORED — leaf nodes still have "children: []"
    auto yaml = glz::write_yaml(root).value_or("error");
    std::cout << "=== YAML (skip_if ignored) ===\n" << yaml << "\n";
    // Actual output includes:
    //   children:
    //     -
    //       name: child1
    //       children:        <-- should be omitted
    //         []             <-- should be omitted

    bool yaml_has_empty_children = yaml.find("children:\n") != std::string::npos && yaml.find("[]") != std::string::npos;

    std::cout << "JSON omits empty children: " << (json_has_empty_children ? "NO (BUG)" : "YES (correct)") << "\n";
    std::cout << "YAML omits empty children: " << (yaml_has_empty_children ? "NO (BUG)" : "YES (correct)") << "\n";

    // Roundtrip check
    Node from_json{};
    glz::read_json(from_json, json);
    std::cout << "JSON roundtrip: " << (root == from_json ? "OK" : "MISMATCH") << "\n";

    Node from_yaml{};
    glz::read_yaml(from_yaml, yaml);
    std::cout << "YAML roundtrip: " << (root == from_yaml ? "OK" : "MISMATCH") << "\n";

    return yaml_has_empty_children ? 1 : 0;
}
```

### DIFF
--- a/tests/yaml_test/yaml_test.cpp
+++ b/tests/yaml_test/yaml_test.cpp
@@ -185,6 +185,88 @@ struct glz::meta<yaml_custom_read_struct>
    static constexpr auto value = object("value", custom<&T::read_value, &T::value>);
 };
 
+// Struct with custom lambda-based read/write for YAML custom_t testing
+struct yaml_custom_lambda_struct
+{
+   std::string label{};
+   int score{};
+
+   bool operator==(const yaml_custom_lambda_struct&) const = default;
+};
+
+template <>
+struct glz::meta<yaml_custom_lambda_struct>
+{
+   using T = yaml_custom_lambda_struct;
+
+   static constexpr auto read_fn = [](T& t, const std::string& str) {
+      auto pos = str.find('|');
+      if (pos != std::string::npos) {
+         t.label = str.substr(0, pos);
+         t.score = std::stoi(str.substr(pos + 1));
+      }
+   };
+
+   static constexpr auto write_fn = [](const T& t) -> std::string { return t.label + "|" + std::to_string(t.score); };
+
+   static constexpr auto value = object("data", glz::custom<read_fn, write_fn>);
+};
+
+// Struct with custom field alongside regular fields
+struct yaml_custom_mixed_struct
+{
+   std::string name{};
+   yaml_custom_lambda_struct item{};
+   int count{};
+
+   bool operator==(const yaml_custom_mixed_struct&) const = default;
+};
+
+template <>
+struct glz::meta<yaml_custom_mixed_struct>
+{
+   using T = yaml_custom_mixed_struct;
+   static constexpr auto value = object("name", &T::name, "item", &T::item, "count", &T::count);
+};
+
+// Struct with compile-time skip for YAML
+struct yaml_skip_struct
+{
+   std::string id{};
+   std::string secret{};
+   int count{};
+};
+
+template <>
+struct glz::meta<yaml_skip_struct>
+{
+   static constexpr bool skip(const std::string_view key, const glz::meta_context&) { return key == "secret"; }
+};
+
+// Struct with runtime skip_if for YAML
+struct yaml_skip_if_struct
+{
+   std::string name{};
+   int age{};
+   std::string city{};
+};
+
+template <>
+struct glz::meta<yaml_skip_if_struct>
+{
+   template <class V>
+   static constexpr bool skip_if(V&& value, std::string_view key, const glz::meta_context&)
+   {
+      using D = std::decay_t<V>;
+      if constexpr (std::same_as<D, int>) {
+         return key == "age" && value == 0;
+      }
+      else {
+         return false;
+      }
+   }
+};
+
 suite yaml_write_tests = [] {
    "write_simple_struct"_test = [] {
       simple_struct obj{42, 3.14, "test"};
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for reporting these issues and the PR fixes. I'll review them ASAP. YAML support is still very new, so thanks for helping put it through its paces.

stephenberry:
@psallandre, thanks for these improvements! I added unit tests and I'm merging.

PR #2462, opened 2026-04-10
Update MSVC version in README to the correct one used in CI

### LINKED ISSUE #2454: write_toml() does not compile with MSVC
Environment:
```
Microsoft Visual Studio Community 2022 (64-bit) - Current
Version 17.14.29 (March 2026)
Compiler Version 19.44.35225 
```

Error:
```
glaze\toml\write.hpp(1104,64): error C2131: expression did not evaluate to a constant
```

Minimal example:
```
#include <glaze/toml.hpp>
#include <string>

struct Settings
{
    int value{42};
    std::string name{"test"};
};

template <>
struct glz::meta<Settings>
{
    using T = Settings;
    static constexpr auto value = object("value", &T::value, "name", &T::name);
};

int main()
{
    Settings s;
    std::string buffer;
    auto ec = glz::write_toml(s, buffer);
    return ec ? 1 : 0;
}
```

I've also attached the minimal CMake/vcpkg project.

[glaze_bug.zip](https://github.com/user-attachments/files/26619742/glaze_bug.zip)

DockedFerret800:
Closes #2454.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ The buffer refills automatically during parsing, enabling reading of arbitrarily
 - Tested for both 64bit and 32bit
 - Supports both little-endian and big-endian systems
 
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks!

PR #2510, opened 2026-04-16
CBOR std::chrono support

### LINKED ISSUE #2507: Add `std::chrono` support for CBOR
CBOR has tags to represent dates so it would be nice if the library would support conversion with `std::chrono` and the `glaze::epoch_*` types.

https://stephenberry.github.io/glaze/chrono/?h=epoch#stdchrono-support

stephenberry:
Closes #2507.

## Summary
Adds CBOR read/write for `std::chrono::duration`, `std::chrono::system_clock::time_point`, `std::chrono::steady_clock::time_point`, and `glz::epoch_time<Duration>`.

## Wire format

| Type | Encoding | RFC |
|------|----------|-----|
| `std::chrono::duration` | bare integer in the duration's native units | — |
| `std::chrono::system_clock::time_point` | tag 0 + RFC 3339 text string | 8949 §3.4.1 |
| `glz::epoch_seconds` (Period ≥ 1s) | tag 1 + integer seconds | 8949 §3.4.2 |
| `glz::epoch_millis` / `epoch_micros` / `epoch_nanos` | tag 1 + float64 seconds | 8949 §3.4.2 |
| `std::chrono::steady_clock::time_point` | bare integer count (no tag) | — |

RFC 8949 §3.4.2 restricts tag 1 content to major types 0, 1, and 7 (int or float); nested tags are forbidden. `epoch_millis` and `epoch_micros` round-trip losslessly through float64 for any realistic modern epoch. `epoch_nanos` degrades to ~microsecond precision for current timestamps (2^53 ns ≈ ~year 2255 of nanosecond-count since 1970). Applications needing lossless nanosecond wire precision can use a bare integer type or build an RFC 9581 tag 1001 wrapper.

## Reader behavior

- `system_clock::time_point` accepts tag 0 + tstr, tag 1 + int/float, or a bare tstr (no tag, lenient for producers that omit tag 0). Bare numbers and other shapes are rejected.
- `epoch_time<Duration>` accepts tag 1 + int/float or a bare int/float. Nested tags are rejected.
- Tag 0 with non-tstr content and tag 1 with nested tags (e.g. tag 4 decimal fractions) are rejected per §3.4.1 / §3.4.2.
- NaN and Infinity on tag-1 floats are rejected.
- Integer and float seconds are bounds-checked against `system_clock::duration` so a large wire value cannot overflow `duration_cast`.
- Writer rejects timestamps outside the RFC 3339 year range (0000–9999) rather than emitting corrupt digits.

## Implementation notes

- A shared RFC 3339 / ISO 8601 parser now lives in `include/glaze/core/chrono.hpp` (`glz::chrono_detail::parse_iso8601`). Both the CBOR tag 0 / bare-tstr reader and the JSON `system_clock::time_point` reader route through it — this collapses ~130 duplicate lines previously inlined in `json/read.hpp`. Any future backend (TOML, YAML, EETF…) can reuse it.
- The CBOR float writer already picks the smallest exact representation (`can_encode_half` → `can_encode_float` → float64), so small tag-1 values use narrower floats automatically.
- `system_clock::time_point` (tag 0 strings) and `epoch_time<Duration>` (tag 1 numbers) do not cross-read: the wire form is type-selected on write.

## Tests
- `tests/cbor_test/cbor_test.cpp` — 222 total; 36 new `chrono_tests` / `chrono_struct_tests` covering roundtrips, RFC 8949 canonical bytes (`0xC1 0x1A 0x51 0x4B 0x67 0xB0`), wire-format assertions (tag 0 bytes, tag 1 + float64 bytes), tag-4 rejection, bare-tstr / bare-number interop, NaN/Infinity rejection, tz offset parsing (`+05:30`, `-08:00`), system_clock at microsecond/nanosecond precision, integer-overflow and float-overflow rejection, raw-bytes tag-1 integer/nint/float16/float32 read paths, out-of-range-year rejection, and a pinned `epoch_nanos` precision-loss characterization test.
- `cbor_test`: 222/222 tests, 876/876 asserts.
- `json_test`: 651/651 tests, 57069/57069 asserts passing on the shared parser after the JSON reader was routed through `chrono_detail::parse_iso8601`.

### DIFF
--- a/docs/cbor.md
+++ b/docs/cbor.md
@@ -229,3 +229,79 @@ Following RFC 8949 preferred serialization, Glaze automatically uses the smalles
 - Double-precision (64-bit) as fallback
 
 This reduces message size while preserving exact values.
+
+## Date and Time (std::chrono)
+
+Glaze serializes `std::chrono` types to CBOR using the standard semantic tags defined by RFC 8949. See [chrono documentation](./chrono.md) for the shared concepts; this section covers the CBOR-specific wire format.
+
+### Wire format
+
+| Type | Encoding | Source |
+|------|----------|--------|
+| `std::chrono::duration<Rep, Period>` | Bare integer count in the duration's native units | — |
+| `std::chrono::system_clock::time_point` | Tag 0 + text string (RFC 3339 date/time) | RFC 8949 §3.4.1 |
+| `glz::epoch_seconds` (Period ≥ 1s) | Tag 1 + integer seconds since epoch | RFC 8949 §3.4.2 |
+| `glz::epoch_millis` / `epoch_micros` / `epoch_nanos` | Tag 1 + float64 seconds since epoch | RFC 8949 §3.4.2 |
+| `std::chrono::steady_clock::time_point` | Bare integer count (no tag) | — |
+
+RFC 8949 §3.4.2 restricts tag 1 content to integers (major types 0/1) or floating-point numbers (major type 7, additional info 25/26/27). Nested tags — including tag 4 decimal fractions — are explicitly forbidden as tag 1 content, so sub-second precision is carried via float64.
+
+### Example
+
+```c++
+#include "glaze/cbor.hpp"
+#include <chrono>
+
+struct Event {
+    std::chrono::system_clock::time_point timestamp;  // tag 0 + RFC 3339 string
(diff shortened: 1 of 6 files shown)


cacharle:
Looks ok

PR #2513, opened 2026-04-16
Add SQLite JSONB binary format support

### LINKED ISSUE #2306: Sqlite JSONB format
Sqlite has its own [binary JSON format](https://sqlite.org/jsonb.html). It is standardized but seen as an internal format. I am currently create JSON with Glaze which is then converted by Sqlite to their JSONB. Avoiding that conversion could be helpful for other Sqlite users.

Would there be any interest to support Sqlite JSONB? Is it hard to adding support?

stephenberry:
Closes #2306

## Summary

Implements read, write, and JSON-conversion support for SQLite's [JSONB](https://sqlite.org/jsonb.html) binary JSON format. Lets applications serialize C++ values straight into SQLite JSONB blobs without first producing JSON text and paying SQLite's JSON→JSONB conversion cost on every insert.

## API

```c++
#include "glaze/jsonb.hpp"

// Write
glz::write_jsonb(value, buffer);                    // error_ctx
auto blob = glz::write_jsonb(value);                // expected<string, error_ctx>
glz::write_file_jsonb(value, "out.jsonb", scratch); // error_ctx

// Read
glz::read_jsonb(value, buffer);                     // error_ctx
auto value = glz::read_jsonb<T>(buffer);            // expected<T, error_ctx>
glz::read_file_jsonb(value, "in.jsonb", scratch);   // error_ctx

// Convert (debugging / strict JSON output)
auto json = glz::jsonb_to_json(buffer);             // expected<string, error_ctx>

// Compose with other opts
constexpr auto opts = glz::set_jsonb<glz::opts{.error_on_unknown_keys = false}>();
glz::read<opts>(value, buffer);
```

`#include "glaze/exceptions/jsonb_exceptions.hpp"` for the throwing equivalents under `glz::ex::*`.

## Coverage

All 13 JSONB element types: `NULL`, `TRUE`, `FALSE`, `INT`, `INT5`, `FLOAT`, `FLOAT5`, `TEXT`, `TEXTJ`, `TEXT5`, `TEXTRAW`, `ARRAY`, `OBJECT`. Reserved codes 13–15 are rejected.

Writable types: all integers, floats (incl. `NaN`/`Inf` → `FLOAT5`), `std::string`/`std::string_view`, `std::vector`, `std::array`, `std::deque`, tuples, `std::map`/`std::unordered_map` (string keys), `std::pair`, reflected structs, `glaze_array`, `std::optional`, `std::unique_ptr`, `std::shared_ptr`, `std::variant`, `glz::generic` (schemaless round-trip), enums.

Supported `glz::opts`: `error_on_unknown_keys`, `error_on_missing_keys`, `skip_null_members`, `skip_default_members`, `error_on_const_read`, `max_string_length`, `write_function_pointers`, `partial_read`.

## Design

- **Container headers**: writer reserves 9 bytes (`u64_follows` form) up front, then patches the actual size after children are written. Spec-compliant single-pass with no `memmove`. Trade-off: a few wasted bytes per small container. (`json_valid(blob, 4)` accepts; `json_valid(blob, 8)` requires canonical form and does not.)
- **Scalar headers**: minimal 1-or-2-byte form.
- **String type pre-scan**: writer scans each string and picks `TEXT` (no escaping needed) or `TEXTRAW` (control char, `"`, or `\` present). `TEXT` lets SQLite's `json()` and Glaze's converter byte-copy without scanning, so the common case of clean ASCII / UTF-8 round-trips with zero per-byte work on read.
- **Variants**: serialize the active alternative directly (no `[index, value]` wrapper), JSON-style. Reader deduces from the JSONB type code. JSONB's INT vs FLOAT distinction lets `variant<int, double>` auto-deduce here even though Glaze's JSON cannot.
- **Tagged variants**: when `tag_v<T>` is set and the alternative is a reflected struct, the writer emits `{tag: id, ...inlined fields}`. Reader expects the tag as the first key and resolves via `variant_id_to_index`. Both string and integer ids supported.
- **Reflected-object reads**: `hash_info<T>` + `decode_hash_with_size` for O(1) key lookup, with TEXTJ/TEXT5 keys decoded into a scratch string before hashing.
- **DoS resistance**: container readers and the converter cap recursion at `max_recursive_depth_limit` (256). Integer parses bounds-check against the target type rather than silently truncating on overflow.

## SQLite interop validation

`tests/jsonb_sqlite_test/` links against `libsqlite3` (≥ 3.45, required for JSONB) and validates empirically in both directions:

- **Glaze → SQLite**: write JSONB, `INSERT` as BLOB, `SELECT json(blob)` and compare to expected JSON.
- **SQLite → Glaze**: `SELECT jsonb(json_text)` to produce a reference blob, read it back with Glaze.
- **Byte-level equivalence**: scalars (`null`, `true`, `false`, `42`, `-12345`, ...) are byte-identical to SQLite's own encoder output.
- **Path queries**: `json_extract`, `json_array_length`, `json_type`, and `WHERE` filters all work against Glaze blobs.

The CMake target uses `find_package(SQLite3 3.45 QUIET)` and is silently skipped when SQLite isn't available or the build is cross-compiled / 32-bit. Apple's system SQLite lacks the JSON1 extension — a `sqlite_has_jsonb()` runtime probe gates the suite bodies on macOS.

## Fuzzing

`fuzzing/jsonb_reflection.cpp` drives three reader entry points per input: typed `read_jsonb<my_struct>`, `jsonb_to_json` converter, and schemaless `read_jsonb<glz::generic>`. Builds with `-fsanitize=address,undefined` under clang and falls back to the file-replay harness on platforms without libfuzzer.

## Documentation

`docs/jsonb.md` covers the type table, options, container encoding strategy, string type selection, variants (auto-deduce and tagged), exceptions interface, and SQLite interop notes. `mkdocs.yml`, `README.md`, and `docs/binary.md` updated to reference it.

## Test plan

- [x] `ctest -R jsonb_test` — 141 tests, 532 assertions
- [x] `ctest -R jsonb_sqlite_test` — 36 tests, 139 assertions
- [x] Full suite (`ctest -j8`) — no regressions
- [x] Clang (C++23) on macOS
- [x] CI: Linux / GCC / Clang / MSVC / P2996 reflection
- [x] Fuzzer survives 200 random inputs locally under ASan/UBSan
- [x] SQLite interop target correctly skipped on cross-compile / 32-bit / Apple-system-SQLite builds


marbub:
I am not an expert in this field, but is some of the code doing a std::in_range<T>(value)?

stephenberry:
> I am not an expert in this field, but is some of the code doing a std::in_range(value)?

Thanks for noticing this. I updated the code to use std::in_range.

PR #2527, opened 2026-04-23
Compilation time optimization and cleanup

### LINKED ISSUE #2525: Forward declaration headers to improve compilation speed for projects that use glaze
### Discussed in https://github.com/stephenberry/glaze/discussions/2524

<div type='discussions-op-text'>

<sup>Originally posted by **cout** April 22, 2026</sup>
On my development machine, just including `<glaze/glaze.hpp>` adds 6.8 seconds to my build for each translation unit that includes that file.  This is using clang with -Og, and all I am doing is including the header.

I have many types for which I want to specialize `glz::meta`, `glz::from`, or `glz::to`.  For non-struct types, these are simple lambda functions or type aliases, so forward declaring them in a header makes sense, and I don't have to include any glaze headers:

```
namespace glz {
  template <typename T>
  struct meta;

  template <uint32_t Format, typename T>
  struct to;

  template <uint32_t Format, typename T>
  struct from;

  template <uint32_t Format>
  struct serialize;

  template <uint32_t Format>
  struct parse;

  template <class T, class From, class To>
  struct custom_t;
}
```

(side note: it might be a good idea to add a type alias for that uint32_t format type)

Since I have a few classes where I want to use `glz::custom`, I've written my own wrapper that does not require including any of glaze:

```
template <auto From, auto To>
constexpr auto my_custom = []() noexcept {
  return [](auto && v) { return glz::custom_t { v, From, To }; };
}();
```

(no deduction guide needed, since the lambda does not get evaluated until I call `read_json` or `write_json`, and the translation unit where I do that includes all the important bits of glaze)

For struct types, since we are not yet using reflection, I also wanted to use `glz::object` without including `<glaze/core/common.hpp>`, which adds about 4 seconds to the compile time.  But this does not work -- I need `glz::detail::Object` or a specialization of it.  I started going down that rabbit hole anyway, and made a tagged specialization of `Object`, then I realized I need to pull in `glaze/tuple/tuple.hpp` as well.  I've decided this is probably more brittle than I want.

Suggestion:
* Provide headers that forward-declare any types that might get specialized
* Separate functions and types used to specialize `glz::meta` from those used to interpret `glz::meta`.
* Be aware of which standard library headers are more expensive to include (iostream, fstream, chrono, filesystem, and complex are some of the more egregious, followed by ranges and iterator

For discussion:
* What parts of glaze are most used when specializing meta/to/from?
* What parts of glaze are most conducive to splitting out into separate header files?
* What parts of glaze are included by default, but include expensive standard library headers, and could be split out? (core/chrono.hpp comes to mind)
* Are there better ways to define meta alongside my classes without pulling in all of glaze core?</div>

stephenberry:
Compile-time and include-surface cleanup, landing as a sequence of independent commits.

## New: `<glaze/forward.hpp>`
Lightweight header with forward declarations for the customization points (`glz::meta`, `glz::to`, `glz::from`, `glz::to_partial`, `glz::skip_value`, `glz::parse`, `glz::serialize`, `glz::serialize_partial`), format id constants, and `glz::custom_t`. Only depends on `<cstdint>`. Users can specialize these in widely-included headers without pulling in full Glaze. Addresses #2525.

## `glaze/glaze.hpp` aggregate trimmed
Uncommon features removed from the top-level aggregate; users who want them include the individual headers directly (same convention CBOR already followed):
- `glaze/msgpack.hpp` (`docs/msgpack.md` updated accordingly)
- `glaze/record/recorder.hpp` (`docs/recorder.md` updated accordingly)

## Core include diet
- `<atomic>` / `<bitset>` / `<complex>` in `core/common.hpp` are now gated behind `#if GLZ_REFLECTION26` — the `specified<>` specializations that need them are the only users and already live inside that block.
- `core/chrono.hpp` dropped from `core/common.hpp`; moved to the direct users `cbor/read.hpp` and `cbor/write.hpp`.
- `glaze/file/write_directory.hpp` now explicitly includes `glaze/json/write.hpp` (previously worked by transitive luck).
- Dead `<any>` include removed from `json/json_ptr.hpp`.

## Internal simplifications
- `unique_per_length_info` in the reflection hash pipeline drops `constexpr std::vector` + `std::ranges::sort` for a copy of the input `std::array` plus a hand-rolled insertion sort — fewer template instantiations per reflected type.
- `glz::boxed<T>` (internal to `struct glz::schema`) now manages a raw `T*` directly instead of wrapping `std::unique_ptr<T>`. Same public API, same semantics; skips the per-T `unique_ptr` + `default_delete` instantiations that every TU including `schema.hpp` otherwise pays.

Full build and all test suites pass (json, yaml, beve, bson, cbor, jsonschema).


stephenberry:
@cout, would you mind looking over this PR concerning Glaze compilation time optimizations? Particularly `<glaze/forward.hpp>` for forward declarations.

cout:
I have just returned from vacation and will look over this PR this week.

stephenberry:
@cout, I decided this PR is in a good place to merge into main, so I'm merging it in. Just let me know if there are additional changes to make to `glaze/forward.hpp` to assist your needs or help compilation speeds. I'm going to do more passes to improve compile times, but this is a first step.

cout:
Sounds good! I did try looking at the code last week, but github was taking too long (i.e. over an hour) to load the diff.  It's fast now.  Hopefully I will have some feedback for you today.

makekryl:
One more suggestion: sometimes it's needed to use `glz::generic` in internal structures, but it pulls the entire json codec, polluting the TUs that don't actually perform encoding/decoding - is it also possible to split up `glaze/json/generic.hpp` into two headers, with one remaining "pure" without `glaze/json/*` includes?

UPD: I prepared a PR: https://github.com/stephenberry/glaze/pull/2560

cout:
Finally got a chance to take a look at this set of changes in detail.

`glaze/forward.hpp` looks good.  There are a few other types that we use that can be forward declared:
* `glz::tuple_size` and `glz::tuple_element` - we have a tuple-like type that specializes these so glaze will see it as a tuple (we also specialize the same functions in the `std` namespace)
* `glz::raw_json` - we use this as a default in a generic lambda (so the definition is not needed until the lambda is instantiated, but the declaration is needed for this to work):
```cpp
static constexpr auto value = our_glz_wrappers::custom<
  [](T & x, double v) { x = T(v); },
  []<typename raw_json = glz::raw_json>(T x) { return raw_json(our_custom_type_to_string(x); }
>;
```
* `glz::error_code` - I don't remember why we wanted to forward-declare this.  I don't think we use it without the definition.
* `glz::custom` - this can't be "forward" declared, but the definition is so trivial (and useful), I think it belongs in `forward.hpp`

cout:
For comparison: I had previously measured compile time for a TU that includes each of the headers below with glaze 7.3.3, and I measured them again with 7.7.1 (same machine, same compiler):

| header |7.3.3 | 7.7.1 |
| ------ | ----- | ----- |
| `glaze/glaze.hpp` | 6.336s | 6.173s |
| `glaze/core/meta.hpp` | 1.463s | 1.315s |
| `glaze/core/wrappers.hpp` | 1.277s | 1.081s |
| `glaze/core/common.hpp` | 4.003s | 3.718s |
| `glaze/core/opts.hpp` | 1.239s | 1.036s |
| `glaze/core/context.hpp` | 0.926s | 0.802s |
| `glaze/core/optimization_level.hpp` | 0.064s | 0.038s |
| `glaze/core/traits.hpp` | 0.099s | 0.070s |
| `glaze/util/inline.hpp` | 0.055s | 0.033s |
| `glaze/util/type_traits.hpp` | 0.680s | 0.578s |
| `glaze/containers/inplace_vector.hpp` | 1.591s | 1.459s |
| `glaze/containers/flat_map.hpp` | 1.314s | 1.149s |
| `glaze/containers/ordered_map.hpp` | 1.336s | 1.314s |
| `glaze/containers/ordered_small_map.hpp` | 1.298s | 1.148s |
| `glaze/tuplet/tuple.hpp` | 0.406s | 0.328s |

PR #2532, opened 2026-04-27
std_tuple_protocol

stephenberry:
#2530

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -2096,7 +2096,7 @@ namespace glz
    };
 
    template <class T>
-      requires(tuple_t<T> || is_std_tuple<T>)
+      requires(tuple_t<T> || std_tuple_protocol<T>)
    struct from<BEVE, T> final
    {
       template <auto Opts>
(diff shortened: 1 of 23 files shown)


wilkolbrzym-coder:
Quick review notes – PR #2532

**Blocking: branch has conflicts and 3 failing checks** (gcc-p2996-reflection, msvc, p2996-reflection). Rebase and fix CI before merge.

The new **`std_tuple_protocol`** concept is the right idea, but it has a regression: **`std::tuple<>`** no longer satisfies it, because **`std::tuple_element<0, T>`** is ill-formed for empty tuples. Previously **`is_std_tuple`** covered this case.

**`has_adl_tuple_get`** checks only rvalue **`get<0>(T&&)`**, so types with only lvalue/const-lvalue **`get`** are rejected even though they would serialize fine.

The unqualified **`get<I>(value)`** inside namespace **`glz`** may resolve to **`glz::get`** instead of ADL; this is a likely MSVC/P2996 failure source. Consider an explicit ADL barrier.

Several read/write paths put **`use_std_protocol`** before **`glaze_array_t`** in **`if constexpr`** branches. If a type satisfies both, **`N`** is taken from reflection but elements are read via **`get<I>`**. Prefer **`glaze_array_t`** first.

**`toml/write.hpp`** was updated, but **`toml/read.hpp`** was not — verify TOML read supports the new protocol or document the asymmetry.

Tests cover JSON, BEVE, CBOR, MSGPACK, but not YAML, TOML, JSONB, EETF, empty tuples, const-only **`get`**, or mixed **`glaze_array_t`**/**`std_tuple_protocol`**.

Overall approach LGTM once the empty-tuple regression, lookup ambiguity, CI failures, and branch ordering are fixed.

PR #2533, opened 2026-04-28
Fix variant deduction for vector<pair<...>> as JSON object (#2529)

### LINKED ISSUE #2529: no_matching_variant_type when the json could be parsed as one of the variant types
I am seeing unexpected behavior in the following program:
```
#include <glaze/exceptions/json_exceptions.hpp>

#include <vector>
#include <string_view>
#include <variant>

int main() {
  using map_type = std::vector<std::pair<std::string_view, std::string_view>>;

  map_type m;
  glz::ex::read_json(m, R"({ "foo": "bar" })"); // OK

  std::variant<std::string_view, std::nullptr_t, map_type> v;
  glz::ex::read_json(v, R"({ "foo": "bar" })"); // no_matching_variant_type
}
```

Since I can read the json object into a vector of pairs, I expected to be able to read the json object into a variant where one of the variant types is a vector of pairs.

Is this behavior intentional or a bug?

stephenberry:
## Summary

- Fixes #2529: `std::variant` containing a `vector<pair<...>>` alternative failed with `no_matching_variant_type` on `{...}` JSON, even though the same type parses fine as a JSON object when read directly.
- Variant deduction relied on the `json_object` concept, which only recognized map-like types via `readable_map_t`/`writable_map_t`. Those concepts require a `key_type` member (gated by `map_subscriptable`), which `vector<pair<...>>` lacks. The direct parser accepts such ranges via a separate concatenate-mode overload (default), so the variant path was inconsistent with direct reads.
- Whether a range-of-pairs is parsed as `{...}` or `[{...}]` depends on the per-call `concatenate` option (default `true`), so the deduction has to be option-aware rather than purely type-level.

## Approach

- Added a new concept `json_pair_range_object` in `json/json_concepts.hpp` for `range<T> && pair_t<range_value_t<T>>` alternatives that are not already `json_object`. `json_object` itself is unchanged.
- Added a matching variant trait `variant_concat_object_type` / `is_variant_concat_object` in `json/read.hpp`.
- The variant `case '{'` path now consults `check_concatenate(Opts)`; when true (default), pair-range alternatives count as object candidates, including in the `n_object == 1` fast path. When `concatenate=false`, they do not, so the variant returns the cleaner `no_matching_variant_type` error instead of dispatching to a parser that would fail with `expected_bracket`.

## Test plan

- [x] New regression suite in `tests/json_test/json_variant_support_test.cpp` covering the issue's repro, deduction across `null`/`string`/`object` alternatives, and a `concatenate=false` case asserting `no_matching_variant_type`.
- [x] All 22 JSON ctest binaries pass locally (`json_variant_support_test`: 80 tests, 395 asserts).

### DIFF
--- a/include/glaze/json/json_concepts.hpp
+++ b/include/glaze/json/json_concepts.hpp
@@ -8,8 +8,13 @@
 namespace glz
 {
    template <class T>
-   concept json_object = glaze_object_t<T> || reflectable<T> || writable_map_t<T> || readable_map_t<T> ||
-                         (range<T> && pair_t<range_value_t<T>>);
+   concept json_object = glaze_object_t<T> || reflectable<T> || writable_map_t<T> || readable_map_t<T>;
+
+   // Range-of-pairs types are parsed as JSON objects only when the `concatenate` option is true (default).
+   // The variant parser uses this separately so deduction can be made option-aware without polluting
+   // the type-level `json_object` concept.
+   template <class T>
+   concept json_pair_range_object = !json_object<T> && range<T> && pair_t<range_value_t<T>>;
 
    template <class T>
    concept json_array = array_t<T>;
--- a/include/glaze/json/json_concepts.hpp
+++ b/include/glaze/json/json_concepts.hpp
@@ -8,7 +8,8 @@
 namespace glz
 {
    template <class T>
-   concept json_object = glaze_object_t<T> || reflectable<T> || writable_map_t<T> || readable_map_t<T>;
+   concept json_object = glaze_object_t<T> || reflectable<T> || writable_map_t<T> || readable_map_t<T> ||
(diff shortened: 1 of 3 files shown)


stephenberry:
Updated to be option-aware after a good catch in review:

- Reverted the eager change to `json_object`. `vector<pair<...>>` is no longer unconditionally classified as a JSON object at the type level.
- Added a separate concept `json_pair_range_object` for range-of-pairs alternatives.
- The variant `case '{'` branch consults this concept only when `check_concatenate(Opts)` is true (the default). With `concatenate=false`, the user has opted into array-shaped pair-ranges, so the variant correctly returns `no_matching_variant_type` for `{...}` input instead of dispatching to a parser that will then fail with `expected_bracket`.
- Added a regression test exercising `concatenate=false` to lock the option-aware behavior in.

PR #2535, opened 2026-04-29
Fix ordered_small_map parse error under namespace pollution (#2534)

### LINKED ISSUE #2534: Parse error using clang with`ordered_small_map.hpp` when `std::hash` is hoisted into global scope
Discovered while bumping the glaze version in one of my projects

`include/glaze/containers/ordered_small_map.hpp` (added in 7.x) fails to parse under Clang when `std::hash` has been brought into the global namespace by an earlier translation-unit header.  This causes a clash with the `std::hash` class template now visible at global scope.

Probably a stray `using namespace std;` at file scope in some upstream header I don't control.

### Minimal repro (no glaze)

```cpp
#include <functional>   // brings std::hash into ::std
using namespace std;    // hoists std::hash into ::

template <class T>
struct M {
    struct E { unsigned hash; };
    static const E* f(const E* p, unsigned target) {
        if (p->hash < target) return p;
        return nullptr;
    }
};
int main() {}
```

```
clang++ -std=c++23 -c repro.cpp
```

```
repro.cpp:9:29: error: expected '>'
    9 |         if (p->hash < target)
      |                             ^
repro.cpp:9:21: note: to match this '<'
```

Confirmed on Apple Clang (libc++) and Clang 22.1.1 + libstdc++. 

Stdlib-agnostic - any standard library brings `std::hash` into `::` once `using namespace std;` is in scope.

### Why this hits glaze

Clang treats `obj.hash <` as the start of a template-id and never finds the closing `>`.  See some examples here; https://github.com/stephenberry/glaze/blob/v7.5.0/include/glaze/containers/ordered_small_map.hpp#L350-L351

Arguably this is an issue with Clang being more eager than required, but this is real-world code that doesn't `using namespace std;` itself, yet still gets broken by transitively-included headers that do. 

I've patched locally with some extra parenth, and my build is green.  I can make a PR of the fix, which would defensively avoid the problem, although aye, it's arguably clang's issue.

Wheest:
Fix ordered_small_map parse error when std::hash is hoisted to global scope (fixes #2534)


stephenberry (commented):
Thanks for the fix. I'll fold the regression test into the existing `ordered_small_map_test` executable as a second source file rather than creating a new one.

stephenberry:
I decided to just rename `hash` to `key_hash` so that we don't need another test executable for this tiny fix.

Wheest:
great, thanks!

PR #2550, opened 2026-05-07
Fix streaming and WebSocket route lookup with query strings (#2549)

### LINKED ISSUE #2549: http_server::stream_get does not properly handle query parameters
Hello,

I am trying to use glaze's http_server (as a replacement for Pistache) and want to use its streaming capabilities.

However, I have noticed, that any request with query parameters (`/path?a=b&c=d`) does not arrive at streaming handlers registered with `http_server::stream` (and similar functions).

Overall it would be nice if streaming routes worked the same as normal route and could also be registered for an `http_router` rather than just the `http_server` and ideally also support `_async`.

stephenberry:
## Summary

Fixes #2549.

`http_server::stream_*` handlers (and the WebSocket upgrade path) failed to match any request that included a query string. A request to `/path?a=b&c=d` would miss the streaming-handler map and fall through to the regular router, returning 404.

The lookup used the raw request target (path + query string) as the map key, while routes are registered by path only. The fix parses path and query from the target *before* the lookup, then keys the lookup by `request.path`.

## Root cause

In `include/glaze/net/http_server.hpp`:

- `process_full_request_with_conn` looked up streaming handlers via `streaming_handlers_.find(conn->request_.target)`.
- `handle_websocket_upgrade_with_conn` looked up WebSocket handlers via `websocket_handlers_.find(conn->request_.target)`.

`conn->request_.target` is the raw target including any `?query=string`, so the find always missed for requests with parameters. The regular router `match()` already handled this correctly (it called `split_target` internally), which is why non-streaming routes worked.

## Fix

Move the path/query parsing above the streaming-handler lookup and use `request.path` as the key. Apply the same change to the WebSocket upgrade lookup. As a side effect, streaming and WebSocket handlers now see `request.path` and `request.query` populated the same way regular handlers do, so handlers can read query parameters directly.

## Test plan

- [x] Add 6 streaming regression tests in `tests/networking_tests/http_chunked_test/http_chunked_test.cpp` covering: the `?a=b&c=d` form from the issue, no-query baseline, trailing `?` with no params, single param, URL-encoded value (`%20` decoded into the parsed map), and `stream_post`. Plus a 7th test asserting wrong-method-with-query still returns 404.
- [x] Add 1 WebSocket regression test in `tests/networking_tests/websocket_test/websocket_client_test.cpp` that connects to `ws://.../ws?token=abc&user=alice`, asserts the upgrade reaches `on_open`, and inspects the captured `request.path`, `request.target`, and parsed `request.query`.
- [x] Verified each new test fails on `main` without the production fix and passes with it.
- [x] Full `http_chunked_test` suite passes (50/50).
- [x] Full `websocket_client_test` suite passes (30/30, ran 3x for stability).


JnCrMx:
I think it would be nice (if possible) if instead of storing the streaming and websocket routes separately, they would be stored just like any other route in the http_router.

In my opinion they should both behave like any other route (i.e. path parameters should ideally also work).

I don't know the internal architecture of Glaze enough to know if it makes sense, but my idea would be to store normal routes, streaming routes and websocket routes in a std::variant and to only handle them differently after root_router.match(...).

JnCrMx:
Ideally I would like to be able to do:
```cpp
glz::http_router router;
router.get("/my/path/:id/normal", [](const glz::request& request, glz::response& response){...});
router.stream_get("/my/path/:id/stream", [](const glz::request& request, glz::streaming_response& response){...});

glz::http_server server{...};
server.mount("/", router);
```

stephenberry:
Thanks for the feedback. I like that syntax. I'm not sure how long it will take to nail down this design, but I'll work on it when I'm able.

stephenberry:
@JnCrMx, I'm merging this targeted fix for your posted issue and then I'll look into supporting your requested syntax in another PR.

PR #2561, opened 2026-05-11
toml: fix inline tables parsing, support for nullables

makekryl:
Seems like it was forgotten to exit from the inline table parsing context, resulting in "unknown key" errors afterwards

Relevant example (parser errored on `b`):
```toml
a = 1
inline_data = { key1 = "value1", key2 = 100 }
b = 2
```

### DIFF
--- a/include/glaze/toml/read.hpp
+++ b/include/glaze/toml/read.hpp
@@ -2859,7 +2859,7 @@ namespace glz
       }
    };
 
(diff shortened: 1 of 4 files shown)


stephenberry:
@makekryl, thanks for this! I just added more robust nullable type support that matches JSON for TOML and YAML. I appreciate the PR.

PR #2562, opened 2026-05-12
fix(toml): render writable_map_t struct fields as inline tables

### LINKED ISSUE #2540: toml: struct writer treats a map field as a scalar
Ran across this bug while migrating from `toml++` to `glaze`:

glaze's struct writer treats a map field as a scalar (PASS 1 in `glaze/toml/write.hpp`), prepending `items =`  and then handing off to the map writer, which emits bare `key = value\n` lines without `{}` wrapping. The two writers compose into invalid TOML. Read fails for the same reason.

# Code for repro:
```
 #include <glaze/glaze.hpp>

  #include <iostream>
  #include <map>
  #include <string>
  #include <vector>

  // Case 1: map<string, struct> as a struct field
  struct Item { bool enabled; };
  struct Cfg1 { std::map<std::string, Item> items; };

  template <> struct glz::meta<Item> {
      using T = Item;
      static constexpr auto value = object("enabled", &T::enabled);
  };
  template <> struct glz::meta<Cfg1> {
      using T = Cfg1;
      static constexpr auto value = object("items", &T::items);
  };

  // Case 2: map<string, bool> as a struct field
  struct Cfg2 { std::map<std::string, bool> items; };
  template <> struct glz::meta<Cfg2> {
      using T = Cfg2;
      static constexpr auto value = object("items", &T::items);
  };

  // Case 3: vector<struct{name, ...}> — workaround
  struct Entry { std::string name; bool enabled; };
  struct Cfg3 { std::vector<Entry> items; };

  template <> struct glz::meta<Entry> {
      using T = Entry;
      static constexpr auto value = object("name", &T::name, "enabled",
  &T::enabled);
  };
  template <> struct glz::meta<Cfg3> {
      using T = Cfg3;
      static constexpr auto value = object("items", &T::items);
  };

  int main() {
      {
          Cfg1 c{.items = {{"foo", {true}}, {"bar", {false}}}};
          std::string buf; glz::write_toml(c, buf);
          std::cout << "--- Case 1: map<string, struct> ---\n" << buf << "\n\n";
      }
      {
          Cfg2 c{.items = {{"foo", true}, {"bar", false}}};
          std::string buf; glz::write_toml(c, buf);
          std::cout << "--- Case 2: map<string, bool> ---\n" << buf << "\n\n";
      }
      {
          Cfg3 c{.items = {{"foo", true}, {"bar", false}}};
          std::string buf; glz::write_toml(c, buf);
          std::cout << "--- Case 3: vector<struct{name, ...}> ---\n" << buf << "\n";
      }
  }
```
  # Observed output (glaze 7.5):
```
  --- Case 1: map<string, struct> ---
  items = bar = enabled = false
  foo = enabled = true

  --- Case 2: map<string, bool> ---
  items = bar = false
  foo = true

  --- Case 3: vector<struct{name, ...}> ---
  [[items]]
  name = "foo"
  enabled = true

  [[items]]
  name = "bar"
  enabled = false
```

# Expected for Case 1 & 2
## either inline:
```
  items = { foo = { enabled = true }, bar = { enabled = false } }
```
  ## or table-header form:
```
  [items.foo]
  enabled = true

  [items.bar]
  enabled = false
```
(perhaps there could be a glaze config to choose?)

genisis0x:
## Summary
- Fixes #2540. glaze's TOML struct writer treated a ``writable_map_t`` field as a scalar in PASS 1: it emitted ``key = `` and then handed off to the top-level map writer, which produced bare ``inner_key = value\n`` lines without any wrapping. The two writers composed into invalid TOML such as

  ```
  items = bar = false
  foo = true
  ```

  and the document could not round-trip through ``glz::read_toml``.

- Add a dedicated ``write_map_field`` helper that renders a map-typed struct field as an inline TOML table on the right-hand side of the field assignment:

  - `map<K, scalar>` -> `items = { foo = true, bar = false }`
  - `map<K, struct>` -> `items = { foo = { enabled = true }, bar = { enabled = false } }` (nested struct via ``write_inline_object`` so the whole field stays inline-valid)
  - `map<K, vector<T>>` -> `items = { foo = [1, 2], bar = [3] }`
  - empty map -> `items = {}`

- Inline tables are syntactically scalars in TOML, so the dispatch sits in PASS 1 -- it must run before any nested ``[table]`` header opens a sub-scope, otherwise the inline-table assignment would be parsed as a key of the nested table instead of the enclosing struct. PASS 2 continues to handle nested objects and array-of-tables sections.

## Validation
- New test executable ``toml_map_field_test`` (see ``tests/toml_test/toml_map_field_test.cpp``) -- 32 assertions across 4 cases, all pass:
  - Case 1: `map<string, struct>` as a struct field
  - Case 2: `map<string, bool>` as a struct field
  - Case 3: `map<string, vector<int>>` as a struct field
  - Case 4: empty `map<string, struct>` field renders as `items = {}` and round-trips back to an empty map
- Each case writes the value, asserts the new inline form is emitted and the broken pre-fix ``items = bar = ...`` shape is NOT present, then reads the emitted output back and asserts the populated map matches the original.
- The test is standalone (no ``glz_test_common`` link) so it can run quickly in CI without depending on the rest of the heavier TOML test suite.

## Notes
- The fix matches the first form the reporter listed as acceptable: ``items = { foo = { enabled = true }, bar = { enabled = false } }``. The alternative table-header form ``[items.foo]\nenabled = true`` was considered but does not currently round-trip through ``glz::read_toml`` for ``map<K, struct>`` fields, so the inline form was chosen as the lower-risk shape that immediately closes the read/write asymmetry the reporter described.

Fixes #2540

### DIFF
--- a/include/glaze/toml/write.hpp
+++ b/include/glaze/toml/write.hpp
@@ -1087,8 +1087,109 @@ namespace glz
       // Check if inline_arrays mode is enabled
       constexpr bool inline_mode = use_inline_arrays<Options>();
 
+      // Helper lambda to write a map-typed struct field as a TOML table.
+      //
+      // Before this helper existed, ``writable_map_t`` fields fell through PASS 1
+      // as scalars: the struct writer emitted ``key = `` and then handed off to the
+      // map writer which produced bare ``inner_key = value\n`` lines, composing
+      // into invalid TOML such as ``items = bar = false\nfoo = true`` (issue
+      // #2540).  Render maps as inline TOML tables on the right-hand side of
+      // the field assignment so the document round-trips through
+      // ``glz::read_toml``.
+      auto write_map_field = [&]<size_t I>() {
+         using val_t = field_t<T, I>;
+         using map_value_t = typename std::remove_cvref_t<val_t>::mapped_type;
+         constexpr bool value_is_object = glaze_object_t<map_value_t> || reflectable<map_value_t>;
+
+         decltype(auto) m = [&]() -> decltype(auto) {
+            if constexpr (reflectable<T>) {
+               return get_member(value, get<I>(t));
+            }
+            else {
+               return get_member(value, get<I>(reflect<T>::values));
+            }
+         }();
+
+         static constexpr auto key = glz::get<I>(reflect<T>::keys);
+
+         if (!ensure_space(ctx, b, ix + padding)) [[unlikely]] {
+            return;
+         }
+         if (!first) {
+            std::memcpy(&b[ix], "\n", 1);
+            ++ix;
+         }
+         else {
+            first = false;
+         }
+
+         std::memcpy(&b[ix], key.data(), key.size());
+         ix += key.size();
+         std::memcpy(&b[ix], " = ", 3);
+         ix += 3;
+
+         if (empty_range(m)) {
+            if (!ensure_space(ctx, b, ix + 2)) [[unlikely]] {
+               return;
+            }
+            std::memcpy(&b[ix], "{}", 2);
+            ix += 2;
+            return;
+         }
+
+         if (!ensure_space(ctx, b, ix + 2)) [[unlikely]] {
+            return;
+         }
+         std::memcpy(&b[ix], "{ ", 2);
+         ix += 2;
+
+         bool inner_first = true;
+         for (auto&& [mk, mv] : m) {
+            if (bool(ctx.error)) [[unlikely]] {
+               return;
+            }
+            const auto mk_view = std::string_view{mk};
+            if (!ensure_space(ctx, b, ix + mk_view.size() + 6)) [[unlikely]] {
+               return;
+            }
+            if (!inner_first) {
+               std::memcpy(&b[ix], ", ", 2);
+               ix += 2;
+            }
+            else {
+               inner_first = false;
+            }
+            std::memcpy(&b[ix], mk_view.data(), mk_view.size());
+            ix += mk_view.size();
+            std::memcpy(&b[ix], " = ", 3);
+            ix += 3;
+
+            if constexpr (value_is_object) {
+               // Render nested struct values as inline tables so the entire
+               // map field remains inline-valid.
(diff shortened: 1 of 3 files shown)


nickdademo:
Would be nice to somehow support the alternative table-header form too eventually ;)

genisis0x:
Yeah good shout — left the header form out of this one to keep the scope tight, but happy to take it on as a follow-up if maintainers are open to it.

stephenberry:
This looks like a good initial fix. What are your thoughts between this implementation and the implementation I was developing in this PR: #2544?

genisis0x:
Skimmed #2544 and they're complementary in scope rather than competing:

**This PR (#2562)** is intentionally minimal:
- Single writer hook for `writable_map_t` struct fields -> emit inline table.
- One test (`tests/toml_test/toml_map_field_test.cpp`) covering the exact `std::map` field from #2540.
- No changes to `read.hpp`; the existing reader already round-trips the new output.

**#2544** is the broader fix:
- Writer-side: cascades inline-mode through nested writers (not just `writable_map_t` fields - also top-level struct/nested-map values).
- Reader-side: fixes the inline-table branch in struct read so `parse_toml_object_members` returns after consuming `}`, which my PR doesn't address.

If you'd rather land #2544 as the canonical fix, happy to close #2562 in favor. Alternatively I can rebase the `toml_map_field_test` here onto #2544 once it lands, so the #2540 regression case stays explicitly covered. Whichever you prefer.

stephenberry:
I just merged #2544, so you can either update this PR is there are gaps to close, or close this PR if the task at hand is finished. Thanks!

genisis0x:
Thanks @stephenberry — happy to defer to #2544 here. It supersedes everything this PR did (the writable_map_t writer hook, plus the inline-mode cascade for nested writers and the inline-table reader fix), so there are no gaps left to close from my side.

Closing this one out. Appreciate the fast turnaround.

PR #2563, opened 2026-05-12
feat(schema): distinguish input vs output JSON schemas (closes #2508)

### LINKED ISSUE #2508: glaze silently has different JSON schemas for parsing and serializaition due to partial parsing
I have detected differences between input and output JSON schemas:
**Output schemas**
- objects should have the `required: [ <required keys list>]` property because after serialization they are guarranteed to exist.
- arrays that come from tuples should also have the `minItems` set to the same value of `maxItems` property because the tuple will always serialize with full length. 

**Input schemas**
- My understanding of the library is not deep enough but I guess due to partial parsing input schemas for objects do not have the `required` property.
- In a recent pull https://github.com/stephenberry/glaze/pull/2461 you added the `maxItems` for tuple parsing but did not include the `minElements` because glaze supports partial parsing (I don't find that anymore but I think I have read it)

I think we should make this difference explicit and pass an option to whether we want the input or output JSON schema  of a C++ struct because strictly speaking they are different since output has other guarrantees than input has requirements.

genisis0x:
## Summary

Closes #2508.

Glaze previously emitted a single JSON Schema for every type even though serialization and parsing contracts differ. The mixed behavior — `required` on objects but no `minItems` on tuples after #2461 — described neither contract fully.

This PR introduces `schema_purpose { input, output }` and threads it through `to_json_schema` via an `opts_with_schema_purpose` wrapper that carries the tag in the existing constexpr `Opts`.

New entry points:

- `glz::write_json_input_schema<T>()` — partial-parse friendly: no `required` on objects, no `minItems` on tuples.
- `glz::write_json_output_schema<T>()` — serialization contract: `required` listed for guaranteed-present keys, `minItems == maxItems` on tuples.

`glz::write_json_schema<T>()` keeps its existing signature and now defaults to `output` purpose. For objects this preserves the existing `required` emission; for tuples this newly emits `minItems == maxItems`, completing the output-schema contract.

User-provided `meta<V>::required` is always honored regardless of purpose, since it expresses a developer-asserted parsing requirement. Variant discriminator tags also stay in `required` for input schemas because the tag must be present to select an alternative.

## Test plan

- [x] `jsonschema_test` builds cleanly.
- [x] Updated `tuple schema uses prefixItems` expected string to include `minItems`.
- [x] New `schema_purpose_test` suite covers:
  - input schema for tuples omits `minItems`,
  - output schema for tuples sets `minItems == maxItems`,
  - input schema for objects omits `required` (with `error_on_missing_keys`),
  - output schema for objects lists non-nullable members in `required`,
  - default `write_json_schema` matches output purpose.

### DIFF
--- a/include/glaze/json/schema.hpp
+++ b/include/glaze/json/schema.hpp
@@ -143,6 +143,28 @@ namespace glz
       }
    };
 
+   // Distinguish JSON Schemas generated for serialization output vs parsing input.
+   //
+   // output: schema describes what glaze will produce. All non-skipped object members
+   //         are guaranteed to be written, and tuples always serialize at their full
+   //         length, so the schema emits `required` and `minItems = maxItems`.
+   //
+   // input:  schema describes what glaze will accept. Partial parsing is supported,
+   //         so `required` is omitted on objects (except for variant discriminator
+   //         tags, which must be present to select an alternative) and `minItems`
+   //         is omitted on tuples.
+   enum struct schema_purpose : uint8_t { input, output };
+
+   consteval schema_purpose check_schema_purpose(auto&& Opts)
+   {
+      if constexpr (requires { Opts.schema_purpose_v; }) {
+         return Opts.schema_purpose_v;
+      }
+      else {
+         return schema_purpose::output;
+      }
+   }
+
    namespace detail
    {
       enum struct defined_formats : uint32_t;
@@ -907,6 +929,11 @@ namespace glz
             });
             s.items = false;
             s.maxItems = N;
+            if constexpr (check_schema_purpose(Opts) == schema_purpose::output) {
+               // Output schemas describe serialized data: tuples always emit every element,
+               // so the array length is fixed (minItems == maxItems == N).
+               s.minItems = N;
+            }
          }
       };
 
@@ -973,6 +1000,8 @@ namespace glz
 
             using V = std::decay_t<T>;
 
+            constexpr auto purpose = check_schema_purpose(Opts);
+
             if constexpr (requires { meta<V>::required; }) {
                if (!s.required) {
                   s.required = std::vector<sv>{};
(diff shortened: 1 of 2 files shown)


wilkolbrzym-coder (changes requested):
Review – PR #2563

Not approved.

The conceptual input/output split is correct. The implementation has public API and portability problems.

1. Old call form likely ambiguous

Existing call:

write_json_schema<T, Opts>(buffer);

The PR adds:

template <class T, auto Opts = opts{}, schema_purpose Purpose = schema_purpose::output, class Buffer>
error_ctx write_json_schema(Buffer&& buffer);

This competes with the existing overload. Since Purpose has a default, write_json_schema<T, Opts>(buffer) is ambiguous.

This is an API break and is not tested.

Add a compile test for the old two-template-argument form.

2. opts_with_schema_purpose construction is fragile

static constexpr auto schema_opts =
    opts_with_schema_purpose<decltype(Opts)>{{Opts}, Purpose};

This depends on aggregate initialization of a derived class containing a base class and a member.

It is not guaranteed to work for every valid custom opts struct accepted by Glaze.

Replace with an explicit constructor:

constexpr opts_with_schema_purpose(const std::decay_t<Opts>& o,
                                   schema_purpose p)
    : std::decay_t<Opts>(o), schema_purpose_v(p) {}

Then construct:

opts_with_schema_purpose<decltype(Opts)>{Opts, Purpose};

3. Input schema ignores error_on_missing_keys without documentation

write_json_input_schema<T, glz::opts{.error_on_missing_keys = true}>();

still omits required.

A user may reasonably expect required to appear because error_on_missing_keys is explicitly enabled.

This behavior must be documented or changed.

4. Required-list construction is fragmented

meta<V>::required and output-purpose requires_key are handled in separate places.

That is correct today, but fragile. Centralize the construction of s.required.

5. write<options> is still hardcoded

The final schema output is written as:

write<options>(...)

not with Opts or schema_opts.

Caller formatting/prettify options are ignored.

Document this or fix it.

6. Missing tests

- old write_json_schema<T, Opts>(buffer) call
- meta<V>::required with input purpose
- variant discriminator required tag
- nested object/tuple schemas
- custom non-aggregate opts
- explicit error_on_missing_keys input schema expectation

Verdict

Request changes.

Fix the overload ambiguity, add an explicit options wrapper constructor, document input-schema semantics, and add compatibility tests.

cc @stephenberry @genisis0x

PR #2564, opened 2026-05-13
feat(rpc, net): allow registry + http_router to compile under -fno-exceptions

### LINKED ISSUE #2265: rpc registry: support for no exceptions
Just tried the registry and came across three try-catch blocks, which is an issue for me since my project does not support exceptions
I removed them and it seems to work for now but there may be issues I dont know about (at least I found no throw in the sources)

It would be nice to be able to use the registry in exception free projects

edit: 
Just ran into the issue that I am not able to 'fail' a request without exceptions - would need some kind of std::expected to jsonrpc error translation..

genisis0x:
## Summary
- Closes #2265. \`registry::call\` has three \`try\`/\`catch\` blocks (REPE message-based, REPE span-based, JSON-RPC) wrapping the registered handler invocation. Transitively the registry pulls in \`net/http_router.hpp\`, which has one \`try\`/\`catch\` plus three \`throw std::runtime_error\` route-misuse sites. Under \`-fno-exceptions\` none of that compiles, so the rpc registry could not be used in exception-free projects at all — the scenario @redrezo described in the issue.

## Fix
- Guard every site with \`#if __cpp_exceptions\`, matching the convention already used by \`glaze_exceptions.hpp\`, \`core_exceptions.hpp\`, and \`cli_menu.hpp\`.

### \`registry.hpp\` — three sites
- When exceptions are off, the \`catch\` is elided and the handler is invoked directly. Any exception escaping user code in that mode is UB per the \`-fno-exceptions\` contract; documented inline.

### \`http_router.hpp\` — four sites
- \`add(...)\` try/catch: under \`-fno-exceptions\` the routing install is performed without try-coverage; misuse below still surfaces on stderr through the matching \`#else\` branches.
- Three route-misuse \`throw\` sites (parameter-name conflict, wildcard-not-last, wildcard-name conflict): under \`-fno-exceptions\` each becomes \`fprintf(stderr, ...)\` + \`std::abort()\`. Misconfiguration is still loud and fatal, which matches the existing throw behaviour for callers that did not catch.

## Verification
\`\`\`
clang++ -std=c++23 -fno-exceptions -Iinclude /tmp/glaze_noexc_check.cpp -o /tmp/glaze_noexc_check
# (formerly: \"cannot use 'try' with exceptions disabled\" x4 + \"cannot use 'throw' with exceptions disabled\" x3)
# now: clean build, exit 0
\`\`\`

Default with-exceptions build: token-identical to before this commit (the \`#else\` branches only activate under \`-fno-exceptions\`). Local \`jsonrpc_test\` and \`repe_test\` rebuild and link unchanged.

## Compatibility
- No public-API change.
- No behaviour change for the default build.
- New: the registry + http_router compile under \`-fno-exceptions\`.

### DIFF
--- a/include/glaze/net/http_router.hpp
+++ b/include/glaze/net/http_router.hpp
@@ -299,6 +299,7 @@ namespace glz
        */
       void add(http_method method, std::string_view path, H handle, const route_spec& spec = {})
       {
+#if __cpp_exceptions
          try {
             // Install in the radix tree first. add_route can throw on conflict
             // (duplicate :param or *wildcard names at the same position); if it
@@ -316,6 +317,16 @@ namespace glz
             std::fprintf(stderr, "Error adding route '%.*s': %s\n", static_cast<int>(path.length()), path.data(),
                          e.what());
          }
+#else
+         // Under -fno-exceptions add_route's misuse-detection throws are
+         // compiled to fprintf+abort (see the matching #else branches below),
+         // so we cannot observe the conflict here and simply propagate the
+         // install. Misconfigured routes still get reported on stderr.
+         add_route(method, path, handle, spec.constraints);
+         auto& entry = routes[std::string(path)][method];
+         entry.handle = std::move(handle);
+         entry.spec = spec;
+#endif
       }
 
       /**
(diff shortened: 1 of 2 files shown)


genisis0x:
Heads up on the red `build (13, Debug, 23)` jobs: the failure is `get_name.hpp:197 [-fpermissive] 'glz_write_error' ... is used but never defined` while compiling `tests/networking_tests/http_router_test.cpp` — same compile error that's red on the comment-only #2567 and that didn't block #2565 from merging. This PR only adds `#if __cpp_exceptions` guards around existing throw sites; the default-exceptions path GCC 13 takes is byte-identical to develop, so the failure isn't introduced here.

Happy to rebase onto develop after #2544 settled if that helps a re-run.

PR #2565, opened 2026-05-13
fix(http): emit structured glz_write_error from response::body<Opts> on serialization failure

genisis0x:
## Summary
Resolves the \`// TODO serialize a struct into proper format for errors\` comment in \`response::body<Opts>(value)\` (\`include/glaze/net/http_router.hpp\`, line 99 on \`main\`).

## What was wrong
On a failed \`glz::write\`, the response body was reset to a hardcoded literal:

\`\`\`
{\"error\":\"glz::write_json error\"}
\`\`\`

Two issues:

1. The underlying \`error_ctx\` (error code + custom_error_message) was discarded — callers had no way to tell *which* write failed or why.
2. Because the failed write may have already pushed bytes into \`response_body\` before erroring, the literal was **appended** to a partial payload, producing malformed JSON like \`{\"name\":\"abc{\"error\":\"glz::write_json error\"}\`.

## Fix
- Clear \`response_body\` before writing the error report (no partial-payload mixing).
- Serialise a small in-scope struct through \`glz::write_json\`:

\`\`\`cpp
struct glz_write_error {
    std::string_view error = \"glaze write failure\";
    uint32_t code{};
    std::string_view message{};
};
\`\`\`

So callers now get structured error info they can introspect:

\`\`\`json
{\"error\":\"glaze write failure\",\"code\":42,\"message\":\"<custom_error_message from error_ctx>\"}
\`\`\`

If the error-struct write itself errors (essentially impossible for a trivial fixed-shape value, but handled for completeness) we fall back to a safe static string so the response is never left empty.

## Compatibility
- No public API change.
- The success path is token-identical to before — only the error branch is touched.
- Existing coverage (\`tests/networking_tests/http_server_api_tests/http_server_api_tests.cpp:704\`) exercises the success path and continues to pass.

### DIFF
--- a/include/glaze/net/http_router.hpp
+++ b/include/glaze/net/http_router.hpp
@@ -96,8 +96,31 @@ namespace glz
          }
          auto ec = glz::write<Opts>(std::forward<T>(value), response_body);
          if (ec) {
-            // TODO serialize a struct into proper format for errors
-            response_body = R"({"error":"glz::write_json error"})"; // rare that this would ever happen
+            // The body may have been partially written before the error
+            // fired; clear it so the response is never a mix of a partial
+            // success payload and the error report.
+            response_body.clear();
+
+            // Serialize a structured glz_write_error object instead of the
+            // legacy hardcoded placeholder. The original error_code and
+            // any custom_error_message are surfaced so callers can debug
+            // failed writes without having to reproduce them locally.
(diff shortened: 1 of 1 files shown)


stephenberry:
This looks great! Thanks for implementing this TODO. I'm merging it in.

PR #2570, opened 2026-05-15
chore(progress_bar): build status string with std::format

genisis0x:
### Summary

`progress_bar::string()` composes its percent / ETA / progress trailer with hand-rolled `std::to_string` concatenation, with a stale comment block right after it sketching a `fmt::format_to` / `std::format` replacement marked `TODO: use std::format when available`.

Glaze now requires C++23 (`cxx_std_23` in `CMakeLists.txt`) and already uses `std::format` in `include/glaze/core/write_chars.hpp`, so the placeholder can finally land:

```cpp
std::format_to(std::back_inserter(s),
               " {}% | ETA: {}m {}s | {}/{}",
               percentage, minutes, seconds,
               one_or_completed, one_or_total);
```

### Output

Byte-identical to the previous concatenation for all in-range values — `percentage` is already `size_t`, and the old `std::lround(percentage)` was a no-op cast to `long` before going through `std::to_string`. Smoke-checked locally (Apple clang 17, `-std=c++23`):

```
[=====-------------] 25% | ETA: 0m 30s | 25/100
 0% | ETA: 0m 0s | 0/1
 100% | ETA: 0m 0s | 100/100
```

### Change

`include/glaze/util/progress_bar.hpp`:

- Add `#include <format>`.
- Replace the three `s += "..." + std::to_string(...)` lines and the stale `fmt::format_to` comment block with a single `std::format_to` call.

### DIFF
--- a/include/glaze/util/progress_bar.hpp
+++ b/include/glaze/util/progress_bar.hpp
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <cmath>
+#include <format>
 #include <ostream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for this, but I decided to go with a lower level implementation in #2573 to reduce compilation times and binary usage.

PR #2571, opened 2026-05-15
Enhance JSON reading logic to validate buffer length before assignment

Zetonem:
Some projects compile Glaze as a header-only library with GCC 16 and strict warnings, especially -Warray-bounds and -Werror. In one place we copy a short JSON string into a tiny 8-byte stack buffer, then build a std::string from that buffer.

The compiler couldn’t always see that the length we pass to assign is safe, so it warned that we might read past the buffer even when the code is actually fine. With -Werror, that warning stops the build.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -1275,7 +1275,11 @@ namespace glz
                   while (it < end) [[likely]] {
                      *p = *it;
                      if (*it == '"') {
-                        value.assign(buffer.data(), size_t(p - buffer.data()));
+                        const size_t n = size_t(p - buffer.data());
+#if __has_cpp_attribute(assume) >= 202207L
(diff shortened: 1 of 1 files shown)


stephenberry:
Can you check and see if [[assume]] can be used to avoid this warning? I would rather not create an unnecessary branch to fix a warning.

Zetonem:
> Can you check and see if [[assume]] can be used to avoid this warning? I would rather not create an unnecessary branch to fix a warning.

For my project this approach doesn't work :(

stephenberry:
@Zetonem, I implemented `[[assume(n <= sizeof(buffer))]];` and a CI check with `-Wall` and it seems to work. Can you try this fix on your codebase and give your error if you still get one? If it still fails, we can #pragma ignore the warning.

PR #2575, opened 2026-05-16
toml: fix parsing arrays with newlines

makekryl:


include/glaze/toml/common.hpp
```diff
@@ -25,6 +25,25 @@ namespace glz
...
+            while (it != end && *it != '\n' && *it != '\r') {
+               ++it;
+            }
+         }
+         else {
+            break;
+         }
+      }
+   }
+
```
makekryl, line 46: It does not seem good (hands are itching to merge with `skip_ws_and_comments` via a bool template parameter), on the other hand this is a simple way to do that without overcomplicating `¯\_(ツ)_/¯`

makekryl: I'm also not sure whether it would require a separate benchmark like `skip_ws_benchmark`

stephenberry: I think a separate function is good and clear to read. No need to add a separate benchmark for this. At some point I'll work on tighter optimization of TOML, but performance isn't as critical typically for TOML usage.

stephenberry:
I like this! Thanks for the PR, merging it in.

PR #2576, opened 2026-05-18
toml: fix reading nested nullable tables and nullable arrays

makekryl:


include/glaze/yaml/read.hpp
```diff
@@ -2113,30 +2113,17 @@ namespace glz
...
-            else if constexpr (constructible<T>) {
-               value = meta_construct_v<T>();
-            }
-            else if constexpr (std::is_pointer_v<T> && can_allocate_raw_pointer<Opts, std::decay_t<decltype(ctx)>>) {
-               if (!try_allocate_raw_pointer<Opts>(value, ctx)) {
+            if (!nullable_emplace(value)) {
+               if constexpr (std::is_pointer_v<T> && can_allocate_raw_pointer<Opts, std::decay_t<decltype(ctx)>>) {
+                  if (!try_allocate_raw_pointer<Opts>(value, ctx)) {
+                     return;
+                  }
```
makekryl: Maybe it would be better to pass the context and options and put raw pointer initialization into `nullable_emplace` too

include/glaze/util/nullable_traits.hpp
makekryl, line 1: Let me know if there's a more suitable place for this file

makekryl: (and maybe name too, it did not include the function initially, but as for now it's definitely no longer _traits_ only)

makekryl:
@stephenberry, I'm not sure why the packit actions failed, can you check please?

stephenberry:
The packit actions often fail because the servers go down (perhaps are overwhelmed). Rarely a concern. I only look at them before I finally merge if they're still failing.

stephenberry:
Thanks for this PR! I made some tweaks to better de-duplicate code across formats. I'm merging this in, but feel free to open more PRs if you think more improvements are needed.

PR #2577, opened 2026-05-18
cmake: allow to skip install for glaze

FrogTheFrog:
Hello o/

Currently, when using the recommended cmake setup:
```
include(FetchContent)
FetchContent_Declare(
        glaze
        GIT_REPOSITORY https://github.com/stephenberry/glaze.git
        GIT_TAG v7.5.0
        GIT_SHALLOW TRUE
)
FetchContent_MakeAvailable(glaze)
```
glaze adds itself to the installation rules by default, even when it's not a top level project, which results in glaze being  picked up by CPack installers:
image

This can be disabled by manipulating global `CMAKE_SKIP_INSTALL_RULES` variable somewhat:
```
include(FetchContent)
FetchContent_Declare(
        glaze
        GIT_REPOSITORY https://github.com/stephenberry/glaze.git
        GIT_TAG v7.5.0
        GIT_SHALLOW TRUE
)
set(CMAKE_SKIP_INSTALL_RULES TRUE)
FetchContent_MakeAvailable(glaze)
set(CMAKE_SKIP_INSTALL_RULES FALSE FORCE)
```
its awkward and works locally, but in CI it results in different issue:
image

For that reason I propose to introduce a new cmake `glaze_SKIP_INSTALL_RULES` variable to makes this less of a hassle:
```
include(FetchContent)
FetchContent_Declare(
        glaze
        GIT_REPOSITORY https://github.com/stephenberry/glaze.git
        GIT_TAG v7.5.0
        GIT_SHALLOW TRUE
)
set(glaze_SKIP_INSTALL_RULES ON)
FetchContent_MakeAvailable(glaze)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,8 @@ target_include_directories(
     INTERFACE "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
 )
 
-if(NOT CMAKE_SKIP_INSTALL_RULES)
+option(glaze_SKIP_INSTALL_RULES "Skip generating installation rules for glaze" OFF)
+if(NOT CMAKE_SKIP_INSTALL_RULES AND NOT glaze_SKIP_INSTALL_RULES)
   include(cmake/install-rules.cmake)
 endif()
 


stephenberry:
@FrogTheFrog, Glaze already used a `PROJECT_IS_TOP_LEVEL` CMake variable. I just implemented #2578 that adds a glaze_INSTALL option that defaults to using `PROJECT_IS_TOP_LEVEL`. I think this is a better solution that should work for you. Let me know if there are any downsides you'll have to this alternative solution. Thanks for this PR and brining this install issue to my attention!

FrogTheFrog:
This is indeed a better solution and it does work for me :).

I'm closing this in favor of the other PR.

stephenberry:
Great, I'll merge in the other PR. Thanks for the feedback.

PR #2585, opened 2026-05-25
feat(yaml): tagged variant read support (closes #2583)

### LINKED ISSUE #2583: YAML Tagged variants are not supported - error_code::no_matching_variant_type
Hi,
Tried to use tagged variants with YAML that are fine in JSON and they are not working
Parsing fails with `error_code::no_matching_variant_type`

Tried to pinpoint the problem and it does seem like there's no tagged variants support for YAML

For now wrote custom read method that does the duty (as a workaround)
But it would be nice to see the feature 'from the box'

Thanks a lot, the Project is awesome!

P.S. by custom reader I mean this:
```
template<>
struct glz::from<glz::YAML, MyVariant>
{
    using TVariant = MyVariant;
    template<auto Opts>
    static void op(TVariant& value, is_context auto&& ctx, auto&& it, auto&& end)
    {
        std::string key{};
        if (!glz::yaml::parse_yaml_key(key, ctx, it, end, true))
        {
            return;
        }
        glz::yaml::skip_flow_ws_and_newlines(ctx, it, end);
        if (it == end || *it != ':')
        {
            ctx.error = glz::error_code::syntax_error;
            return;
        }

        ++it;
        glz::yaml::skip_inline_ws(it, end);

        std::string typeID{};
        glz::from<glz::YAML, std::string>::template op<Opts>(typeID, ctx, it, end);

        const auto index =
            glz::variant_id_to_index<TVariant>::op(typeID.data(), typeID.data() + typeID.size(), typeID.size());

        if (index >= glz::ids_v<TVariant>.size())
        {
            ctx.error = glz::error_code::no_matching_variant_type;
            return;
        }

        glz::emplace_runtime_variant(value, index);
        std::visit(... ... ...
```

genisis0x:
## Why

The YAML reader's `from<YAML, std::variant<...>>` only handled auto-deducible variants (distinguishable by JSON-like shape). Variants that declare a `glz::meta` `tag` + `ids` — the same dispatch the JSON reader already uses — fell through and produced `error_code::no_matching_variant_type`, forcing users to provide a hand-rolled `from<YAML, T>` specialization to read them. Issue [#2583](https://github.com/stephenberry/glaze/issues/2583) shows the workaround.

## What

Add a tagged-variant specialization of `from<YAML, T>` guarded on `!tag_v<T>.empty()` that mirrors the JSON tag-driven dispatch:

- parse the node preamble (anchor / tag / alias) like a normal mapping
- read the first mapping key and require it to match `tag_v<T>`
- read the id value (string or integral), look up the alternative via `variant_id_to_index<T>`, and `emplace_runtime_variant`
- rewind the iterator to the mapping start and re-dispatch the same mapping into the chosen alternative's struct reader, with `error_on_unknown_keys` flipped off so the tag key passes through silently on the second pass

The existing `is_variant T` specialization is now constrained on `tag_v<T>.empty()` so partial ordering picks the tagged version whenever a tag is declared.

```cpp
struct alt_a { int value{}; };
struct alt_b { std::string text{}; };
template <> struct glz::meta<alt_a> { static constexpr auto value = object(\"value\", &alt_a::value); };
template <> struct glz::meta<alt_b> { static constexpr auto value = object(\"text\", &alt_b::text); };

using V = std::variant<alt_a, alt_b>;
template <> struct glz::meta<V> {
    static constexpr std::string_view tag = \"type\";
    static constexpr auto ids = std::array{\"alt_a\", \"alt_b\"};
};

V v;
glz::read_yaml(v, \"type: alt_b\\ntext: hello\"); // now succeeds, v holds alt_b{\"hello\"}
```

## Tests

`tests/yaml_test/yaml_test.cpp` — new `yaml_tagged_variant_tests` suite covers both flow `{type: alt_a, value: 42}` and block `type: alt_a\\nvalue: 42` for each alternative, plus an unknown-id error path.

## Scope

- The tag must be the **first** key of the mapping. The JSON path tolerates out-of-order tags via field-set deduction; happy to extend that here in a follow-up.
- Only struct alternatives (`glaze_object_t` / `reflectable`) are dispatched. Scalar / sequence alternatives under a tag surface `no_matching_variant_type` rather than silently mis-parsing.

Closes #2583

### DIFF
--- a/include/glaze/yaml/read.hpp
+++ b/include/glaze/yaml/read.hpp
@@ -5013,8 +5013,135 @@ namespace glz
       }
       process_yaml_variant_alternatives<Variant, is_yaml_variant_str>::template op<Opts>(value, ctx, it, end);
    }
+   // Tagged variants (glz::meta declares `tag`/`ids`). Mirrors the JSON
+   // reader's tag-driven dispatch: read the tag key, look up the matching
+   // variant alternative by id, emplace it, then re-dispatch the same
+   // mapping into the alternative struct. The inner parse runs with
+   // unknown-key tolerance so the tag key itself is silently skipped on
+   // the second pass (the alternative struct does not declare it).
+   //
+   // Restriction: the tag must be the *first* key of the mapping (flow
+   // or block). The JSON path supports out-of-order tags via field-set
+   // deduction, which is reasonable to extend here in a follow-up.
+   template <is_variant T>
+      requires(!tag_v<T>.empty())
+   struct from<YAML, T>
+   {
+      template <auto Opts, class It>
+      static void op(auto&& value, is_context auto&& ctx, It&& it, auto end) noexcept
+      {
+         if (bool(ctx.error)) [[unlikely]]
+            return;
+
+         // Honor the standard preamble (anchors / tags / aliases) so the
+         // tagged variant behaves like a regular mapping for these
+         // properties.
+         yaml::node_preamble_state preamble{};
+         if (yaml::parse_node_preamble<Opts, yaml::node_property_policy{true, true, true, false}, true>(
+                value, ctx, it, end, preamble, [](yaml::yaml_tag tag) { return yaml::tag_valid_for_map(tag); }))
+            return;
+
+         if (it == end) [[unlikely]] {
+            ctx.error = error_code::unexpected_end;
+            return;
+         }
+
+         const bool flow_mapping = (*it == '{');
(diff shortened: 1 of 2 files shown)


stephenberry:
Do you want to work on the failing CI cases?

stephenberry:
Thanks for this, but I implemented a more mature solution with #2586, so I'm closing this one.

PR #2590, opened 2026-05-30
CI workflow for MSVC O2

annihilatorq:
I noticed that glaze uses `asio::awaitable` in its code, and immediately recalled a time debugging a crash on O2 caused by `asio::co_composed`, and even though I didn’t see any use of `asio::co_composed_future` in glaze - I thought it wouldn’t hurt to add an extra check for the O2 build for MSVC, because it's probably one of the most finicky compilers out there.

Motivation is my recent very negative experience with the MSVC optimizer: https://github.com/chriskohlhoff/asio/pull/1724.
In short, MSVC emitted an aligned 128-bit `movaps`, even though the operand address was aligned to 8 bytes, not 16, so this resulted in `STATUS_ACCESS_VIOLATION`

### DIFF
--- a/.github/workflows/msvc.yml
+++ b/.github/workflows/msvc.yml
@@ -16,10 +16,6 @@ on:
       - 'docs/**'
   workflow_dispatch:
 
-env:
-  CTEST_OUTPUT_ON_FAILURE: 1
-  BUILD_TYPE: Debug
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for this!

PR #2593, opened 2026-05-30
Guard std::min & std::max calls from minwindef.h defines

annihilatorq:
This one is trivial, but also a fatal user-facing issue when building on Windows & including `Windows.h` before **some** glaze headers (some headers already use `()` guards). any glaze function compiled into the final target whose headers were included after `Windows.h` and that uses `std::min` or `std::max` without guards will cause the compilation to fail.
fix assumes that not all users are aware of this historical issue and `#define NOMINMAX`, but in any case, there's no point in forcing them to use it

```
[build] C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\shared\minwindef.h(197,29): note: expanded from macro 'min'
[build]   197 | #define min(a,b)            (((a) < (b)) ? (a) : (b))
[build]       |                             ^
[build] In file included from C:\artem\!cpp\proj\src\main.cpp:10:
[build] C:\artem\!cpp\proj\build\_deps\glaze-src\include\glaze/containers/inplace_vector.hpp(852,44): error: expected unqualified-id
[build]   852 |             for (size_type i = 0; i < std::min(n, this->storage_size() - pos_idx); ++i) {
[build]       |                                            ^
...
```

### DIFF
--- a/benchmarks/http_benchmark/http_benchmark.cpp
+++ b/benchmarks/http_benchmark/http_benchmark.cpp
@@ -529,7 +529,7 @@ int main()
    static constexpr uint16_t beast_port = 18766;
    static constexpr int batch = 1000;
 
-   const unsigned hw_threads = std::max(1u, std::thread::hardware_concurrency());
+   const unsigned hw_threads = (std::max)(1u, std::thread::hardware_concurrency());
    const int client_threads = static_cast<int>(hw_threads);
 
    const auto plaintext_req = make_get_req("/plaintext");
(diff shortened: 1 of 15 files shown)


stephenberry:
I prefer to guard `std::min` and `std::max` like this: `(std::min)(a, b)`. Adding parentheses are how Glaze does it in general, but it looks like you caught some missing locations.

annihilatorq:
> but it looks like you caught some missing locations.

Yeah, I happened to notice this in `glz::inplace_vector`, so I decided to quickly scan the whole project. I'm used to guarding `min` and `max` in a similar way, too.

stephenberry:
Oh, I see you've implemented just that. Thanks!

annihilatorq:
@stephenberry The test that was failing reminded me a lot of a flaky test, since it relied on the scheduler granting at least one write-lock in 100ms, while continuously acquiring read-locks from other threads. Now it should work as intended, please review it when you have a moment

stephenberry:
Thanks for fixing that flaky test. The fix looks good.

PR #2597, opened 2026-06-05
Guard std::numeric_limits<T>::min/max

annihilatorq:
Same as #2593 (Guard std::min & std::max calls from minwindef.h defines), but for `std::numeric_limits`.


stephenberry:
Thanks!

PR #2598, opened 2026-06-05
Use std::in_range for integer bounds checks

annihilatorq:
Replaces checks similar to
```cpp
value >= (std::numeric_limits<int8_t>::min)() && value <= (std::numeric_limits<int8_t>::max)()
```
by shorter C++20 alternative
```cpp
std::in_range<int8_t>(value)
```

P.S. I haven't touched functions with templated types, since passed char-like types can satisfy `std::integral` and have `std::numeric_limits` specializations, but at the same time `std::in_range<Rx>` can fail for those types, because standard constrain `Rx` to standard integer types.

### DIFF
--- a/include/glaze/bson/write.hpp
+++ b/include/glaze/bson/write.hpp
@@ -12,6 +12,7 @@
 #include <cstring>
 #include <limits>
 #include <string_view>
+#include <utility>
 
 #include "glaze/bson/header.hpp"
 #include "glaze/core/buffer_traits.hpp"
@@ -321,7 +322,7 @@ namespace glz
                return;
             }
             if constexpr (std::same_as<T, uint64_t>) {
-               if (value > static_cast<uint64_t>((std::numeric_limits<int64_t>::max)())) [[unlikely]] {
+               if (!std::in_range<int64_t>(value)) [[unlikely]] {
                   ctx.error = error_code::invalid_length;
                   return;
                }
--- a/include/glaze/msgpack/write.hpp
+++ b/include/glaze/msgpack/write.hpp
@@ -5,14 +5,12 @@
 
 #include <algorithm>
 #include <chrono>
-#include <limits>
 #include <utility>
 
 #include "glaze/core/buffer_traits.hpp"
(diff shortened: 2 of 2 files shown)


stephenberry:
Thanks for this, and for the care that went into it — the semantics are correct throughout, and I appreciate that you deliberately left the templated/char-like paths alone, since `std::in_range` would be ill-formed there. So this isn't a correctness concern at all.

I'm going to decline it, though, and the reason is about what glaze optimizes for rather than anything wrong with the change. Compile time is a first-class concern for this library, and this is effectively a lateral move: readability improves a little, but it adds template instantiations (`std::in_range` plus the `std::cmp_*` helpers it pulls in) where the previous code lowered to plain builtin comparisons at zero front-end cost.

To put rough numbers on it: in a representative TU exercising these write paths (Clang, `-O2`, `-ftime-trace`), the change adds ~3.5 ms of front-end template-instantiation time (≈1.7 ms for `std::in_range<…>` and ≈1.8 ms for the `std::cmp_*` helpers it instantiates), against a ~1.58 s front-end — about 0.2%, and lost in wall-clock noise at the single-TU level (~2.13 s either way). It's a one-time-per-TU cost: 12 distinct `in_range<R, T>` instantiations memoized across all ~28 call sites, not per-site. So per TU it's genuinely small — my concern is that these costs accumulate across translation units and set a precedent that propagates to future bounds checks, with no correctness or runtime win to bank against it (the generated comparisons are essentially equivalent).

The other thing is that the readability gain is concentrated in the two-sided signed checks (`value >= min && value <= max`), which genuinely read better as `in_range`. But most of the diff replaces one-sided unsigned checks (`value <= max`), which were already clear, and `in_range<uint8_t>(value)` arguably obscures that only an upper bound matters there.

For those reasons I'd prefer to keep the explicit comparison idiom. Really do appreciate the contribution — please don't let this discourage you from sending others.

annihilatorq:
Oh, I see, that does sound completely reasonable. Thanks for explaining!

PR #2600, opened 2026-06-06
Fix WebSocket client handling data received with handshake response

annihilatorq:
## Main Issue
The main issue is that Glaze currently invoking `ws_conn->set_initial_data` **before** the user callbacks are set up, and it leads to:
```
1. set_initial_data
2. process_frames
3. handle_frame
4. dispatch_message {
     if (client_message_handler_) {
       client_message_handler_(...)
     }
   }
```
which expects message callback to already be installed.
This caused the message received as a result of reading data from the socket after the double-CRLF delimiter to be completely lost.

I also fixed an issue with reading the remaining data from `asio::streambuf`, which, in the current Asio implementation wasn't actually a bug, but according to the contract, Asio states that this approach should be avoided. The current `asio::basic_streambuf` implementation uses a single contiguous char array, but the class's API contract is designed for multiple buffer segments (which is why `.data` returns a `ConstBufferSequence` rather than for example a `std::span`), and if tomorrow Asio decides to implement `basic_streambuf` differently, the current code for reading the remaining data from `response_buf` may encounter some problems, since it relies on there being a single buffer segment left in `response_buf`, but this is not guaranteed by the asio API.

I also changed the signature of `set_initial_data` to accept a `std::vector<uint8_t>` to avoid copying the received data twice (via `asio::buffer_copy` and `frame_buffer_.insert()`). I replaced `std::string_view` with `std::vector<uint8_t>` specifically because using `std::string` makes moving impossible due to the different `value_type` inside the `frame_buffer_` and `data` iterators.

## Part that requires a review
I also fixed one issue that seemed wrong to me, but I can't be sure that it isn't part of the API design. It seemed strange to me that the read loop (`start_read`) is started before `on_open` was called. Couldn't this create a problem where the user's message callback might be called before the `on_open` callback, leading to an unobvious API?

### DIFF
--- a/include/glaze/net/websocket_client.hpp
+++ b/include/glaze/net/websocket_client.hpp
@@ -452,18 +452,9 @@ namespace glz
                                       ws_conn->set_client_mode(true);
                                       ws_conn->set_max_message_size(self->max_message_size);
 
-                                      if (response_buf->size() > 0) {
-                                         std::string_view initial_data{
-                                            static_cast<const char*>(response_buf->data().data()),
-                                            response_buf->size()};
-                                         ws_conn->set_initial_data(initial_data);
-                                      }
-
                                       if (self->on_message && *self->on_message) ws_conn->on_message(*self->on_message);
                                       if (self->on_close && *self->on_close) ws_conn->on_close(*self->on_close);
                                       if (self->on_error && *self->on_error) ws_conn->on_error(*self->on_error);
-
-                                      ws_conn->start_read();
 
                                       {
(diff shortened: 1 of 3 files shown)


stephenberry:
> I also fixed one issue that seemed wrong to me, but I can't be sure that it isn't part of the API design. It seemed strange to me that the read loop (start_read) is started before on_open was called. Couldn't this create a problem where the user's message callback might be called before the on_open callback, leading to an unobvious API?

You're right here and your fix looks good.

PR #2601, opened 2026-06-07
RFC 6455: fix WebSocket close frame validation

annihilatorq:
Currently, Glaze does not validate incoming WebSocket close frames, which does not fully comply with RFC 6455 (according to Section 10.7, all data validation is a MUST requirement).

This PR fixes the following issues:
- Validation of payload length (based on Sections 5.5 and 5.5.1)
- Validation of the incoming close-code, which is a MUST requirement that is not explicitly stated in the RFC, so I can't reference a section for it. I've added a comment with explanation in the code for this.
- Validation of the UTF-8 correctness of the incoming close-reason based on 5.5.1.

I've added `fail_connection` because, from what I could tell, the existing `close` function is not entirely suitable for a connection failure, since it waits for a close-frame in response and is intended for a graceful-close, which is probably not the best usecase for a connection failure.

### DIFF
--- a/include/glaze/net/websocket_connection.hpp
+++ b/include/glaze/net/websocket_connection.hpp
@@ -6,7 +6,6 @@
 #include <algorithm>
 #include <array>
 #include <atomic>
-#include <chrono>
 #include <cstring>
 #include <deque>
 #include <functional>
@@ -16,6 +15,7 @@
 #include <string>
 #include <string_view>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 
 // Optional OpenSSL support - detected at compile time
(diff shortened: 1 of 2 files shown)


include/glaze/net/websocket_connection.hpp
```diff
@@ -49,7 +49,10 @@ namespace glz
       policy_violation = 1008,
       message_too_big = 1009,
       mandatory_extension = 1010,
-      internal_error = 1011
+      internal_error = 1011,
+      service_restart = 1012,
+      try_again_later = 1013,
+      bad_gateway = 1014,
```
annihilatorq, line 55: I also added a few missing standard codes that were added after RFC 6455 was written (https://www.iana.org/assignments/websocket/websocket.xhtml WebSocket Close Code Number Registry)

stephenberry:
Thanks for these fixes!

PR #2605, opened 2026-06-08
RFC 9112: fix Host authority in HTTP and WebSocket clients

annihilatorq:
[RFC 9112 Section 3.2](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2-5) states: "If the target URI includes an authority component, then a client MUST send a field value for Host that is identical to that authority component...".

In the case of a URI such as "ws://127.0.0.1:3829/ws", the authority will be "127.0.0.1:3829", so this is what should be set in the Host header.

Currently, Glaze HTTP and WebSocket clients can only connect correctly to hosts running on default ports, as [RFC 9110 Section 4.2.3](https://datatracker.ietf.org/doc/html/rfc9110#section-4.2.3-4.1) specifies that omitting the port in the authority when connecting to a default port (80/443) is a normal form.

I also removed a small section of duplicate code in `http_client.hpp http_client::send_stream_request` by reusing the existing `detail::build_http_request_bytes`

### DIFF
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -273,7 +273,7 @@ namespace glz
          request_str.append("Host: ");
          request_str.append(url.host);
          if (!is_default_port) {
-            char port_buf[5];
+            char port_buf[8]; // a uint16_t port is at most 5 digits; pad so the sizing does not depend on itoa internals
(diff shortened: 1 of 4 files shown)


stephenberry:
Very nice.

PR #2606, opened 2026-06-08
RFC 6455: reject oversized WebSocket control frames

annihilatorq:
As described in [RFC 6455 Section 5.5](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5): "All control frames MUST have a payload length of 125 bytes or less and MUST NOT be fragmented.", currently only close frame payload length is validated, but ping frame with length >125 is still considered valid.


stephenberry:
Thanks!

PR #2610, opened 2026-06-09
Add compile-time diagnostics for object member serialization

wilkolbrzym-coder:
### Summary

This PR introduces member-level compile-time diagnostic `static_assert` checks inside
the object serializer (`to<JSON, T>`) and deserializer (`from<JSON, T>`) template
specializations for reflectable objects and `glaze_object_t` types.

### Problem

When a struct contains a member whose type is not supported by the Glaze serialization
engine (e.g. `std::thread`, a raw socket handle, or any type without a `glz::meta`
specialization or aggregate reflection), the compilation currently fails deep inside
nested template instantiation chains — sometimes hundreds of frames deep — producing
error backtraces that are extremely difficult to interpret, even for experienced C++
developers.

### Solution

A `static_assert` is placed at the top of both `to<JSON, T>` and `from<JSON, T>`
immediately after template instantiation. The assertion iterates over all reflected
fields of `T` at compile time and checks that each one satisfies `write_supported` or
`read_supported` respectively.

Fields that are intentionally excluded from standard serialization are bypassed:
- **Function pointers and member function pointers** (`is_any_function_ptr`) — used
  by RPC registries and invocable endpoints.
- **Always-skipped fields** (`always_skipped`) — fields marked with `glz::skip`,
  `glz::hidden`, or `glz::includer`.

This preserves full correctness for types such as `glz::registry`-registered objects
that expose callable members as RPC endpoints.

### Effect

**Before** (unsupported member type):
```
include/glaze/json/write.hpp:1449:30: error: ...
... (100+ lines of template instantiation noise) ...
include/glaze/util/for_each.hpp:42:5: error: no matching function
```

**After**:
```
include/glaze/json/write.hpp:2130:7: error: static assertion failed:
One of the object's members is not serializable.
Check if member's type has glz::meta or is reflectable.
```

### Changes

| File | Change |
|------|--------|
| `include/glaze/json/write.hpp` | Added `static_assert` to `struct to<JSON, T>` |
| `include/glaze/json/read.hpp` | Added `static_assert` to `struct from<JSON, T>` |

**Total: 22 lines added, 0 lines removed.**

### Compatibility

- Public API (`glz::write`, `glz::read`) signatures are unchanged — SFINAE behavior
  on all public overloads is fully preserved.
- All existing tests pass including `jsonrpc_registry_test` (function pointer members),
  `boost-asio`, `standalone-asio`, `clang-cl`, ARMv7 NEON cross-compile,
  P2996 reflection, and YAML/BEVE/CBOR/MSGPACK format tests.
- Verified on GCC 13/14/15, Clang 17/18/19/20, MSVC (clang-cl), ARMv7 (NEON).

### DIFF
--- a/include/glaze/core/reflect.hpp
+++ b/include/glaze/core/reflect.hpp
@@ -489,6 +489,12 @@ namespace glz
    struct reflect<T>
    {
       static constexpr auto size = 0;
+
+      template <size_t I>
+      using elem = std::nullptr_t;
+
+      template <size_t I>
+      using type = std::nullptr_t;
    };
 
    // The type of the field before get_member is applied
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -2954,7 +2954,7 @@ namespace glz
       static_assert([]() {
          if constexpr (glaze_object_t<T> || reflectable<T>) {
             return []<size_t... I>(std::index_sequence<I...>) {
-               return (read_supported<field_t<T, I>, JSON> && ...);
+               return ((is_any_function_ptr<field_t<T, I>> || always_skipped<field_t<T, I>> || read_supported<field_t<T, I>, JSON>) && ...);
             }(std::make_index_sequence<reflect<T>::size>{});
          }
          else {
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -2952,14 +2952,9 @@ namespace glz
    struct from<JSON, T>
    {
       static_assert([]() {
-         if constexpr (glaze_object_t<T> || reflectable<T>) {
-            return []<size_t... I>(std::index_sequence<I...>) {
-               return ((is_any_function_ptr<field_t<T, I>> || always_skipped<field_t<T, I>> || read_supported<field_t<T, I>, JSON>) && ...);
-            }(std::make_index_sequence<reflect<T>::size>{});
-         }
-         else {
-            return true;
-         }
+         return []<size_t... I>(std::index_sequence<I...>) {
+            return ((is_any_function_ptr<field_t<T, I>> || always_skipped<field_t<T, I>> || read_supported<field_t<T, I>, JSON>) && ...);
+         }(std::make_index_sequence<reflect<T>::size>{});
       }(), "One of the object's members is not deserializable. Check if member's type has glz::meta or is reflectable.");
 
       template <auto Options, string_literal tag = "">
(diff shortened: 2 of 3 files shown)


wilkolbrzym-coder:
I've pushed a couple of minor cleanup commits to optimize the diagnostic assertions and reduce template parsing overhead:

1. **Removed redundant `if constexpr` checks**: The `static_assert` blocks inside both `to<JSON, T>` and `from<JSON, T>` were using an internal `if constexpr (glaze_object_t<T> || reflectable<T>)` branch to guard the fold expression. This check was redundant because the class specializations themselves are already constrained by their respective `requires` clauses.
2. **Simplified map-like reflection**: To safely remove the check from the deserializer `from<JSON, T>` (which also supports `readable_map_t` types that don't have reflected fields), fallback `elem` and `type` template aliases were defined in `reflect<T>` for maps inside `glaze/core/reflect.hpp`. 

This keeps the template structure clean, removes dead branches, and prevents unnecessary compiler overhead during template evaluation. All tests build and pass successfully.

wilkolbrzym-coder:
Closing this one and rebuilding it from scratch, for every format that has an object writer rather than JSON alone.

Two things I found going back over it:

**The assert sits at class scope, which breaks the concept it is built on.** `write_supported` is `requires { to<Format, T>{}; }`, so asking the concept a question instantiates `to<JSON, T>` and fires the assert. On this branch, for a struct with one unsupported member,

```cpp
static_assert(!glz::write_supported<Outer, glz::JSON>);
```

reports

```
error: static assertion failed: One of the object's members is not serializable.
```

instead of answering. The feature probe becomes a hard error rather than `false`, so any generic code that *tests* support to select a different path now fails to compile. The convention against this is already recorded in the library, on the JSONB and YAML variant writers: *"Inside op() rather than at class scope: write_supported is `requires { to<Format, T>{}; }`, so a class-scope assert turns that feature probe into a hard error instead of `false`."* Moving the assert into the body of `op()` keeps the probe queryable, because `op` is a member template and is only instantiated when the writer actually runs.

**The message does not name the member.** "One of the object's members is not serializable" still means bisecting a struct by hand. Without any assert at all the failure is not much worse:

```
include/glaze/json/write.hpp:2729:65: error: incomplete type 'glz::to<10, Opaque>'
used in nested name specifier
```

That at least names the member's type, buried in ~200 lines of instantiation frames. The rebuild names the offending member directly — its index in `glz::reflect<T>`, its key, and the name of its type — in one error at the writer's own line.

The rebuild covers every format with an object writer (JSON, JSONB, BEVE, CBOR, MSGPACK, BSON, YAML, TOML, CSV and EETF, which this one left at seven), and it accounts for the members a format skips by design — `meta::skip`, `always_skipped`, and function pointers — so it cannot fire on a field the format intentionally ignores. I will open it as a new pull request when it is ready.

PR #2614, opened 2026-06-10
Add cmake option to optimize for size

Oipo:
On embedded targets there is usually very limited flash available. The itoa 40kb tables cost 40kb of flash and provide a speedup for integer conversion, but json is almost never in the critical path for embedded code and saving flash becomes a more important need.

glaze_OPTIMIZE_SIZE when set to 1 doesn't include the 40kb header.

Tested by running unit tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,3 +108,8 @@ if(glaze_ENABLE_REFLECTION26)
   target_compile_definitions(glaze_glaze INTERFACE GLZ_REFLECTION26=1)
 endif(glaze_ENABLE_REFLECTION26)
 
+option(glaze_OPTIMIZE_SIZE "Optimize for size by not including itoa tables worth 40kb at the expense of throughput" OFF)
+
+if(glaze_OPTIMIZE_SIZE)
+  target_compile_definitions(glaze_glaze INTERFACE GLZ_OPTIMIZE_SIZE=1)
+endif(glaze_OPTIMIZE_SIZE)
--- a/include/glaze/core/write_chars.hpp
+++ b/include/glaze/core/write_chars.hpp
@@ -30,7 +30,9 @@
 #include "glaze/core/opts.hpp"
 #include "glaze/util/dump.hpp"
 #include "glaze/util/itoa.hpp"
+#ifndef GLZ_OPTIMIZE_SIZE
 #include "glaze/util/itoa_40kb.hpp"
+#endif
 #include "glaze/util/simple_float.hpp"
 #include "glaze/util/zmij.hpp"
 
@@ -351,8 +353,12 @@ namespace glz
                ix += size_t(end - start);
             }
             else {
+#ifndef GLZ_OPTIMIZE_SIZE
                // Normal mode: use to_chars_40kb (40KB digit_quads table)
                const auto end = glz::to_chars_40kb(start, value);
+#else
+               const auto end = glz::to_chars(start, value);
+#endif
                ix += size_t(end - start);
             }
          }
@@ -366,8 +372,12 @@ namespace glz
                ix += size_t(end - start);
             }
             else {
+#ifndef GLZ_OPTIMIZE_SIZE
                // Normal mode: use to_chars_40kb
                const auto end = glz::to_chars_40kb(start, static_cast<X>(value));
+#else
+               const auto end = glz::to_chars(start, static_cast<X>(value));
+#endif
                ix += size_t(end - start);
             }
          }


stephenberry:
Thanks for the PR, I'm wondering if you're aware of Glaze's size optimizations options already available:

Glaze has a first-class `optimization_level` option, documented here:

- [Optimization Levels](https://stephenberry.github.io/glaze/optimization-levels/)
- [linear_search option](https://stephenberry.github.io/glaze/options/#linear_search)

The `size` level already does what this PR is reaching for, and more:

```cpp
glz::write<glz::opts_size{}>(obj, buffer);
glz::read<glz::opts_size{}>(obj, buffer);
```

At that level Glaze drops the 40KB integer `digit_quads` table (uses `glz::to_chars`), **also** drops the ~16KB float pow-10 tables (`to_chars<T, OptSize=true>`), and defaults to linear key search. So `GLZ_OPTIMIZE_SIZE` only captures the ~40KB integer portion and leaves the float/hash savings on the table.

A couple of questions to understand the motivation:

1. Have you tried `glz::opts_size{}` (or a custom opts struct with `optimization_level::size`)? Does it not cover your case?
2. The one real gap I see: `digit_quads` is `inline constexpr`, so it's only dropped if no path uses default opts *and* the linker GCs unused data (`-ffunction-sections -fdata-sections -Wl,--gc-sections`). Is the goal to **guarantee** the table is never linked regardless of stray default-opts call sites? That's legitimate, but I'd rather solve it with a build flag that flips the *default* optimization level to `size` (reusing existing machinery) than a parallel `#ifdef` path that only handles integers and diverges from the documented model.

Mainly I want to avoid two partially-overlapping mechanisms for the same concept. Could you share your constraints (which path you're serializing, and whether `opts_size` was tried)?

Oipo:
Hey Stephen,

Yes indeed, I am aware of the `optimization_level` option and use it everywhere in my own code. However, forgetting to use the `optimization_level` in even one of the calls would result in it being included. Catching this accident is hard, as there is no warning or way to detect it easily, other than constant reviewing.

But the bigger problem I have, and why `optimization_level` in its current form is insufficient, is that `glz::format_error(error_ctx, buffer)` uses the default `opts` in `generate_error_string`, making it a moot point. Even with the flags you mentioned, which I also use as it's an embedded project, that means it will not be discarded by the linker.

Here's a common pattern I have in my code:

```c++
struct voltsoft_glaze_opts {
    glz::optimization_level optimization_level = glz::optimization_level::size;

    uint32_t format = glz::JSON;
    bool null_terminated = false;
    bool comments = false;
    bool error_on_unknown_keys = false;
    bool skip_null_members = true;
    bool prettify = false;
    bool minified = false;
    bool error_on_missing_keys = true;

    bool partial_read = false;

    // INTERNAL OPTIONS
    uint32_t internal{}; // default should be 0

    [[nodiscard]] constexpr bool operator==(const voltsoft_glaze_opts &) const noexcept = default;
};

MyJsonStruct mjs{};
auto ec = glz::read<voltsoft_glaze_opts{}>(mjs, input);
if(ec) {
    std::string error_msg = glz::format_error(ec, input);
    voltsoft::println_compiled<256>(FMT_COMPILE("[{}] error parsing json length {}: {}"), __LINE__, json_size, error_msg);
    send_error_response = UsbResponseError::MALFORMED;
    continue;
}
```

I thought that having a compile-time option to permanently disable the 40kb tables would be preferable because of this. I understand that having two mechanisms to achieve related things is less maintainable. I would be open to increasing the scope of this PR, perhaps to also deal with the float side of things.

Another way to approach this would be to have `glz::format_error` also accept my own opts struct, though that still doesn't deal with the accidental forgetting part.

Perhaps it is an idea to have read and write overloads which re-use the same `glx::context` while parsing and track info, being able to use the given opts struct as well as not needing to re-iterate the given input buffer?

stephenberry:
Thanks for your detailed response. I see the issue in that `glz::format_error` ends up needing the larger tables because it doesn't consider the size opts. We could just always use the smaller table version for `glz::format_error`, because this function isn't designed for absolute peak performance. I'm leaning towards this solution.

Would this be a sufficient solution for you?

Oipo:
For me, personally, it would solve half of my point, the other half being the fragile nature of it leading to accidental unwarranted inclusions of the 40kb tables again. I'm not sure if `glz::format_error` doesn't have the same desire for high throughput in some scenarios. In non-embedded settings I can imagine APIs dealing with wrong json input constantly and wanting to give proper feedback or logs.

stephenberry:
@Oipo, I see where you're coming from and your point is valid. I'm going to think on it a bit more, but I think adding a global CMake option like this is good.

Oipo:
If there's anything I can do, let me know.

If it helps, here's my thoughts. I see two routes we can take that lead to a clear non-overlapping solution:

1. Add a template parameter to `glz::format_error` so that it can use those as opts. This makes it more consistent with the current approach Alternatively:
2. Add a cmake option to disable the int & float tables entirely so those cannot accidentally be used, regardless of opts given or not. No change to `glz::format_error` API needed, only implementation.

stephenberry:
Thanks again for the detailed report — the "accidental forgetting" framing and the `format_error` footgun were exactly the right things to point at.

I went with your route 2 (a build-wide switch), but implemented through the existing `optimization_level` machinery rather than a separate `#ifdef` path, so it covers the integer **and** float tables **and** per-struct hash tables in one flag: **#2633**.

How it works:
- `cmake -Dglaze_DEFAULT_OPTIMIZATION_SIZE=ON` (or `#define GLZ_DEFAULT_OPTIMIZATION_SIZE`) flips the *default* optimization level to `size`. Any call that doesn't specify a level — including a bare `glz::write(obj, buf)` — now resolves to `size`, so a forgotten call site can't relink the 40 KB integer / ~16 KB float / hash tables.
- The only way back to the fast tables is an explicit `optimization_level::normal`, which is deliberate and greppable, not "forgetting."
- I also fixed the two internal leaks: `format_error` now formats with size-optimized integer conversion, and a BEVE float write path that was hardcoding normal opts.

It's value-based (no preprocessor branches in the serialization code), so it also stays clean for a future C++20 module interface. There's an `nm`-based test asserting a size-default binary doesn't link the 40 KB table, so this can't silently regress.

Oipo:
Ah super, that sounds like a good solution! I think I should close this PR then. Thanks a lot!

stephenberry:
I merged the option into `main`. Thanks for this feedback!

PR #2615, opened 2026-06-10
Add WebSocket autobahn client testsuite

annihilatorq:
[Autobahn](https://github.com/crossbario/autobahn-testsuite) is used to test the correctness of WebSocket protocol implementations by industry giants such as Mozilla Firefox, Google Chrome, WebSocket++, Java-WebSocket, Boost-Beast, Facebook Tornado, [and others](https://github.com/crossbario/autobahn-testsuite#users). This test will be extremely helpful in preventing regressions during future development of the client-side WebSocket component of Glaze. Fortunately, at this point, autobahn detects only 4 non-strict cases related to UTF-8 stream validation for incoming continuation frames.

A non-strict case according to autobahn is:

> Test case was executed and passed non-strictly. A non-strict behavior is one that does not adhere to a SHOULD-behavior as described in the protocol specification or a well-defined, canonical behavior that appears to be desirable but left open in the protocol specification. An implementation with non-strict behavior is still conformant to the protocol specification.

... and all other test cases have a "Pass" status, which is certainly good news.

Passing all of the autobahn tests is a strong indicator that the implementation is likely trustworthy, which we can mention in the README in the future, so I think this is another reason why it will be useful for Glaze.

Let me know if you need to add an autobahn fuzzingclient workflow to validate the correctness of the `glz::websocket_server` implementation, or if you need a PR with fix for the "Non-strict" behavior section that arose in 6.4, I’d be happy to help.

### DIFF
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -58,7 +58,7 @@ jobs:
              "url": "ws://127.0.0.1:9001",
              "outdir": "/reports",
              "cases": ["*"],
-             "exclude-cases": [],
+             "exclude-cases": ["9.*", "12.*", "13.*"],
              "exclude-agent-cases": {}
           }
           JSON
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -24,6 +24,7 @@ jobs:
       CC: gcc-14
       CXX: g++-14
       AUTOBAHN_AGENT: glaze-websocket-client
+      AUTOBAHN_CASE_COUNT_FILE: ${{ github.workspace }}/autobahn-case-count.txt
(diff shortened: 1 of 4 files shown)


stephenberry:
Thanks for this! It would be good to move to a more strict websocket implementation as much as possible. This is helpful for confirming compliance.

annihilatorq:
Glad to help! Are we planning to address 6.4? As far as I know, a proper UTF-8 stream validator shouldn't add a lot of overhead compared to validating a complete buffer. If a fix is planned, I can put together a PR (~within a day) and try to make it as efficient as I can. I'd especially appreciate your review on the performance side, since that's definitely your forte.

stephenberry:
@annihilatorq, yeah, a proper UTF-8 stream validator would be great. I would be happy to review and optimize a PR!

PR #2617, opened 2026-06-11
Update workflows to ubuntu-26.04

DockedFerret800:
This pull request updates workflows to `ubuntu-26.04` and compilers and dependencies to the latest versions. The PR also removes some unnecessary workflows that can be removed thanks to natively installed compilers on the runner. I've slightly changed readme to show compiler support information more clearly and placed it before example, so it's more noticeable. Finally, I've added a dependabot GitHub Actions check in order to prevent some dependencies being severely outdated in the future.


DockedFerret800:
@stephenberry Hi Stephen. 

Just wanted to ask your opinion on replacing handwritten CMake calls with [lukka/run-cmake](https://github.com/lukka/run-cmake). This is a popular action that provides many advantages. Some advantages are code annotations, environment setup when necessary, automatic dumping of log files created by CMake, etc. The biggest advantage for us is that we can reduce and replace a large amount of handwritten command-line calls with cleaner presets, which would obviously improve readability and maintainability.

DockedFerret800:
We use it already in UT - https://github.com/openalgz/ut/blob/main/.github/workflows/modules.yml.

stephenberry:
I'd rather not use [lukka/run-cmake](https://github.com/lukka/run-cmake), and I don't think it would remove much redundancy and complexity. I try to keep Glaze's dependencies as small as possible. That said, updating workflows is good, and I'll look at this in more detail.

CramBL:
> removes some unnecessary workflows that can be removed thanks to natively installed compilers on the runner.

This is a **liability**.

It makes environment requirements more implicit, and couples the workflow to the specific setup of the runner.

Github (microsoft) changes the tooling of runners without much regard to developers, and they've broken many workflows (speaking from experience) through the years, when they decide to add or remove stuff from their runners.

It's not like they remove random minor stuff, they literally removed the rust toolchain from runners a couple of years ago.

DockedFerret800:
@CramBL I don't understand what you are trying to say in this PR. What Rust has to do with the pull request to the C++ library? This looks like spam. 

GitHub clearly communicates the differences between images, for example, see the recent announcement post for [Ubuntu 26](https://github.com/actions/runner-images/issues/14226). 
Previous runners also remain available. Newly released images don't replace older ones immediately, there is usually a beta period, followed by a migration period before the newest runner is distributed under the `-latest` label. If you do not want this behavior, you can pin the runner version directly, e.g. `ubuntu-26.04`. 

That said, I still don't understand what you try to say here.

CramBL:
> @CramBL I don't understand what you are trying to say in this PR. What Rust has to do with the pull request to the C++ library? This looks like spam.

You're not reading it right, I'm just using it as an example, they've modified **existing** runners before, not just when changing to new ones. And yes they document it but it's still a nuisance.

Do as you please, I just wanted to note it.

PR #2620, opened 2026-06-12
Use size optimization for error formatting

stephenberry:
## Summary

- format diagnostic indices, lines, and columns with `opts_size`
- prevent `glz::format_error` from retaining the 40 KiB integer lookup table
- preserve existing formatted error output

This is a focused follow-up to the size concern discussed in #2614.

## Verification

- `cmake --build build --target json_test -j 4`
- `ctest --test-dir build -R '^json_test$' --output-on-failure`
- dead-stripped size reproducer: text segment reduced from 64 KiB to 16 KiB, with no `digit_quads` or `to_chars_40kb` symbol

### DIFF
--- a/include/glaze/util/validate.hpp
+++ b/include/glaze/util/validate.hpp
@@ -108,14 +108,14 @@ namespace glz
 
          if (info.context.empty()) {
             dump("index ", b, ix);
-            write_chars::op<opts{}>(info.index, ctx, b, ix);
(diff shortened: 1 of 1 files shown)


stephenberry:
Superseded by #2633, which merged this exact fix (formatting diagnostic index/line/column with `opts_size{}` so `glz::format_error` no longer pulls in the 40 KB integer table) into main as part of the broader build-wide default optimization level work. The change is already on main, so closing this.

PR #2624, opened 2026-06-13
guard fixed-size arrays in beve bool and complex readers

uwezkhan:
Repro: read a BEVE bool or complex typed array into a fixed-size std::array smaller than the encoded element count. Writing a std::vector<std::complex<double>> of 8 and reading into std::array<…, 2> trips ASAN with a heap-buffer-overflow WRITE of 128 bytes at the bulk memcpy; the bool path overflows the same way through value[i], and a bool array needs only ceil(n/8) input bytes to drive n writes.

Cause: the typed-array reader resizes for resizable containers, but for a non-resizable target only the numeric branch checks n against value.size(). The bool and complex branches go straight to the write.

Fix: apply the numeric branch's existing non-resizable check (n > value.size()) in the bool and complex branches.

Before: an oversized n writes past the array. After: it reports syntax_error like the numeric path. Exact-size arrays and resizable containers are unaffected.

### DIFF
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -948,6 +948,12 @@ namespace glz
                   value.shrink_to_fit();
                }
             }
+            else {
+               if (n > value.size()) {
+                  ctx.error = error_code::syntax_error;
+                  return;
+               }
+            }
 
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this fix!

PR #2629, opened 2026-06-15
feat: SIMD-accelerated glz::validate_utf8 (NEON/AVX2/SSSE3) with scalar fallback

stephenberry:
## Summary

Makes `glz::validate_utf8` substantially faster, especially for large and multibyte-heavy text, by adding a SIMD UTF-8 range validator with a scalar fallback. Builds on the earlier consolidation in this PR (validate_utf8 backed by a single `utf8_stream_validator` engine).

Two layers:

1. **Scalar path (all sizes < 512 B, and any build without SIMD):** `glz::validate_utf8` runs one pass of `utf8_stream_validator` (`consume` + `complete`). Single decoding engine, no rule duplication.
2. **SIMD path (>= 512 B):** a range/lookup validator (Lemire / simdjson algorithm) implemented for **NEON, AVX2, and SSSE3** in `glaze/simd/utf8.hpp`, dispatched from `validate_utf8`.

The 512 B threshold is chosen so **no input size regresses** — below it, the fixed per-call SIMD block/remainder overhead doesn't pay off, so scalar wins; above it, SIMD wins decisively.

## Performance (Apple M1, NEON, vs the scalar path)

| payload | speedup |
|---|--:|
| 512 B ascii | +64% |
| 1 KiB ascii | +82% |
| 16 KiB ascii | +100% |
| 64 KiB ascii | +107% |
| 4 KiB multibyte-heavy | **+320%** |
| 256 KiB multibyte-heavy | **+318%** |

The multibyte case is the headline: the scalar path validates codepoint-by-codepoint (~2.4 GB/s); the SIMD path validates 16/32 bytes per iteration regardless of content (~10 GB/s).

## Correctness

- Each ISA implementation (NEON, AVX2, SSSE3) was fuzzed against an **independent scalar oracle**: 19M cases per ISA covering random bytes, boundary-clustered bytes, every codepoint class, surrogates, overlong/too-large encodings, and truncations. The x86 paths were executed via Rosetta on the dev machine. **0 mismatches.**
- The integrated `glz::validate_utf8` (with dispatch) was differentially verified on all four configs — NEON, AVX2, SSSE3, x86_64 scalar baseline — across sizes spanning the 512 B threshold. **0 mismatches.**
- A new differential test in `utf8_test.cpp` cross-checks `validate_utf8` against a scalar reference (~300k cases/run, sizes weighted around 512 B), so **every CI architecture verifies its own SIMD path** going forward.
- Existing `utf8_test` (now 18 tests / 227 asserts), `websocket_close_test`, and `websocket_client_test` pass.

## Notes / portability

- Dispatch is **compile-time**, matching glaze's existing `GLZ_USE_*` SIMD convention: AVX2 if `__AVX2__`, else the 128-bit path if `GLZ_USE_SSE2 && __SSSE3__`, else NEON on AArch64, else scalar. A default-baseline x86_64 build (no `-mssse3`/AVX2) uses the scalar path — no runtime CPU dispatch.
- `parse.hpp` now includes `glaze/simd/utf8.hpp` (which pulls the intrinsics header on SIMD targets). Disable with `GLZ_DISABLE_SIMD`.
- `glz::validate_utf8` remains a public, stateless API with unchanged semantics.

### DIFF
--- a/include/glaze/net/websocket_connection.hpp
+++ b/include/glaze/net/websocket_connection.hpp
@@ -986,15 +986,12 @@ namespace glz
                      }
                      utf8_validator_.reset();
                   }
-                  else {
-                     // Validate the complete payload with the faster streaming validator.
-                     // A local instance keeps utf8_validator_ reserved for multi-read and
-                     // fragmented text streams, so there is no shared state to reset here.
-                     utf8_stream_validator validator;
-                     if (!validator.consume(payload, length) || !validator.complete()) {
-                        fail_connection(ws_close_code::invalid_payload, "Invalid UTF-8");
-                        return false;
-                     }
+                  else if (!glz::validate_utf8(payload, length)) {
+                     // validate_utf8 wraps utf8_stream_validator, so a complete single-read
+                     // payload gets the same fast validation without touching the member
+                     // utf8_validator_ reserved for multi-read and fragmented text streams.
+                     fail_connection(ws_close_code::invalid_payload, "Invalid UTF-8");
+                     return false;
                   }
                }
 
--- a/include/glaze/simd/utf8.hpp
+++ b/include/glaze/simd/utf8.hpp
@@ -15,8 +15,11 @@
 #include "glaze/simd/simd.hpp"
 #include "glaze/util/inline.hpp"
 
-// The 128-bit x86 path needs SSSE3 (pshufb/palignr), which the x86_64 baseline (SSE2) lacks.
-#if defined(GLZ_USE_NEON) || defined(GLZ_USE_AVX2) || (defined(GLZ_USE_SSE2) && defined(__SSSE3__))
+// The NEON kernel uses AArch64-only intrinsics (vqtbl1q_u8, vmaxvq_u8); 32-bit ARM (AArch32/armv7)
+// defines __ARM_NEON but lacks them, so it falls back to the scalar validator. The 128-bit x86 path
+// needs SSSE3 (pshufb/palignr), which the x86_64 baseline (SSE2) lacks.
+#if (defined(GLZ_USE_NEON) && (defined(__aarch64__) || defined(_M_ARM64))) || defined(GLZ_USE_AVX2) || \
+   (defined(GLZ_USE_SSE2) && defined(__SSSE3__))
 #define GLZ_HAS_SIMD_UTF8 1
 #endif
 
--- a/include/glaze/simd/utf8.hpp
+++ b/include/glaze/simd/utf8.hpp
@@ -1,0 +1,233 @@
+// Glaze Library
+// For the license information refer to glaze.hpp
+
+#pragma once
+
+// SIMD UTF-8 validation using the range/lookup algorithm popularized by Daniel Lemire and
+// implemented in simdjson ("Validating UTF-8 In Less Than One Instruction Per Byte"):
+// https://arxiv.org/abs/2010.03090. One algorithm, three instruction sets (NEON, AVX2, SSSE3)
+// plus a scalar fallback in parse.hpp. glz::validate_utf8 dispatches here for large buffers.
+
+#include <cstddef>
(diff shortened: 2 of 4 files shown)


annihilatorq:
Hi Stephen, I'm slowly getting started with SIMD and wanted to ask you a question that's been on my mind for a while regarding Glaze and other projects that use SIMD without runtime dispatch. If it's not too much trouble, could you please explain why Glaze uses compile-time dispatch to determine the executable instruction set?

As far as I understand, for server applications this is definitely more useful than runtime dispatch, since you can build the project with `-march=native`, in which case there will be absolutely no overhead on the `if` statements or indirect calls (assuming that the `cpuid` was parsed only once and then cached somewhere), and consequently, there would be no potential branch misses and/or no indirect branch overhead. Still, wouldn't it be useful to have an option to enable runtime dispatch for the SIMD part in Glaze (mainly for projects that run on different CPUs, e.g. desktop applications)?

I understand that this would likely have a slight impact on the instruction cache, degrade (or probably even remove) possibility for inlining, etc., which would result in a slight performance drop. But what about client applications, for example, where it’s unknown which instruction sets are supported? Or am I underestimating the cost of runtime dispatch, and is it actually much more expensive than I think?

stephenberry:
@annihilatorq, I don't think the cost of runtime dispatch would be too large for bulk SIMD operations like UTF8 validation. In my experience, the benefits of `-march=native` go beyond SIMD, and include other instructions like BMI2, that can have additional performance improvements. Glaze allows you to compile without SIMD, and instead uses SWAR (SIMD Within A Register) so you still get high performance if you want to be more flexible. But, if you're targeting peak performance with SIMD then you probably want the other native optimizations as well. I just don't think there is a very big gap between SWAR and native SIMD that justifies the additional complexities of a runtime dispatched SIMD. But, if there's a real use case for it and benchmarks to show for it, I'm always eager to learn.

wilkolbrzym-coder:
## Quick review notes – PR #2629

- **SIMD kernels look solid** – NEON, AVX2, and SSSE3 implementations correctly follow the range‑lookup algorithm from simdjson. The lookup tables and block logic are correct.

- **Good testing** – the differential fuzzing against an independent scalar reference gives strong confidence. 0 mismatches across 19M cases.

- **Portability** – NEON correctly gated to AArch64; builds without SIMD fall back to scalar. The 512‑byte threshold makes sense.

**Minor observations:**

- The scalar fallback lost the SWAR ASCII fast‑path present in the previous version. For medium ASCII buffers (100‑500 bytes) this might be a tiny regression. Not a blocker.

- `websocket_connection.hpp` diff shows an `else { ... } else if` block – probably a formatting artifact, but worth double‑checking that the actual code compiles correctly.

wilkolbrzym-coder (commented):
## Review notes – PR #2629

**Blocking:** branch has merge conflicts and must be rebased before merge.

I did not find a hard correctness bug in the SIMD validator. The implementation is a faithful port of the Lemire/simdjson range algorithm, including incomplete-sequence tracking and tail handling. NEON is correctly gated to AArch64.

A few things to verify, not correctness blockers:

- `parse.hpp` now includes `glaze/simd/utf8.hpp` unconditionally. That may increase compile times for every TU using parse.hpp, even without SIMD. Worth measuring.
- The scalar fallback now uses `utf8_stream_validator` instead of the previous SWAR ASCII fast path. Small ASCII buffers below 512 B may have regressed. A quick benchmark on 64–511 B ASCII would confirm.
- `validate_utf8(nullptr, 0)` must be well-defined in the new scalar path. Please add a tiny test for the zero-length/null pointer case.
- The strong local fuzzing evidence for AVX2/SSSE3 was run under Rosetta. CI covers native x86, so not a blocker, but worth remembering that the differential test is the main cross-platform check.

Once the conflict is resolved and the three verification points are confirmed, this is good to merge from my side.

cc @stephenberry

PR #2637, opened 2026-06-17
guard skip_number against end of non-null-terminated input

uwezkhan:
Repro: read a number-valued JSON pointer over a non-null-terminated buffer that ends right after the digits, e.g. `glz::get_view_json<"/h", opts{.null_terminated=false}>` over `{"h":-0}` with no closing brace. ASAN reports a heap-buffer-overflow read of size 1 in skip_number_with_validation.

Cause: after scanning the digit run, skip_number and skip_number_with_validation peek at `*it` for a trailing `.`/`e`/exponent sign without rechecking `it != end`. The JSON pointer path (parse_value) forces validate_skipped, so it routes through the validating skip. skip_ws only guarantees a byte at the number's start, not after the digits, so a number that ends the buffer over-reads by one byte.

Fix: carry null_terminated through skip_number_opts and bound each post-advance peek against end, the same way skip_ws already handles non-null-terminated input. Behavioral difference is only at the buffer edge: before, a number ending at end peeked past it; after, that number is treated as complete and a dangling sign or exponent is a syntax error. The null-terminated path is unchanged and still relies on the trailing sentinel, so there is no cost there.

### DIFF
--- a/include/glaze/util/parse.hpp
+++ b/include/glaze/util/parse.hpp
@@ -1246,13 +1246,27 @@ namespace glz
       return {};
    }
 
+   // null_terminated == false means the buffer has no trailing sentinel, so each peek that
+   // follows an advance past a digit run is guarded against end. A number whose last digit sits
+   // at end is complete, so those guards return without error; only a dangling sign/exponent is
+   // a syntax error.
+   template <bool null_terminated = true>
    GLZ_ALWAYS_INLINE void skip_number_with_validation(is_context auto&& ctx, auto&& it, auto end) noexcept
    {
       it += *it == '-';
       const auto sig_start_it = it;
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for catching this, and apologies for the overlap. This exact fix landed in #2636 (merged earlier today), which guards the same two functions in `util/parse.hpp`: `skip_number` (gated non-validating loop) and `skip_number_with_validation` (end-bounded peeks). I verified your repro against current `main` under ASAN: it reports the heap-buffer-overflow before #2636 and runs clean after, and the `non_null_terminated_scanner_bounds` suite already covers the validating number-skip boundary. Closing as superseded, but the parallel discovery is much appreciated.

PR #2641, opened 2026-06-17
guard front_hash key lookup against keys shorter than the prefix

uwezkhan:
`decode_hash_with_size` selects a `front_hash` path for key sets distinguished by their first 2, 4, or 8 bytes, and that specialization reads `front_hash_bytes` from the key with `memcpy` without checking the key length, unlike every sibling hash type which guards its read against `end`. BSON, MessagePack, CBOR, CSV, and TOML decode in place and call it with `end = key.data() + key.size()`, so a document key shorter than `front_hash_bytes` reads past the key and off the end of the input buffer. A one-byte key at the tail of a minimal BSON document reads seven bytes past the allocation, which ASAN flags as a heap-buffer-overflow read of size 8.

The guard rejects keys shorter than `front_hash_bytes` before the read and returns the not-found index. `front_hash` is only chosen when every key is at least `front_hash_bytes` long, so before the change a valid key always carried enough bytes, and after it the only keys turned away are ones too short to match any field, which leaves correct lookups unchanged. Putting the bound in the shared hash callee covers all five readers in one place rather than each reader, and it mirrors the `decode_hash<JSON>` `front_hash` variant that already guards against `end`.

### DIFF
--- a/include/glaze/core/reflect.hpp
+++ b/include/glaze/core/reflect.hpp
@@ -2766,8 +2766,16 @@ namespace glz
       static constexpr auto N = reflect<T>::size;
       static constexpr auto bsize = bucket_size(hash_type::front_hash, N);
 
-      GLZ_ALWAYS_INLINE static constexpr size_t op(auto&& it, auto&&, const size_t) noexcept
+      GLZ_ALWAYS_INLINE static constexpr size_t op(auto&& it, auto&& end, const size_t) noexcept
       {
+         // front_hash is only chosen when every key is at least front_hash_bytes long, so a
+         // shorter key cannot match and reading the prefix would run past it. The non-padded
+         // readers (BSON, MessagePack, CBOR, CSV, TOML) pass end = key.data() + key.size(),
+         // so an unguarded read here is an out-of-bounds read of the input buffer. The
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for catching this and for the clean ASAN repro.

The same short-key over-read turns out to affect the sibling key-hash readers too — the sized `unique_index` path, `mod4`/`xor_mod4`/`minus_mod4`, and `three_element_unique_index` (with `unique_index == 0`). So rather than guard `front_hash` alone, I've opened #2643, which pre-screens the key length once in the shared `decode_hash_with_size` entry point: a key whose length is outside `[min_length, max_length]` can't match any field, and rejecting it there bounds every reader (including `front_hash`) in one place. That supersedes the per-reader guard here.

uwezkhan:
Agreed, #2643 is the better fix. Since front_hash_bytes <= min_length, the single [min_length, max_length] screen subsumes the guard here and also bounds the unique_index, mod4, and three_element readers I didn't touch, so one check instead of five per reader. Good call leaving unique_per_length its own end check, since its length-indexed table maps absent lengths to 255 and that can land in a gap the range filter won't reject.

PR #2654, opened 2026-06-19
reject transfer-encoding requests in finish_request

uwezkhan:
The server frames request bodies with Content-Length only and ignores Transfer-Encoding, since chunked request bodies are unsupported. finish_request reaches the Content-Length step with a request that carries Transfer-Encoding and no Content-Length, treats it as a zero-length body, and leaves the encoded payload sitting in the keep-alive buffer. A front-end proxy that does honor Transfer-Encoding forwards the whole thing as one request, so those leftover bytes get parsed here as a second request the client never sent (TE/CL desync, request smuggling).

Before: a Transfer-Encoding request was silently accepted with an empty body and the chunk data was reinterpreted as a pipelined request. After: finish_request answers 501 and closes the connection when a Transfer-Encoding header is present, which RFC 7230 3.3.1 calls for on an encoding the server cannot decode. The check sits next to the Content-Length parsing because that is the layer that frames the body and decides what stays in the buffer; the tradeoff is that a (currently unsupported) chunked client now gets a hard 501 instead of a confusing empty-body 200. Content-Length requests are unaffected.

### DIFF
--- a/include/glaze/net/http_server.hpp
+++ b/include/glaze/net/http_server.hpp
@@ -1801,7 +1801,7 @@ namespace glz
          // Transfer-Encoding cannot be framed by Content-Length. Treating it as a
          // zero-length body leaves the encoded payload in the buffer, where the
          // keep-alive loop parses it as a second, smuggled request (TE/CL desync).
-         // RFC 7230 6.3.3 requires answering an undecodable Transfer-Encoding with
+         // RFC 7230 3.3.1 requires answering an undecodable Transfer-Encoding with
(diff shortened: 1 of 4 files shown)


annihilatorq:
I'd like to note that RFC 7230 doesn't have a section 6.3.3. The section you're talking about is 3.3.1

uwezkhan:
Good catch, that was a typo. It's 3.3.1, where the 501 for an unknown transfer coding lives. Fixed the code comment and the description.

PR #2660, opened 2026-06-20
cap recursion depth in yaml and toml generic value readers

uwezkhan:
The YAML and TOML readers parse a generic value by dispatching on the first byte and recursing back through the variant reader for each nested element. Block-style nesting is bounded by the indent stack, but flow collections (`[ ]`, `{ }`) and TOML arrays never touch that stack, so they recurse one level per opening bracket with nothing counting depth. A document that is a few thousand `[` therefore overflows the call stack and crashes (SIGSEGV) when read into `glz::generic`, which is reachable from any untrusted YAML or TOML input.

Before, flow nesting could only stop at the stack limit. After, the variant reader (the single point every nested generic value passes through) takes the same `depth_guard` the BSON and JSONB readers already use, so a value nested past `max_recursive_depth_limit` returns `exceeded_max_recursive_depth` rather than recursing further. Guarding the variant reader instead of each container parser keeps it to one site per format and also covers the block mappings reached from flow context, which the indent stack alone misses. The cost is a fixed 256 level cap on generic values, the same limit the block parsers and the binary readers already enforce.

### DIFF
--- a/include/glaze/core/context.hpp
+++ b/include/glaze/core/context.hpp
@@ -132,6 +132,32 @@ namespace glz
       { ctx.depth } -> std::same_as<uint32_t&>;
    };
 
+   // RAII guard for recursive descent: bumps ctx.depth on entry and restores it on exit,
+   // erroring out before the nesting reaches max_recursive_depth_limit so adversarial deeply
+   // nested input can't overflow the stack. Mirrors the per-reader guards already used by the
+   // BSON and JSONB binary readers; placed here so the text-format readers can share one copy.
+   template <class Ctx>
+   struct depth_guard
+   {
+      Ctx& ctx;
+      bool entered = false;
+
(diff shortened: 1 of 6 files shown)


stephenberry:
Thanks @uwezkhan — your commit is carried into #2662, which builds on it with depth guards for the TOML dotted-key/table-header and YAML skip paths and unifies `depth_guard` onto a single shared copy. Closing in favor of #2662; your authorship is preserved in the commit history.

uwezkhan:
Sounds good, thanks for folding it in. #2662 covers more of the recursive paths than my version did, the dotted-key/table-header descent and skip_flow_content were both reachable and I'd only guarded the variant choke point. Collapsing onto one shared depth_guard is cleaner too. Happy with this landing in #2662.

PR #2674, opened 2026-06-24
enforce rfc 6455 frame masking in process_frames

uwezkhan:
process_frames unmasks a frame when the mask bit is set, but it never checks that the bit had to be set. RFC 6455 Section 5.1 requires a server to fail the connection on any unmasked frame from a client, and a client to fail on any masked frame from a server. As written, a peer can send an unmasked data frame (0x81 0x02 'H' 'i') and the server unmasks nothing, then passes the payload to handle_frame and on to the message handler.

The check sits next to the existing reserved-bits guard and keys on client_mode_, so one comparison covers both roles, and it fails the connection with 1002 before the unmask loop or any dispatch runs. Before, an unmasked frame reached the application; after, it is rejected at the header stage with the other framing violations process_frames already handles. The tradeoff is strictness over leniency: a client that skips masking is now dropped rather than served, which is what the spec requires.

### DIFF
--- a/include/glaze/net/websocket_connection.hpp
+++ b/include/glaze/net/websocket_connection.hpp
@@ -865,6 +865,15 @@ namespace glz
                return length; // Consume all
             }
 
+            // RFC 6455 Section 5.1: a server MUST close the connection upon receiving an unmasked
+            // frame from a client, and a client MUST close upon receiving a masked frame from a server.
+            // https://datatracker.ietf.org/doc/html/rfc6455#section-5.1
+            if (header.mask() == client_mode_) {
+               fail_connection(ws_close_code::protocol_error,
+                               client_mode_ ? "Masked frame from server" : "Unmasked frame from client");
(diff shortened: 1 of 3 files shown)


stephenberry:
Thanks for all these fixes! You're awesome!

uwezkhan:
Glad these help. This one's a small masking check at the header stage keyed on client_mode_, so it covers both roles with a single comparison and should be a straightforward merge whenever you get to it.

annihilatorq:
You forgot about the test coverage for `glz::websocket_client`. Also, the comment above `make_unmasked_client_frame` mostly restates the name, and its parenthetical explains why you're calling it in your test rather than what the function itself does. Otherwise, LGTM.

It's a small change and I'd like to save Stephen's time he'd spend on finishing the tests and reviewing them, so I'll add the missing websocket_client coverage myself and just merge it right away.

PR #2676, opened 2026-06-25
RFC 9112: validate Host header in HTTP server request parser

annihilatorq:
## RFC 9112, Section 3.2:

> A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field and to any request message that contains more than one Host header field line or a Host header field with an invalid field value.

- [x] A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that lacks a Host header field.
- [x] A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that ... contains more than one Host header field line.
- [x] A server MUST respond with a 400 (Bad Request) status code to any HTTP/1.1 request message that ... contains a Host header field with an invalid field value.

---

This PR partially addresses the issue with MUST validation of Host headers, but even with this PR (at the current stage) Glaze still does not validate the Host header field value as required by RFC 9112.

I haven't implemented this as part of the PR yet, since I haven't gotten around to implementing it in my project.
I'm currently writing a project that implements a HTTP/1.1 server according to the RFC as strictly as I can, and I plan to check after each implemented part and after reading the RFC whether Glaze meets the MUST or SHOULD requirements. If not, I will fix it or at least notify you of any non-compliance. Of course, it's impossible to guarantee that I won't miss a single HTTP/1.1 server detail, but I hope this will at least partially help in achieving the goal of full RFC compliance.

If you don't consider this a urgent issue and don't plan to address it yourself - I plan to write such validator for my project within the next couple of days or so, and I'll be able to port it to Glaze if it's still relevant by then.

### DIFF
--- a/.github/workflows/boost-asio.yml
+++ b/.github/workflows/boost-asio.yml
@@ -77,6 +77,8 @@ jobs:
             http_client_test \
             http_client_pool_test \
             http_server_api_tests \
+            http_server_headers_validation_test \
+            ip_address_validation_test \
             https_test \
             jsonrpc_registry_test \
             openapi_test \
@@ -93,4 +95,4 @@ jobs:
 
       - name: Test
         working-directory: build
-        run: ctest -j $(nproc) --output-on-failure -R "^(asio_repe|http_examples|http_router_test|http_client_test|http_client_pool_test|http_server_api_tests|https_test|jsonrpc_registry_test|openapi_test|repe_buffer_test|repe_plugin_test|repe_test|registry_view_test|repe_to_jsonrpc_test|websocket_close_test|websocket_client_test|utf8_test|shared_context_bug_test|wss_test)$"
+        run: ctest -j $(nproc) --output-on-failure -R "^(asio_repe|http_examples|http_router_test|http_client_test|http_client_pool_test|http_server_api_tests|http_server_headers_validation_test|ip_address_validation_test|https_test|jsonrpc_registry_test|openapi_test|repe_buffer_test|repe_plugin_test|repe_test|registry_view_test|repe_to_jsonrpc_test|websocket_close_test|websocket_client_test|utf8_test|shared_context_bug_test|wss_test)$"
--- a/.github/workflows/standalone-asio.yml
+++ b/.github/workflows/standalone-asio.yml
@@ -64,6 +64,8 @@ jobs:
           http_router_test \
           http_client_test \
           http_server_api_tests \
+          http_server_headers_validation_test \
+          ip_address_validation_test \
           https_test \
           jsonrpc_registry_test \
           openapi_test \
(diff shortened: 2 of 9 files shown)


annihilatorq:
I also added a new test directory, `http_server_test`, since I couldn't find a directory where these tests would fit. The `http_smuggling_test` test checks for something similar about content-length header, but directory is still about smuggling, not the correctness of the HTTP/1.1 request headers validation. The same applies to `http_header_strictness_test`, since it checks not so much the correctness of HTTP/1.1 request according to RFC 9110 and RFC 9112, but rather the parsing of headers according to RFC 7230.

I think this directory and these tests will continue to grow in the future, but if the idea of a new directory for these tests seems like an overkill, I'm totally fine with moving the tests to an existing directory.

stephenberry:
I think a new directory is good. Any test organization you want to do is great!

annihilatorq:
@stephenberry, sorry for the delay, it took me some time to implement IPv4/IPv6 address validators, it turned out to be a bit harder than I expected 😁

The implementation has some deliberate limitations:
1. Current hostname validator supports only ASCII hostnames, so any unicode hostname will be rejected.
2. No support for percent-encoding. It's likely worth implementing this in the future, since, as far as I know, this is a fairly common feature. If implementation is planned, security is definitely should be the top priority (see curl [CVE-2022-27780](https://github.com/advisories/GHSA-82rv-h33p-2xgc)).
3. No support for IPvFuture to avoid increasing the validator's complexity and sacrificing performance. For example, even curl does not support IPvFuture, so I'm not sure whether we need this or not.
4. [RFC 3986 Section 3.2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.3), specifies `port = *DIGIT` without mentioning any restrictions on the port value itself, but it would be logical to reject any port larger than 65535 (the maximum port for TCP).

# IPv4/IPv6 validation

At first, I thought about using `asio::ip::make_address_v4/v6` to validate the received IP address within the uri-host part inside Host header value, so I went to check how this function works and was a little disappointed to find out that:

1. For `std::string_view` overload, asio copies the string to a `std::string`. Although most IPv4 addresses in most STL implementations will fit into an SSO, this isn't guaranteed, and a potential attacker could start sending large host header values, leading to unnecessary memory allocations for each request.
2. `asio::ip::make_address_v4/v6` internally uses `asio::detail::socket_ops::inet_pton`, which calls a good number of system and CRT functions. For example, the path for Windows looks like this:
```
strlen
_alloca
MultiByteToWideChar
WSAStringToAddressW
WSAGetLastError
memcpy
```
It's quite clear that most calls in such a chain come at a cost. Of course, I'm not saying that `asio::ip::make_address_v4/v6` is a poorly implemented function, but the fact that it isn't suitable for our scenario in terms of performance seems quite obvious.

Since performance is important for Glaze, I decided to implement an IPv4 and IPv6 validator that would be platform-independent, wouldn't allocate memory, and wouldn't use any heavy operations.

## IPv4/IPv6 address validation benchmarks
I'll back up what I said above with some benchmarks that I ran on my local machine.
```
CPU: AMD Ryzen 5 5600 6-Core (3.50 GHz)
Compiler: clang-cl, x64, -O3
```
| Payload | `asio::make_address_v4/v6` | `glz::detail::validate_ipv4/ipv6_address` |
|---|---:|---:|
| IPv4 valid literals 228 KiB | 20.3 MB/s | **1,030 (+4,968.9%) MB/s** |
| IPv4 invalid literals 216 KiB | 14.4 MB/s | **1,354 (+9,270.6%) MB/s** |
| IPv4 mixed literals 376 KiB | 24.5 MB/s | **1,202 (+4,809.7%) MB/s** |
| IPv6 valid literals 360 KiB | 25.5 MB/s | **547 (+2,045.2%) MB/s** |
| IPv6 invalid literals 396 KiB | 23.1 MB/s | **616 (+2,563.2%) MB/s** |
| IPv6 mixed literals 556 KiB | 28.1 MB/s | **611 (+2,073.5%) MB/s** |

stephenberry:
Thanks for all your effort and optimization for this! Thanks for the detailed explanation, limitations, and comments on performance.

annihilatorq:
@stephenberry, if everything looks fine to you, then I think we can merge it. If you have any issues with the current hostname validation limitations, let me know and I'll tune the implementation

annihilatorq:
@stephenberry, I also have a few follow-up HTTP parser changes that depend on this PR. Would you prefer me to open them as stacked PRs, or should I wait until this one is merged and then open the next PRs based on main?

stephenberry:
I would build new PRs off this branch or wait until this is merged. I plan to look at it this weekend.

stephenberry:
This is looking good to me! I made a few tweaks, you can see my recent pushes. Merge if you think it's all good.

annihilatorq:
Yeah, the fixes look great and totally make sense! Especially with OWS - for some reason, I didn't notice that while reading the RFC.

PR #2683, opened 2026-07-01
lazy JSON: opt-in streaming cursor for read_into + sequential iteration

psiha:
## Motivation

Large **nested tabular JSON** — a top-level catalog of entity arrays, each with a `rows`
matrix of scalar cells — is dominated by **sequential bulk ingest**: walk every row,
`read_into` every cell. This PR fixes a correctness bug that made a lazy optimization
silently inert, then adds two ingest optimizations on top.

### 1. Fix: custom `opts` are silently ignored by the lazy JSON API

The lazy-JSON types (`lazy_document`, `lazy_json_view`, `lazy_iterator`, `indexed_lazy_view`,
the streaming-slot selector, and the `skip_*` helpers) were declared `template <opts Opts>`
— a concrete, base-class NTTP. A user options struct that **derives from `glz::opts`** to
carry extra fields — the pattern the header comments recommend ("*Add these fields to a
custom options struct if you want to use them*") — is **sliced to `glz::opts`** at that
boundary, dropping every derived field. So any `if constexpr (requires { Opts.my_field; })`
inside the lazy code evaluated false. The rest of Glaze's read/write API already uses
`template <auto Opts>` (which preserves the exact type); this PR aligns the lazy surface
with that convention.

### 2. Opt-in streaming cursor

`lazy_streaming_cursor_policy` (default `disabled`): `read_into` and finished container
iterators record the consumed byte extent `[start, end)`; the next `operator++` jumps to
`end` instead of re-scanning with `skip_value_lazy`. Zero storage / no layout change when
disabled. **Because of the slicing bug above, the cursor never actually compiled in before
this fix** — every prior "cursor on/off" comparison was two identical builds.

### 3. SIMD structural skip (expanded from SWAR)

`skip_value_lazy` / `skip_to_depth_zero` locate the next JSON structural byte
(`" [ ] { }`) to advance past scalar runs. This started as an 8-byte SWAR pre-scan; it is
now a **32-byte portable SIMD scan built on GCC/Clang vector extensions**, which the
compiler lowers to **AVX2/SSE2 on x86 and NEON on ARM** with no intrinsics. The first
match is located via a memcpy to 4×`uint64` + `countr_zero`, so it stays portable across
clang and gcc (no target-specific `movemask`). MSVC's `cl.exe` (no vector extensions)
falls back to the SWAR loop; the scan is skipped during constant evaluation. Gated by an
opt-out `lazy_swar_skip` (default on) for A/B isolation.

### 4. Opt-in slim `lazy_json_view` (48 → 24 bytes)

`lazy_json_view` carries a per-value object key (`std::string_view`, 16 B) and an
`error_code`, even though the key is only meaningful while iterating an object and the
error only tags views returned by fallible lookups. On 64-bit this makes every value view
48 bytes; a consumer that copies the view per cell pays for those 24 bytes on every copy.

`lazy_slim_view` (default off) drops both from the value, shrinking it to 24 bytes (the
three pointers). The object key moves onto the iterator (`lazy_iterator::key()` /
`indexed_lazy_iterator::key()`), and error state is signalled by a null value pointer
(`has_error() == (data() == nullptr)`; the specific `error_code` is not retained in this
layout). The full 48-byte layout stays the default, so existing users and the whole test
suite are unaffected; the key/error storage is selected with the same
empty-specialization + `[[no_unique_address]]` pattern used for the streaming slot, so the
dropped members occupy zero space under the policy.

### Streaming cursor on vs off — `lazy_load_benchmark` (Release)

| Fixture | Windows (Core Ultra 7) | Mac (Apple M1) | Linux (Clang) |
|---------|----------------------:|---------------:|--------------:|
| Mixed-warehouse ~2.2 GB (rotating scalar/numeric/object rows) | **+10%** | **+6.5%** | **+10%** |
| Split-catalog scalar-cell load ~12 MB | **+18%** | **+39%** | **+22%** |
| Flat tables smoke ~3.8 MB | **+28%** | **+39%** | **+35%** |

Object records in the mixed fixture are consumed key-by-key via `operator[]` (which the
cursor does not accelerate), so the mixed number reflects only the row shapes that benefit.

### SIMD structural skip vs 8-byte SWAR (same fixtures, cursor on)

The vector scan lifts absolute throughput a further **~5–14%** over SWAR (e.g. scalar-cell
load 307→323 MB/s, flat 411→443 MB/s), and shaves **~3–4% CPU** off a downstream
multi-GB string-heavy import. Full lazy test suite: 70/70 at both AVX2 and SSE2 widths.

### Downstream: multi-GB import, cursor + SIMD vs an external parser (simdjson)

An analytics engine (Farseer/Rama) parsing multi-GB dumps into columnar storage. usr+krn
CPU, median of interleaved runs, Windows, Release; four builds from one tree:

| Import | simdjson | vanilla glaze | +cursor | +cursor +SIMD |
|--------|---------:|--------------:|--------:|--------------:|
| 325 MB mixed | 8390 | 8655 | 8405 | **7906** |
| 3.4 GB string-heavy | 32734 | 33985 | 32311 | **30296** |

Vanilla lazy trails simdjson (+3–9%); cursor + SIMD turn that into a **−6% / −7% win** on
the large imports that dominate load time. (Peak RSS is also ~halved vs simdjson,
though that's a property of lazy parsing generally, not this PR.)

### Slim view: measured throughput, and a clang-cl caveat

Benchmarked on the realistic `read_into`-per-cell ingest (GB-scale flat cell-heavy fixture),
full 48-byte vs slim 24-byte view **with the streaming cursor on**, interleaved
separate-process runs, pinned:

| toolchain / target | slim vs full (streaming on) |
|---|---|
| Apple clang, arm64 | **+8%** (slim faster) |
| Linux clang, x86-64 | ≈ neutral (+1.4%, noise) |
| Windows clang-cl, x86-64 | **−26%** (regression) |

Without the streaming cursor, slim and full are within noise on all three. So the slim view
is a modest win or neutral everywhere **except clang-cl + streaming cursor**, where it
regresses ~26% on cell-access-bound ingest. VTune attributes that to L1 store-ordering stalls
(elevated *Store Early Blk* / *L1 Latency Dependency*) — not DRAM/L3-bound, not a
store-forwarding violation, and **not a struct-layout issue** (padding `lazy_document` to keep
the streaming slot's offset fixed does not move it). Linux clang on the *same hardware* is
fine, so this is a **clang-cl-specific codegen artifact, not algorithmic**. It is documented
as a known limitation in `docs/lazy-json.md`, and the slim view stays **off by default**.

## Notes / test plan

- [x] Lazy JSON unit tests pass with the streaming cursor **actually enabled** (the
  `auto Opts` fix — previously these exercised the disabled path only); added an
  out-of-order-key regression test on an end-bounded + cursor opts.
- [x] SIMD skip verified at AVX2 and SSE2 widths (70/70); MSVC/constexpr fall back to SWAR.
- [x] Downstream import parity: force-reimport + recalc, 0 variables diverged vs reference.
- [x] Slim view: full lazy suite 70/70 (+ BEVE 57/57) on the default (full) layout; slim
  produces byte-identical `read_into` results (checksums match full across all fixtures);
  throughput measured on arm64 / Linux-x86 / clang-cl-x86 (see the slim-view caveat above).
- Default remains `lazy_streaming_cursor_policy::disabled`; the `auto Opts` change is
  source-compatible with existing `glz::opts{}` call sites.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -35,6 +35,29 @@ target_compile_options(lazy_benchmark PRIVATE
   $<$<CXX_COMPILER_ID:MSVC>:/O2>
 )
 
+# Tabular bulk-load pattern (nested arrays-of-arrays + scalar read_into)
+add_executable(lazy_load_benchmark lazy_load_benchmark.cpp)
+target_link_libraries(lazy_load_benchmark PRIVATE
+  glaze
+  bencher::bencher
+)
+target_compile_options(lazy_load_benchmark PRIVATE
+  $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-O3 -march=native>
+  $<$<CXX_COMPILER_ID:MSVC>:/O2>
+)
+
+# Header-scan stage on large split-catalog fixture (skip/iterator guard; no row read_into loop)
+add_executable(lazy_load_scalar lazy_load_scalar.cpp)
+target_compile_definitions(lazy_load_scalar PRIVATE GLAZE_LAZY_LOAD_FORK)
+target_link_libraries(lazy_load_scalar PRIVATE
+  glaze
+  bencher::bencher
+)
+target_compile_options(lazy_load_scalar PRIVATE
+  $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-O3 -march=native>
+  $<$<CXX_COMPILER_ID:MSVC>:/O2>
+)
+
 # Simple float benchmark
 add_executable(simple_float_benchmark simple_float_benchmark.cpp)
 target_link_libraries(simple_float_benchmark PRIVATE
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -46,6 +46,14 @@ target_compile_options(lazy_load_benchmark PRIVATE
   $<$<CXX_COMPILER_ID:MSVC>:/O2>
 )
 
+# Single-config GB-scale ingest, selected by argv, for interleaved separate-process A/B
+add_executable(flat_ingest_single flat_ingest_single.cpp)
+target_link_libraries(flat_ingest_single PRIVATE glaze)
+target_compile_options(flat_ingest_single PRIVATE
+  $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-O3 -march=native>
+  $<$<CXX_COMPILER_ID:MSVC>:/O2>
+)
+
 # Header-scan stage on large split-catalog fixture (skip/iterator guard; no row read_into loop)
 add_executable(lazy_load_scalar lazy_load_scalar.cpp)
 target_compile_definitions(lazy_load_scalar PRIVATE GLAZE_LAZY_LOAD_FORK)
--- a/benchmarks/flat_ingest_single.cpp
+++ b/benchmarks/flat_ingest_single.cpp
@@ -1,0 +1,101 @@
+// Single-config GB-scale flat-tables ingest, selected by argv[1], for interleaved
+// separate-process A/B measurement (avoids the sequential-arm ordering/thermal confound of
+// running all four configs in one process). Prints one line: "<cfg> cpu_ms=<n> MB/s=<n>".
+#include <chrono>
+#include <cstdio>
+#include <cstring>
+#include <string>
+
+#include "lazy_load_common.hpp"
+
+namespace
+{
+   using namespace lazy_load_bench;
+
+   template <auto Opts>
+   long timed_ingest(const std::string& json, double& ms)
+   {
+      auto doc = glz::lazy_json<Opts>(json);
+      if (!doc) {
+         std::abort();
+      }
+      const auto t0 = std::chrono::steady_clock::now();
+      const auto sum = load_flat_tables_scalars<Opts>((*doc)["tables"]);
+      const auto t1 = std::chrono::steady_clock::now();
+      ms = std::chrono::duration<double, std::milli>(t1 - t0).count();
+      return static_cast<long>(sum);
+   }
+
+   // Models RAMA config A: iterate with the full-view iterator, but per cell round-trip the
+   // value through a 24-byte {doc,data,parse_pos} handle and reconstruct a FULL view to read
+   // through (as RAMA's makeValue/GlzViewHandle::view() does). Opts must be a FULL-view opts.
+   template <auto Opts>
+   long timed_ingest_recon(const std::string& json, double& ms)
+   {
+      auto doc = glz::lazy_json<Opts>(json);
+      if (!doc) {
+         std::abort();
+      }
+      struct SlimHandle
+      {
+         const glz::lazy_document<Opts>* d{};
+         const char* data{};
+         const char* pp{};
+      };
+      long sum = 0;
+      std::size_t nulls = 0;
+      const auto t0 = std::chrono::steady_clock::now();
+      for (auto& table : (*doc)["tables"]) {
+         auto rows = table["rows"];
+         for (auto& row : rows) {
+            for (auto& cell : row) {
+               // deref -> slim 24-byte handle -> reconstruct full view -> read (RAMA's path)
+               const SlimHandle h{cell.document(), cell.data(), cell.parse_pos()};
+               glz::lazy_json_view<Opts> v{h.d, h.data};
+               v.set_parse_pos(h.pp);
+               sum += ingest_cell<Opts>(v, nulls);
+            }
+         }
+      }
+      const auto t1 = std::chrono::steady_clock::now();
+      ms = std::chrono::duration<double, std::milli>(t1 - t0).count();
+      return sum + static_cast<long>(nulls);
+   }
+} // namespace
+
+int main(int argc, char** argv)
+{
+   const char* cfg = (argc > 1) ? argv[1] : "full-on";
+
+   // ~1.14 GB flat cell-heavy fixture (same shape as lazy_load_benchmark's GB-scale stage).
+   const auto json = generate_flat_tables_json(60, 120000, lazy_load_bench::k_default_cols_per_row);
+
+   double ms = 0;
+   long sum = 0;
+   if (std::strcmp(cfg, "full-off") == 0) {
+      sum = timed_ingest<kEndBounded>(json, ms);
+   }
+   else if (std::strcmp(cfg, "full-on") == 0) {
+      sum = timed_ingest<kStreamingEnabled>(json, ms);
+   }
+   else if (std::strcmp(cfg, "slim-off") == 0) {
+      sum = timed_ingest<kEndBoundedSlim>(json, ms);
+   }
+   else if (std::strcmp(cfg, "slim-on") == 0) {
+      sum = timed_ingest<kStreamingEnabledSlim>(json, ms);
+   }
+   else if (std::strcmp(cfg, "recon-off") == 0) {
+      sum = timed_ingest_recon<kEndBounded>(json, ms); // RAMA slim-handle + fat view, streaming off
+   }
+   else if (std::strcmp(cfg, "recon-on") == 0) {
+      sum = timed_ingest_recon<kStreamingEnabled>(json, ms); // RAMA slim-handle + fat view, streaming on
+   }
+   else {
+      std::printf("unknown cfg '%s'\n", cfg);
+      return 2;
+   }
+
+   const double mb = static_cast<double>(json.size()) / (1024.0 * 1024.0);
+   std::printf("%s cpu_ms=%.1f MB/s=%.1f sum=%ld\n", cfg, ms, mb / (ms / 1000.0), sum);
+   return 0;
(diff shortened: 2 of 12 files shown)


stephenberry:
Thanks for this effort! Whenever you're finished iterating and optimizing let me know and I'll hammer it myself.

psiha:
@stephenberry 
> Thanks for this effort! Whenever you're finished iterating and optimizing let me know and I'll hammer it myself.

Hi, thank you for the prompt response and shared interest :)

I'm done with this for now - tried to do basic commit hygiene so you can do a clean merge commit/preserve history if you want.
Unfortunately the results are not so dramatic (double digit percentage slower than simdjson) as initially claimed - Cursor agent(s) pretty much lost the plot in the end so the results were skewed in all directions - final results are above, more in the modest range of a total 10-15% (like from 6% slower to 7% faster).

Main area of 'defect' was the (macro optimization) missing "streaming cursor" machinery - when you 'stream' over a huge JSON with arrays/objects of large subobjects (so when you proceed to the next subobject onto the next one it rescans the entire subobject that was just consumed). Further work could be done to have this streaming optimization also cover arrays (more involved and with trade-off decision - like whether to update to position on every [] access and the like).
Next was an asymmetry - missing SIMD accelerated skip in the lazy machinery that you already have for the structural parse - the PR introduces a new implementation for on_lazy, ideally you'll probably want to dedup this.

As always, tried to follow the 'do not pay for what you don't use' rule with the added flexible policies - up to you to chose what to expose finally...

You take over now, I'm still available for clarifications and collaboration...good luck :D

psiha:
Heads-up on a small stacked follow-up: farseerdev/glaze#1 (`perf(lazy): heap-free parse context on the scalar read_into fast path`) is based on this PR's branch. It makes `lazy_json_view::read_into<T>` select a minimal heap-free parse context for `num_t`/`bool_t` reads (leaving `glz::context` untouched), which lets toolchains DCE the dead `std::string` setup/teardown on the hot scalar ingest path — measured ~+3-5% on Linux/clang. It's kept on this branch (not `main`) purely to keep its diff self-contained; it can be retargeted to `main` / reopened upstream once this merges.

psiha:
@stephenberry 
Hi Stephen, just a gentle nudge - on whether you intend to pick this up or should we just proceed on our own merry way/fork? Either way is fine ;)

stephenberry:
@psiha thanks for pinging me. I'll look into this ASAP.

stephenberry:
@psiha, can you get `main` merged into this PR so that it is up to date?

psiha:
> @psiha, can you get `main` merged into this PR so that it is up to date?

@stephenberry done

stephenberry:
@psiha Thanks for the patience, and for the honest write-up at the end — the note about the agents losing the plot and the revised numbers saved me a lot of time, because it told me which claims to re-measure first.

I've taken this apart and landed it as three separate PRs rather than merging as-is. You're credited as co-author on all three and they're all your ideas; the changes are about scope and defaults, not about the substance. All three are now on main:

- **#2744** — `auto Opts` (`5cc5012b`)
- **#2745** — streaming cursor (`37437224`)
- **#2746** — wide number skip (`2512de17`)

**#2744 — the `auto Opts` fix.** This is the one I'd have merged on its own regardless. Your diagnosis was exactly right, and it turns out `beve/lazy.hpp` had the identical bug (32 more template heads), so that's fixed too. I added regression tests that assert a derived-only field actually reaches the parser — `bools_as_numbers` for JSON, `allow_conversions` for BEVE — plus the type-identity check. All four assertions fail against the old headers, which is the proof that the option was being silently dropped rather than merely unused.

**#2745 — the streaming cursor.** Kept, with one correctness fix and a smaller surface.

The fix: `read_into` recorded `[data_, it)` after `finalize_read_context` had already downgraded `partial_read_complete` (and a depth-zero `end_reached`) to success. Both leave `it` short of the value's true end, so with `partial_read` on, the recorded extent lands the next advance inside the object:

```
partial_read = true, 3 rows:
  cursor off -> count=3  sum=6    correct
  cursor on  -> count=11 sum=3    runaway
```

Completeness is now captured before `finalize_read_context` runs, and nothing is recorded under `partial_read` — correct, just not accelerated.

There was a second failure mode I only found by testing the cursor against readers that stop early on purpose. A custom `from` specialization that under-consumes its input — `glz::text` is the one in-tree example, it takes the rest of the buffer and reports success — made the recorded extent cover everything after the element, and iteration returned 1 element instead of 3. Validating where the reader landed was not enough; a reader stopping on an interior comma still diverged. What ships only records an extent for a container that ended exactly on its own closing bracket, which is a property of the bytes rather than a claim about the reader's honesty.

The surface: I collapsed the four-value policy to a plain `bool`. `packed` and `packed_with_fallback` were identical in release builds (the non-fallback path asserted, then cleared anyway), and the `uint32_t` packing was saving 8 bytes on a per-*document* struct rather than per view, so it was buying nothing for a real overflow-handling risk. I also dropped the new `core/` header — it made every glaze TU include a lazy-JSON-only file for one `check_*` helper that was never called. The option now lives in the `opts.hpp` comment list like `linear_search` and friends.

Tests are differential: every traversal runs cursor-on and cursor-off and the results must match exactly, so the option can only ever be a speed difference. That covers the cases the original didn't — nesting, interleaved iterators over one document, copy/move with a live extent, and `partial_read`. I measured +66% on a `read_into`-per-row loop, and checked `sizeof(lazy_document)` is unchanged when disabled, since the whole point of your first finding is that this thing can silently fail to be there.

**#2746 — the structural skip.** This is where I ended up disagreeing with the default, and I want to show you the numbers rather than just assert it.

Full traversal, bounded buffers, M1, clang `-O3`, interleaved best-of-5:

| shape | always-scan vs byte-at-a-time |
|---|---:|
| nested mixed (typical API payload) | **−31%** |
| tiny objects | **−24%** |
| telemetry records | −8% |
| long strings | −2% |
| long numeric runs (13-digit cells) | **+38%** |

The gap between structural bytes in ordinary JSON is two or three characters, which is far too short for a wide scan to amortize its setup. Your fixtures are dominated by long scalar cells, which is a real workload — it's just not the common one, and `lazy_swar_skip` defaulting to on made everyone else pay for it.

I tried the obvious rescues before giving up on the default. Restructuring so the scan only fires on scalar runs — never wasting a load on a byte that's already structural — still cost 24–31% on the dense shapes. Narrowing it to numbers only still cost 24–31%, because most JSON numbers are one to three digits and each paid a full scan setup.

What shipped is `lazy_wide_number_skip`, off by default: cheap table walk for short numbers, escalating to the SWAR scan only once a run is still going past 8 bytes. That gets you +34% on the numeric-run shape, costs 2–6% elsewhere if enabled on the wrong document, and measures within 0.1% of `main` when off. Your workload should get most of what you were after by turning it on; nobody else pays.

One detail worth mentioning because it was invisible to tests: the escalation set has to match `lazy_char_class` exactly, and dropping `"`, `[`, or `{` from it left all 517 assertions passing. It's now derived from a single list with a `static_assert` tying it to `lazy_char_class`, so a mismatch fails to compile instead of silently mis-skipping.

On your dedup point — agreed, and the scan now lives as `glz::find_first_of` in `util/parse.hpp` next to the `has_char` tests it's built from. I deliberately did *not* convert the four `skip_until_closed` overloads to use it: they're padded-buffer variants on the main read hot path, it'd be behavior-neutral with no measured upside, and it doesn't belong in a lazy-JSON PR. Happy to do that separately if you want the consolidation.

**Not carried over: the slim view.** Your own numbers were +8% / ≈0% / −26%, and it has no test coverage in the suite — `lazy_slim_view` appears only in the benchmark header, so CI never compiles that layout. For a permanent second layout of the primary public type, plus losing `error_code` fidelity and splitting what `key()` means depending on a flag, I don't think the measured case is there yet. If the clang-cl store-forwarding thing ever gets understood I'd look again.

**One more thing your PR led to, indirectly.** Working through the `null_terminated = false` paths turned up a memory-safety bug in `write_json` on a `lazy_json_view` that predates all of this — the writer found the value's extent with a hardcoded `opts{}`, whose `null_terminated = true` made it scan past the end of a bounded buffer looking for a sentinel that isn't there. Writing a view of the two-byte document `42` emitted `429187` and reported success. That's #2748, along with two more defects in the same function and a `write_beve` path on lazy BEVE views that never compiled at all. Not your bug, but I wouldn't have gone looking without your PR.

Two things I'd genuinely like your input on:

1. Does `lazy_wide_number_skip` actually recover the throughput you were seeing on the real multi-GB imports, or was the win coming from somewhere else in the 32-byte version that my 8-byte escalation misses? I only have synthetic fixtures; you have the real thing.
2. The `parse_pos()` / `set_parse_pos()` accessors — I left those out because they expose an internal scan cursor as mutable public API, which is hard to walk back later. What were you actually doing with them? If it's a real need there's probably a cleaner shape for it.

Thanks again for pushing on this. The `auto Opts` catch alone was worth the PR.

stephenberry:
[v8.0.0](https://github.com/stephenberry/glaze/releases/tag/v8.0.0) has been released with the critical improvements from the pull request implemented.

PR #2688, opened 2026-07-05
perf(lazy): heap-free parse context on the scalar read_into fast path

psiha:
## Heap-free parse context on the scalar `read_into` fast path

`lazy_json_view::read_into<T>` constructs a full `glz::context` for every value it parses. `glz::context` holds two `std::string` members — `current_file` and `scratch` — that object/string parsing needs but that number/bool parsing never touches. Since the context escapes by reference into the parser, some toolchains can't prove the strings stay empty and fail to DCE their construction/destruction, leaving dead `std::string` setup/teardown on the hottest scalar path (GB-scale numeric cell ingest).

### Change
- New `detail::scalar_parse_context { error_code error; uint32_t depth; string_view custom_error_message; }` — satisfies `is_context`, no heap-owning members (guarded by `static_assert(is_context<...>)`).
- `read_into<T>` selects the context type via `std::conditional_t<num_t<T> || bool_t<T>, detail::scalar_parse_context, context>`. Scalars run heap-free; objects/strings keep the full `glz::context`.
- `glz::context` is **untouched** → zero impact on any other format or reader. The scalar op set was verified against `from<JSON, num_t>`, `from<JSON, bool_t>` and `finalize_read_context`, all of which touch only `ctx.error`/`ctx.depth`.

### Correctness
- json + beve lazy test suites: **70/70** and **57/57**, unchanged.
- Flat scalar-ingest benchmark: checksum **bit-identical** between the full-context and scalar-context paths.

### Performance
GB-scale (~1.14 GB) cell-by-cell numeric ingest, pinned interleaved before/after A/B:

| toolchain | config | before | after | delta |
|---|---|---|---|---|
| Linux / clang 22 (`-O2 -march=native`) | full-view | 307.4 MB/s | 317.9 MB/s | **+3.4%** |
| Linux / clang 22 | handle-reconstruct | 270.8 MB/s | 283.7 MB/s | **+4.8%** |
| Windows / clang-cl | full-view | — | — | neutral (already DCEs the dead strings) |

A consistent ~3–5% on Linux/clang, neutral on Windows clang-cl.

---
**Stacked on #2683** (streaming cursor). This branch is based on that PR's branch, so until #2683 merges the diff/commit list here includes its commits; once #2683 lands this narrows to the single `read_into` commit. Review/merge after #2683.

### DIFF
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -35,6 +35,29 @@ target_compile_options(lazy_benchmark PRIVATE
   $<$<CXX_COMPILER_ID:MSVC>:/O2>
 )
 
+# Tabular bulk-load pattern (nested arrays-of-arrays + scalar read_into)
+add_executable(lazy_load_benchmark lazy_load_benchmark.cpp)
+target_link_libraries(lazy_load_benchmark PRIVATE
+  glaze
+  bencher::bencher
+)
+target_compile_options(lazy_load_benchmark PRIVATE
+  $<$<CXX_COMPILER_ID:GNU,Clang,AppleClang>:-O3 -march=native>
+  $<$<CXX_COMPILER_ID:MSVC>:/O2>
+)
+
+# Header-scan stage on large split-catalog fixture (skip/iterator guard; no row read_into loop)
+add_executable(lazy_load_scalar lazy_load_scalar.cpp)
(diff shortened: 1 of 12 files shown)


psiha:
Closing — this should be stacked so it merges into #2683's branch (feat/lazy-streaming-cursor), not into main. Reopening as a fork PR based on that branch so the diff stays a single self-contained commit; it flows upstream with #2683.

PR #2690, opened 2026-07-07
reject foreign column headers in colwise csv struct readers

uwezkhan:
Repro: read a headed colwise CSV whose column name is not a field but whose key hash collides onto one (e.g. a struct {alpha,bravo,charlie,delta} read with the header "uJXtL").

Cause: the two colwise readers route each header through decode_hash_with_size and check only index < N. The hash returns an in-range index for non-member keys that collide, so the foreign column is accepted and its value lands in an unrelated field.

Fix: compare reflect<T>::keys[index] against the header text before accepting, the same guard the rowwise object reader and the enum reader already apply.

Before: "uJXtL\n77\n" colwise returns success and sets alpha = 77.
After: the same input returns unknown_key, matching the rowwise path. Legitimate headers, including array-element headers like v3s[0], are unchanged.

Tradeoff: one extra string compare per column at parse time. Regression test added under the csv suite.

### DIFF
--- a/include/glaze/csv/read.hpp
+++ b/include/glaze/csv/read.hpp
@@ -1097,7 +1097,10 @@ namespace glz
                   const auto member_idx = decode_hash_with_size<CSV, U, HashInfo, HashInfo.type>::op(
                      key.data(), key.data() + key.size(), key.size());
 
-                  if (member_idx >= N) [[unlikely]] {
+                  // Confirm the decoded index actually matches the header string. The hash can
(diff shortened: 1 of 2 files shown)


stephenberry:
Excellent fix, and proper performance focus of doing the string compare on the right side of the `&&`.

uwezkhan:
Thanks for merging. With this the colwise readers now behave the same as the rowwise and enum paths on unknown keys.

Issue #2689, opened 2026-07-05
Does `null_terminated` really require NULL termination or just any readable character past the end?


Tester7373:
Consider the following example
```c++
#include <cassert>
#include <iostream>
#include <string>
#include <string_view>

#include <glaze/glaze.hpp>

struct Person {
	std::string name{};
	int			age{};
};

int main() {
	// One big buffer: the JSON document followed by unrelated payload.
	const std::string big = R"({"name":"Alice","age":30}#unrelated trailing data)";

	// View onto just the JSON part. big[json.size()] == '#', not '\0'.
	const std::string_view json{big.data(), big.find('#')};

	Person		   person{};
	constexpr auto options = glz::opts{.null_terminated = false};
	const auto	   ec	   = glz::read<options>(person, json);

	if (ec) {
		std::cerr << "Parse error: " << glz::format_error(ec, json) << '\n';
		return 1;
	}

	std::cout << "name = " << person.name << ", age = " << person.age << '\n';
	return 0;
}
```

Here using `null_terminated = false` is technically correct as the `string_view` is in fact not null terminated. However, as far as I understand from [this issue](https://github.com/stephenberry/glaze/issues/1054), glaze only requires the one-past-the-end character to be readable. So my question is: For this specific (but quite typical) setup of parsing a `string_view` obtained from a null terminated `std::string`, is using `null_terminated = true` safe? And if so, should the option maybe be named differently?

stephenberry:
You're correct that in this case where you have a valid input that is fully parsed, the null termination is not required. But, this is based on the fact that you have a complete typed struct that is parsed and the input is valid. If the input were invalid, such as a String that is never closed with an end quote, then null termination is required to stop parsing forever. Note that a # in a String is valid JSON.

If you can guarantee correct input then you don't need null termination. But, that's typically not true for most use cases.

Now, when it comes to null termination the rule for Glaze is that somewhere in the trailing allocated bytes beyond the valid section there must be a null character. So, the null character doesn't necessarily need to be the last character in the buffer.

stephenberry:
To add, in your example your string_view is null terminated according to Glaze's requirements, because it points to memory allocated in the std::string that is null terminated. Hope this makes sense.

Tester7373:
[...]

PR #2691, opened 2026-07-07
Reject truncated JSON arrays in null-terminated reads

stephenberry:
## Problem

With `null_terminated = true` (the default), Glaze silently accepts truncated JSON arrays. For example:

```cpp
std::vector<int> v{};
const auto error = glz::read_json(v, std::string{"[1,"}); // truncated JSON
// error.ec == none, v == {1}  -- the truncated array is accepted
```

Inputs like `[1,`, `[1,2,`, and a bare `[` all parse as "successful" complete arrays. Setting `validate_skipped`, `validate_trailing_whitespace`, or `error_on_missing_array_elements` does not help, since none of them gate the affected code path.

## Root cause

The *growing* array reader (used for `std::vector`, `std::deque`, `std::list`, `inplace_vector`, `glz::generic`, etc.) looped with `while (it < end)`. In null-terminated mode a trailing comma advances `it` to the terminating sentinel (or a bare `[` leaves the growing loop empty), so the loop condition becomes false and the loop **falls through without setting an error**.

This only affected the growing path. Fixed-size arrays (`std::array`), `std::set` (which pre-counts via `number_of_array_elements`, with an explicit `case '\0'`), and pre-sized containers with enough slots already rejected truncation because they parse the next value directly and fail on the sentinel. Object/map readers use `while (true)` and were never affected.

## Fix

A valid array always returns from inside the loop upon reaching `]`, so in null-terminated mode a fall-through can only mean the input was truncated. Report `unexpected_end` there:

```cpp
if constexpr (Opts.null_terminated) {
}
```

The guard is limited to the null-terminated case because non-null-terminated / streaming buffers already surface truncation through `skip_ws`/depth tracking (`end_reached`) or the streaming refill break, and those must keep their existing behavior. The check sits outside the loop and only runs on the fall-through path, so there is **no overhead on well-formed input**.

## Tests

Added `truncated array` and `truncated nested array` cases to `json_conformance.cpp`, covering `std::vector`, `std::deque`, `std::list`, and `glz::generic`. They run under all three `should_fail<Opts>()` configurations.

Verified locally:
- The reported repro now returns `unexpected_end` instead of `none`.
- `json_conformance`: 99 tests / 252 asserts pass.
- `json_test`: 688 tests / 60,433 asserts pass (no regressions).
- `istream_buffer_test` (streaming / non-null-terminated), `inplace_vector_test`, `int_parsing`, `json_reflection_test`, `ostream_buffer_test`, `mock_json_test` all pass.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -2426,6 +2426,17 @@ namespace glz
                      return;
                   }
                }
+
+               // A valid array always returns from inside the loop upon reaching ']'.
+               // For a null-terminated buffer the loop can only exit here by falling
+               // through when `it` reaches the terminator without a closing bracket,
+               // which means the input was truncated (e.g. "[", "[1," or "[1,2,").
+               // Non-null-terminated buffers surface this through skip_ws/depth
+               // tracking (or the streaming refill break), so this guard is limited
+               // to the null-terminated case.
+               if constexpr (Opts.null_terminated) {
+                  ctx.error = error_code::unexpected_end;
+               }
             }
             else {
                ctx.error = error_code::exceeded_static_array_size;
--- a/tests/json_conformance/json_conformance.cpp
+++ b/tests/json_conformance/json_conformance.cpp
@@ -52,6 +52,36 @@ inline void should_fail()
       }
    };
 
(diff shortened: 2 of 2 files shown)


stephenberry:
Addresses the issue raised in #2689

PR #2693, opened 2026-07-10
RFC 9112: validate request-line in HTTP server parser

annihilatorq:
Hi Stephen, sorry for the long diff - I just thought it would be easier to review than a bunch of stacked PRs.

This PR adds a strict `detail::parse_request_line` with full request-target validation (origin-form, absolute-form, asterisk-form), reusing the authority validators from #2676.

The current implementation of the HTTP request-line parser in Glaze has several issues, each of which relates to data validation within the request-line - `method SP request-target SP HTTP-version`

I deliberately started by writing tests that would cover the entire RFC 9112 on request-line validation, as well as the relevant sections of RFC 9110 and RFC 3986. Initially, I noticed that Glaze parsed the HTTP version not as `DIGIT "." DIGIT`, but as `*DIGIT "." *DIGIT`. After that, I started looking through the request-line parser code and noticed that validation for method, target, and HTTP version was missing, as well as the lack of normalization for the HTTP minor version.

## RFC 9112, 9110, 3986 non-compliance
**Total: 17** distinct non-compliances fixed. One of them is MUST-level (empty host in absolute-form request-target), rest are RFC grammar violations where a 400 response is a SHOULD per RFC 9112, Section 3, and one SHOULD about minor version handling.

Below is a list of RFC violations that will be fixed by this PR. The list is grouped by input class rather than by individual requirement, so some items cover several violated requirements at once:
1. `HTTP/1.2`-`HTTP/1.9` are not processed as HTTP/1.1.
   > RFC 9110, Section 6.2: A recipient that receives a message with a major version number that it implements and a minor version number higher than what it implements SHOULD process the message as if it were in the highest minor version within that major version to which the recipient is conformant.

2. `HTTP/11.1` and `HTTP/1.11` are accepted.
   > RFC 9112, Section 2.3: HTTP-version = HTTP-name "/" DIGIT "." DIGIT

3. Malformed method tokens (`GE/T`, `GE:T`, control bytes, `\tGET`) get 501 instead of 400, although 501 is reserved for valid but unsupported method tokens.
   > RFC 9112, Section 3: method = token

   > RFC 9110, Section 5.6.2: token = 1*tchar

   > RFC 9112, Section 3: Recipients of an invalid request-line SHOULD respond with either a 400 (Bad Request) error or a 301 (Moved Permanently) redirect with the request-target properly encoded.

4. Leading whitespace is accepted: `" GET ..."` parses fine, while `"\tGET ..."` gets 501.
   > RFC 9112, Section 3: request-line = method SP request-target SP HTTP-version

5. The request-target form is not validated. Targets such as `products`, `?x=1`, `example.com:443`, `**`, `*?x=1` and `GET *` are all accepted.
   > RFC 9112, Section 3.2: request-target = origin-form / absolute-form / authority-form / asterisk-form

   > RFC 9112, Section 3.2.1: origin-form = absolute-path [ "?" query ]

   > RFC 9112, Section 3.2.3: The "authority-form" of request-target is only used for CONNECT requests (Section 9.3.6 of [HTTP]).

   > RFC 9112, Section 3.2.4: When a client wishes to request OPTIONS for the server as a whole ... the client MUST send only "*" (%x2A) as the request-target

6. Raw whitespace and control bytes inside the request-target are accepted (`\t`, `\v`, `\f`, `\0`, `\x7f`).
   > RFC 9112, Section 3: No whitespace is allowed in the request-target.

   > RFC 9112, Section 3: A recipient SHOULD NOT attempt to autocorrect and then process the request without a redirect, since the invalid request-line might be deliberately crafted to bypass security filters along the request chain.

7. Characters outside the pchar grammar are accepted in path and query (`"`, `<`, `>`, `[`, `]`, `\`, `^`, `` ` ``, `{`, `|`, `}`, raw non-ASCII bytes).
   > RFC 3986, Appendix A: pchar = unreserved / pct-encoded / sub-delims / ":" / "@"

8. Fragment marker is accepted (`/products#section`).
   > RFC 9110, Section 7.1: The target URI excludes the reference's fragment component, if any, since fragment identifiers are reserved for client-side processing ([URI], Section 3.5).

9. Malformed pct-encoding is accepted (`%`, `%0`, `%GG`, `%2G`, `%u1234`).
   > RFC 3986, Section 2.1: pct-encoded = "%" HEXDIG HEXDIG

10. Invalid absolute-form targets are accepted: empty host (`http:///products`), userinfo (`http://user@example.com/`), non-digit port (`http://example.com:abc/`), unterminated IPv6 bracket (`http://[2001:db8::1/products`).
    > RFC 9110, Section 4.2.1: A sender MUST NOT generate an "http" URI with an empty host identifier. A recipient that processes such a URI reference MUST reject it as invalid.

    > RFC 9110, Section 4.2.4: Before making use of an "http" or "https" URI reference received from an untrusted source, a recipient SHOULD parse for userinfo and treat its presence as an error; it is likely being used to obscure the authority for the sake of phishing attacks.

    > RFC 3986, Section 3.2.3: port = *DIGIT

Most of the items above are grammar violations where responding with 400 is a SHOULD (RFC 9112, Section 3), not a MUST. Hard MUST violations are the empty host in absolute-form target and bare CR handling from RFC 9112, Section 2.2.
Still, the RFC states the security reasoning for respecting this SHOULD in Section 3: an invalid request-line "might be deliberately crafted to bypass security filters along the request chain".

---

I have a few more changes coming up regarding the `glz::http_server` RFC compliance, but I think it would be best to first fine-tune this parser's performance.


annihilatorq:
I've intentionally pushed failing tests to start with, and I'll now push the changes that will fix them. If the previous request-line validation needs to be checked, you can clone ea90efd

annihilatorq:
@stephenberry, functionally, the current validator and parser should be correct, but they can definitely be sped up by eliminating a few extra loops in `find` and other operations. The current implementation is as vanilla as possible for ease of reading and review; in the next few days, I plan to optimize it.

Also, if you find the `chore` commit (551f7af) renaming `validate_*` to `is_valid_*` to be controversial and think it breaks with the project's coding style, I'm happy to rename everything back. I just thought that `is_valid_*` reads more naturally in this case (see 551f7af description).

stephenberry:
`is_valid` is good. And, correctness is best to focus on initially. Performance can be improved in the future.

annihilatorq:
Sure, in that case, everything is ready for the review. Just as a note for performance - current request-line parser could be optimized significantly, but you're right, it's probably makes more sense to first ensure RFC compliance and write tests to avoid behavior regressions due to optimizations

stephenberry:
@annihilatorq, this all looks good. Feel free to merge it in.

PR #2696, opened 2026-07-10
Add additional write_file_yaml overloading with a buffer argument

### LINKED ISSUE #2695: Feature request: allow the buffer as an input argument of write_file_yaml
Hi,

I would like to the same overloading for `write_file_yaml` as for `write_file_json`, which allows the string buffer an input argument.

With the JSON interface, we have

https://github.com/stephenberry/glaze/blob/75e2fbb03129af549b6093e9c16493ceb1c953db/include/glaze/json/write.hpp#L2808-L2820

This is very useful when a detailed error message is needed if the parsing failed.

Similarly, I would like the same with `write_file_yaml`:

```cpp
   template <auto Opts = yaml::yaml_opts{}, class T>
   [[nodiscard]] error_ctx write_file_yaml(T&& value, const sv file_path, auto&& buffer) noexcept;
```

YanzhaoW:
Fixes #2695

### DIFF
--- a/include/glaze/yaml/write.hpp
+++ b/include/glaze/yaml/write.hpp
@@ -1634,11 +1634,10 @@ namespace glz
    }
 
    template <auto Opts = yaml::yaml_opts{}, class T>
-   [[nodiscard]] error_ctx write_file_yaml(T&& value, const sv file_path) noexcept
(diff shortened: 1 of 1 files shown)


stephenberry:
Thanks for this!

PR #2704, opened 2026-07-14
RFC 9112: fix HTTP/1.0 Host header field validation

annihilatorq:
This PR applies the Host field validation rules as specified in RFC 9112 for HTTP/1.0 requests.
In one of my previous PRs (#2676), I misinterpreted a subtle distinction in the text of RFC 9112, Section 3.2, not noticing the distinction between an HTTP/1.1 request and a request of any version, which is described separately:

> A server MUST respond with a 400 (Bad Request) status code to any
> HTTP/1.1 request message that lacks a Host header field and to any
> request message that contains more than one Host header field line or
> a Host header field with an invalid field value.

If we take a closer look, there is a clear distinction here:

> ... HTTP/1.1 request message ...
> ... and to "any request message" ...

### DIFF
--- a/include/glaze/net/http_server.hpp
+++ b/include/glaze/net/http_server.hpp
@@ -2581,31 +2581,28 @@ namespace glz
          auto& headers = conn->request_.headers;
          bool host_header_parsed = false;
 
+         // RFC 9112, Section 3.2:
+         // A server MUST respond with a 400 (Bad Request) status code to any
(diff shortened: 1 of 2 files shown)


stephenberry:
Is this good to merge?

annihilatorq:
Yes, it's ready (as soon as the CI is green)

PR #2705, opened 2026-07-15
Use ephemeral ports for test servers

annihilatorq:
This will fix the flaky CI runs that have recently started occurring quite frequently ([1](https://github.com/stephenberry/glaze/actions/runs/29376184736/job/87230024448), [2](https://github.com/stephenberry/glaze/actions/runs/29157147577/job/86556016016), [3](https://github.com/stephenberry/glaze/actions/runs/29157147577/job/86556016016)).

The most stable solution is to use ephemeral-ports, by binding port 0, the OS will assign a random, unoccupied port, after which we simply need to read the local endpoint and use the port from it. I've already [encountered this problem before](https://github.com/blazeauth/aero/pull/26/changes/797693013a4560c85be302bf475a2fd78a98c7f9) and this has proven to be a reliable fix.

I'd like to note that, specifically for this PR, I used an AI to scan and perform this boilerplate task. I couldn't find anything regarding Glaze's policy on AI use, but I've reviewed the final diff and it's LGTM.

### DIFF
--- a/tests/networking_tests/asio_repe/asio_repe.cpp
+++ b/tests/networking_tests/asio_repe/asio_repe.cpp
@@ -22,9 +22,8 @@ struct notify_api
 
 void notify_test()
 {
-   static constexpr int16_t port = 8431;
    std::latch ready{1};
-   glz::asio_server server{.port = port, .concurrency = 4};
+   glz::asio_server server{.port = 0, .concurrency = 4};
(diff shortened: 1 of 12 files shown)


stephenberry:
Thanks for this. I'm good with AI usage and use it myself. It just needs to be vetted, unit tested, and the changes well motivated.

PR #2708, opened 2026-07-16
guard lazy json scans against end of non-null-terminated input

uwezkhan:
The lazy JSON parser in json/lazy.hpp scans values on demand, and several of its helpers dereference the cursor without checking it against end. On a non-null-terminated buffer (a std::string_view or std::span, where the reader appends no NUL sentinel) a value or key that begins at end-of-input reads one byte past the buffer: skip_value_lazy reads the tag and returns data+4 past end for a truncated true/false/null, parse_key reads the opening quote, and the object scans read the ":" and "," separators. operator[], size(), index(), and range iteration reach these on a truncated document.

The fix bounds every scan dereference against the remaining input, clamps the literal skips to end, and stops the object, index, and iterator scans from returning a view whose data pointer sits at end, since each view operation reads *data_. The scanner is the only layer that sees end; a caller holding a string_view cannot make a truncated value complete. Null-terminated input keeps its sentinel fast path, so valid documents behave exactly as before.

### DIFF
--- a/include/glaze/json/lazy.hpp
+++ b/include/glaze/json/lazy.hpp
@@ -161,15 +161,27 @@ namespace glz
             }
             return p;
          }
-         default:
-            // Number
+         default: {
+            // Number, or an unrecognized byte.
+            const char* const start = p;
             if constexpr (Opts.null_terminated) {
                while (numeric_table[uint8_t(*p)]) ++p;
             }
             else {
                while (p < end && numeric_table[uint8_t(*p)]) ++p;
             }
+            // Guarantee forward progress. A byte that begins no valid value (not a string,
+            // literal, container, or number - impossible in well-formed JSON) leaves p unmoved,
+            // which would spin every caller that skips values in a loop (size(), index(),
+            // operator[], iteration) forever on malformed input. Treat it as a one-byte scalar so
+            // the scan always advances whenever p < end. Bounded by end because json_end_ is the
+            // authoritative message end: a byte within it is safe to step over, while at or past
+            // end we must not move (the caller's own p >= end guard then stops the loop). On
+            // valid input p is always a real value start, so this branch never fires - no
+            // behavior change.
+            if (p == start && p < end) ++p;
             return p;
+         }
          }
       }
    } // namespace detail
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for the fix @uwezkhan. I pushed a follow-up commit (b59d6ae1) that extends the guarding and closes two remaining gaps I found while reviewing.

### 1. Extend the end-of-input guards to `null_terminated` mode

The original change guards non-null-terminated buffers. But `null_terminated` in Glaze only promises a `'\0'` sentinel exists *somewhere* the byte scanners may rely on. It does **not** guarantee the sentinel sits at `json_end_` (it can be trailing buffer content further on), so `json_end_` is the authoritative message end in both modes.

`size()`, `index()`, `operator[](size_t)`, `operator[](std::string_view)`, `empty()`, and iteration now bound by `json_end_` in `null_terminated` mode too, so on truncated/unclosed input they neither read past the buffer nor run off the logical message into trailing bytes — and they now agree with each other on element counts (`size()` matches `index()`/iteration, `empty()` matches `size() == 0`). On well-formed JSON the closing bracket always precedes `json_end_`, so these are dead branches: no behavior change on valid input.

### 2. Guarantee forward progress in `skip_value_lazy` (fixes an infinite loop / DoS)

A byte that begins no valid value — not a string, literal, container, or number — made the number branch consume nothing, leaving the scan position unmoved. Every value-skipping loop (`size`/`index`/`operator[]`/iteration) then spun forever on malformed input such as `[@` or `{"a":@}`. Since `lazy_json()` only validates the first byte, such input was accepted and then hung the accessors — a DoS on untrusted input.

The fix treats an unrecognized value-start byte as a one-byte scalar so the scan always advances whenever `p < end`. It is bounded by `end` (never steps over `json_end_`) and is unreachable on well-formed JSON, so valid input is unaffected.

### Tests

Added regression tests across all four accessors plus iteration, in both modes. Backing stores are exact-sized heap buffers (no trailing `'\0'`, or a single `'\0'` at `json_end_`), so the ASan/UBSan CI builds abort on any read past the message end. Full lazy suite is 85 tests / 436 asserts, green under ASan + UBSan. Each new test was confirmed to fail (or, for the internal loops, hang) against the un-fixed code.

uwezkhan:
Thanks for merging. The follow-up makes sense, json_end_ as the authoritative end in both modes is the right call, and good catch on the stalled number branch. I'd missed that the scan could stop advancing on a bad value-start byte.

PR #2709, opened 2026-07-18
Add HTTP headers API

annihilatorq:
A more refined version which supersedes #2402.

## Motivation
Main issues with using `std::unordered_map` to store header fields are:
1. **Data loss for repeated fields**. [RFC 9110, Section 5.3](https://datatracker.ietf.org/doc/html/rfc9110#section-5.3) specifies that header fields do not necessarily have to be concatenated before being sent over the wire. Therefore, a header section where multiple header-fields separated by CRLF have the same name is valid, and the implementation must handle such cases. Current behavior: `response_headers[...] = ...` - last wins.
2. **Names are stored in lowercase**. If the original case is important to the user during iteration, they won't be able to retrieve it. This would make sense if Glaze implemented HTTP/2 or HTTP/3, but it doesn't at this time. The issues this might cause are mostly cosmetic (e.g. logging/debugging), but it's still a problem. This can be fixed using a custom comparator, but that won't resolve the issue above.
3. **Parsing becomes more complicated when correcting data loss**. Even if we solve the problem with the comparator for `std::unordered_map`, a problem arises with concatenating header fields to avoid "last-wins" scenarios. When we receive multiple headers with the same name, , we need to concatenate their values, separating them with a comma (", "), which complicates the parser's behavior and makes storage less straightforward.
4. **Loss of the original field order**. `std::unordered_map` is unordered, so this behavior will cause problems during logging and/or debugging.

## Proposed solution
The class I propose is a low-level storage for header fields with a modern interface that satisfies the `std::ranges::forward_range` concept, allowing Glaze users to easily use this class into a wide variety of algorithms and scenarios in just a few lines of code.

The entire class iteration is a simple view idiom, no copies are created during iterations, and the entire immutable API is built on std::string_view.

### Short code example
```cpp
glz::http_headers headers{
   {"Content-Type", "application/json"},
   {"X-Request-ID", "7d42"},
   {"X-Rate-Limit", "100"},
   {"Cache-Control", "no-cache, no-store"},
   {"Set-Cookie", "theme=dark"},
};

headers.add("Set-Cookie", "session=7d42");
headers.set("Content-Type", "application/json; charset=utf-8");

if (auto first_cookie = headers.first_value("Set-Cookie")) {
   std::println("First cookie: {}", *first_cookie);
}

auto custom = headers | std::views::filter([](const glz::http_header& header) {
                 return header.name.starts_with("X-");
              });

for (const glz::http_header& header : custom) {
   std::println("{}: {}", header.name, header.value);
}

for (const auto& [name, value] : headers.fields("Set-Cookie")) {
   std::println("{}: {}", name, value);
}

if (headers.contains_token("Cache-Control", "no-store")) {
   std::println("response must not be stored");
}

std::print("{}", headers.serialize());
```

### API
```cpp
struct http_header
{
  std::string name;
  std::string value;

  [[nodiscard]] bool contains_token(std::string_view token) const;
};

class http_headers
{
 public:
  using value_type = glz::http_header;
  using iterator = std::vector<value_type>::iterator;
  using const_iterator = std::vector<value_type>::const_iterator;
  using size_type = std::vector<value_type>::size_type;

  private:
  template <bool IsConst>
  class matching_iterator
  {
      // std::forward_iterator_tag
  };

  public:
  using range_iterator = matching_iterator<false>;
  using const_range_iterator = matching_iterator<true>;

  http_headers() = default;
  http_headers(std::initializer_list<http_header> fields);

  [[nodiscard]] bool empty() const noexcept;
  [[nodiscard]] size_type size() const noexcept;
  [[nodiscard]] size_type capacity() const noexcept;
  [[nodiscard]] iterator begin() & noexcept;
  [[nodiscard]] iterator end() & noexcept;
  [[nodiscard]] const_iterator begin() const& noexcept;
  [[nodiscard]] const_iterator end() const& noexcept;
  [[nodiscard]] const_iterator cbegin() const& noexcept;
  [[nodiscard]] const_iterator cend() const& noexcept;
  [[nodiscard]] http_header& front() & noexcept;
  [[nodiscard]] const http_header& front() const& noexcept;
  [[nodiscard]] http_header& back() & noexcept;
  [[nodiscard]] const http_header& back() const& noexcept;

  void reserve(size_type count);
  void clear() noexcept;

  [[nodiscard]] iterator find(std::string_view name) & noexcept;
  [[nodiscard]] const_iterator find(std::string_view name) const& noexcept;
  [[nodiscard]] auto fields(std::string_view name GLZ_LIFETIMEBOUND) &;
  [[nodiscard]] auto fields(std::string_view name GLZ_LIFETIMEBOUND) const&;
  [[nodiscard]] auto names() const&;
  [[nodiscard]] auto values() const&;
  [[nodiscard]] auto values(std::string_view name GLZ_LIFETIMEBOUND) const&;
  [[nodiscard]] size_t count(std::string_view name) const& noexcept;
  [[nodiscard]] std::optional<std::string_view> first_value(std::string_view name) const& noexcept;
  [[nodiscard]] bool contains(std::string_view name) const noexcept;
  [[nodiscard]] bool contains_token(std::string_view name, std::string_view token) const;

  [[nodiscard]] std::string serialize() const;

  iterator set(std::string name, std::string value) &;
  iterator add(std::string name, std::string value) &;
  void append(const http_headers& other);
  void append(http_headers&& other);
  void erase(std::string_view name);

  // Disallow use on temporary objects
  iterator begin() && = delete;
  iterator end() && = delete;
  const_iterator begin() const&& = delete;
  const_iterator end() const&& = delete;
  const_iterator cbegin() const&& = delete;
  const_iterator cend() const&& = delete;
  http_header& front() && = delete;
  const http_header& front() const&& = delete;
  http_header& back() && = delete;
  const http_header& back() const&& = delete;
  iterator find(std::string_view name) && = delete;
  const_iterator find(std::string_view name) const&& = delete;
  auto fields(std::string_view name) && = delete;
  auto fields(std::string_view name) const&& = delete;
  auto names() const&& = delete;
  auto values() const&& = delete;
  auto values(std::string_view name) const&& = delete;
  std::optional<std::string_view> first_value(std::string_view name) const&& = delete;
  iterator add(std::string name, std::string value) && = delete;
  iterator set(std::string name, std::string value) && = delete;

 private:
  std::vector<value_type> headers_;
};
```

### Key design idea
Current `glz::http_headers` is intended to store header fields without the parser formatting their values. In other words, if the parser receives a header section containing multiple fields with the same name, the parser places all of these fields into `glz::http_headers` after validation, without concatenating their values.

The main API point is that by implementing the std::range interface, all non-trivial operations not covered by the current shortcut API can be easily implemented using C++2x ranges. This is one of the key advantages of the current design, because code that uses this API will look really idiomatic alongside modern heavy-std code.

Here are a few methods for understanding the design of this class:
| Method | What it solves |
|---|---|
| `fields(name)` | Returns every header with the given name as a range. Useful for headers that can appear more than once, e.g. `Set-Cookie`. |
| `names()` | Returns a lazy view over all header names. |
| `values()` | Returns a lazy view over all header values. The `values(name)` overload limits the view to header fields whose names match the passed one. |
| `serialize()` | Serializes header fields into HTTP wire format, including the final CRLF to finish the header section. |
| `contains_token(name, token)` | Checks for a comma-separated token in header fields whose names match the passed one. Token matching is case-insensitive and ignores optional whitespace. Useful for list-based headers, such as `Connection`, `Cache-Control`, etc. |
| `first_value(name)` | Returns the value of the first matching header, or `std::nullopt` if the header is not present. Provides a convenient layer for callers that only need a single value. |
| `set(name, value)` | Sets a header field to exactly one value. If the header already exists, its first occurrence is updated and all duplicates are removed. |
| `add(name, value)` | Appends a new header without replacing existing ones. Useful when repeated header lines are meaningful or required. |

### Pros
1. **Preserving the order of header fields** due to the use of `std::vector`, as well as manually preserving the insertion order when using `.set()`. Instead of simply using `.erase()` followed by `.add()`, implementation performs an in-place insertion while removing duplicates using the `erase-remove` idiom.
2. **Preservation of name case**. The container does not modify any values inserted by the user, searches are performed using the SWAR-optimized `glz::striequal` and have linear complexity.
3. **The parser stays straightforward and simple**. Concatenating the values of header fields with the same name is not required due to the current class design.
5. A clean and intuitive API for anyone who has ever worked with HTTP header fields.

### Cons
One potential issue that may arise and which the design takes into account - is temporary objects. Personally, I don't consider this a problem, but I understand that such cases can occur. It's more of a trade-off you have to accept when working with `std::ranges` and `std::views`, and this class simply follows that idiom.

Here's the example:
```cpp
// 1. Save range that was built from a temporary string to use it later
auto fields = h.fields(build_name()); // build_name() returns an std::string
for (auto& field : fields) { ... } // UB, std::string_view inside the "fields" iterator is pointing to a dead buffer

// 2. Return lazy range from function, in which the header name is a local string
auto cookie_values(const headers& h, std::string name) {
    return h.values(name); // name will get destroyed on return, UB
}

// 3. Rare, but still: header name was changed while range is alive
std::string name = "set-cookie";
auto r = h.fields(name);
name = "access-control-request-headers";
// Best case - iterators are now searching for other name
// Worst case - UB if new string assignment resulted in reallocation
```

### Tackling the downsides
- With C++23 support, we have a clear advantage that we can leverage, this is [P2718R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2718r0.html), which solves the problem of the lifetimes of temporary objects inside a range-loop.
Such code stays valid: `for (const auto& f : h.fields(std::string{prefix} + "-id")) { ... }`.
- Using the `[[compiler::lifetimebound]]` attribute practically eliminates the risk of errors when passing a temporary string to the `.fields()` and `.values()` methods. This attribute causes the compiler to warn the user when temporary `std::string` objects are used as arguments.
Unfortunately, GCC does not support this attribute, so users will only see warnings when using Clang or MSVC.

### Afterword
If I forgot to go into detail about any aspect of the design, I apologize. I'll be happy to answer any follow-up questions you may have regarding the decisions made.
It's also worth mentioning that I'm not finalizing the proposed API design. I'm simply presenting an initial version that's already ready for practical use.

As soon as we finalize the API - if that's okay with you - I'll start replacing the current code in the `net/` directory that uses `std::unordered_map`, and then I'll write the documentation for `glz::http_headers`.

### DIFF
--- a/include/glaze/net/http_headers.hpp
+++ b/include/glaze/net/http_headers.hpp
@@ -1,0 +1,324 @@
+// Glaze Library
+// For the license information refer to glaze.hpp
+
+#pragma once
+
+#include <algorithm>
+#include <optional>
+#include <ranges>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "glaze/util/attributes.hpp"
+#include "glaze/util/compare.hpp"
+
+namespace glz
+{
+   namespace detail
+   {
+      [[nodiscard]] inline std::string_view trim_optional_whitespace(std::string_view text)
+      {
+         size_t first_non_whitespace = text.find_first_not_of(" \t");
+         if (first_non_whitespace == std::string_view::npos) {
+            return {};
+         }
+         size_t last_non_whitespace = text.find_last_not_of(" \t");
+         return text.substr(first_non_whitespace, last_non_whitespace - first_non_whitespace + 1);
+      }
+   } // namespace detail
+
+   struct http_header
+   {
+      std::string name;
+      std::string value;
+
+      [[nodiscard]] bool contains_token(std::string_view token) const&
+      {
+         if (token.empty()) {
+            return false;
+         }
+
+         for (auto&& split_value : value | std::views::split(',')) {
+            std::string_view candidate{split_value};
+
+            if (glz::striequal(detail::trim_optional_whitespace(candidate), token)) {
+               return true;
+            }
+         }
+
+         return false;
+      }
+   };
+
+   class http_headers
+   {
+     public:
+      using value_type = glz::http_header;
+      using iterator = std::vector<value_type>::iterator;
+      using const_iterator = std::vector<value_type>::const_iterator;
+      using size_type = std::vector<value_type>::size_type;
+
+     private:
+      template <bool IsConst>
+      class matching_iterator
+      {
+         using owner_type = std::conditional_t<IsConst, const http_headers*, http_headers*>;
+
+        public:
+         using iterator_concept = std::forward_iterator_tag;
+         using iterator_category = std::forward_iterator_tag;
+         using difference_type = ptrdiff_t;
+         using value_type = http_headers::value_type;
+         using reference = std::conditional_t<IsConst, const value_type&, value_type&>;
+         using pointer = std::conditional_t<IsConst, const value_type*, value_type*>;
+
+         matching_iterator() = default;
+         matching_iterator(const matching_iterator&) = default;
+         matching_iterator(matching_iterator&&) = default;
+         matching_iterator& operator=(const matching_iterator&) = default;
+         matching_iterator& operator=(matching_iterator&&) = default;
+         ~matching_iterator() = default;
+
+         matching_iterator(owner_type owner, size_t start_index, std::string_view key)
+            : owner_(owner), index_(start_index), key_(key)
+         {
+            seek_forward();
+         }
+
+         [[nodiscard]] reference operator*() const noexcept { return owner_->headers_[index_]; }
+         [[nodiscard]] pointer operator->() const noexcept { return std::addressof(owner_->headers_[index_]); }
+
+         matching_iterator& operator++() noexcept
+         {
+            if (index_ < owner_->headers_.size()) {
+               ++index_;
+               seek_forward();
+            }
+            return *this;
+         }
+
+         matching_iterator operator++(int) noexcept
+         {
+            auto copy = *this;
+            ++(*this);
+            return copy;
+         }
+
+         [[nodiscard]] friend bool operator==(const matching_iterator& left, const matching_iterator& right) noexcept
+         {
+            return left.owner_ == right.owner_ && left.index_ == right.index_ && glz::striequal(left.key_, right.key_);
+         }
+
+        private:
+         void seek_forward() noexcept
+         {
+            while (index_ < owner_->headers_.size()) {
+               const auto& current = owner_->headers_[index_].name;
+               if (glz::striequal(current, key_)) {
+                  return;
+               }
+               ++index_;
+            }
+         }
+
+         owner_type owner_{};
+         size_t index_{0};
+         std::string_view key_;
+      };
+
+     public:
+      using range_iterator = matching_iterator<false>;
+      using const_range_iterator = matching_iterator<true>;
+
+      http_headers() = default;
+      http_headers(std::initializer_list<http_header> fields) : headers_(fields) {}
+
+      [[nodiscard]] bool empty() const noexcept { return headers_.empty(); }
+      [[nodiscard]] size_type size() const noexcept { return headers_.size(); }
+      [[nodiscard]] size_type capacity() const noexcept { return headers_.capacity(); }
+      [[nodiscard]] iterator begin() & noexcept { return headers_.begin(); }
+      [[nodiscard]] iterator end() & noexcept { return headers_.end(); }
+      [[nodiscard]] const_iterator begin() const& noexcept { return headers_.begin(); }
+      [[nodiscard]] const_iterator end() const& noexcept { return headers_.end(); }
+      [[nodiscard]] const_iterator cbegin() const& noexcept { return headers_.cbegin(); }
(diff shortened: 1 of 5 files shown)


stephenberry:
This is shaping up nicely!

Can you make this a `struct`, glaze prefers structs, especially when the next line is just making things public:

```c++
class http_headers
{
 public:
```

annihilatorq:
Done! Unrelated question out of curiosity - what's the rationale behind glaze preferring structs, even with private members? Is the reason related to reflection?

stephenberry:
> Unrelated question out of curiosity - what's the rationale behind glaze preferring structs, even with private members? Is the reason related to reflection?

The main reason Glaze prefers structs is to know that all structs can be forward declared with `struct`. If you mix structs and classes then you have to keep track of whether you forward declare with `struct` or `class`. Compilers will complain if you mismatch forward declarations with the actual class vs struct. It's a minor nit, but it makes development more standardized and structs works for all use cases of `class` in this context.

stephenberry:
I'd say this is good to merge.

PR #2711, opened 2026-07-20
Call shrink_to_fit after growing JSON array reads when requested

ays7:
When `opts.shrink_to_fit` is enabled, JSON array reads now call `shrink_to_fit()`
on the emplace_back growing path after the closing `]`.
That option was already honored for empty arrays and for the fixed-size overwrite
path, but not when the container grew element-by-element via `emplace_back`.
Streaming reads benefit most, since they use geometric growth and can leave excess
capacity.

### DIFF
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -2419,6 +2419,11 @@ namespace glz
                         --ctx.depth;
                      }
                      ++it;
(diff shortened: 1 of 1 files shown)


stephenberry:
I built off this work and improved shrink_to_fit handling in #2712, which is now merged, so I'm closing this

PR #2718, opened 2026-07-22
Replace the HTTP header maps with glz::http_headers

annihilatorq:
Ports the net stack to glz::http_headers (#2709).

Rewriting the header lookups for the new container resulted in fixing two WebSocket bugs. `glz::http_headers` keeps repeated field names instead of overwriting them, and its `.contains_token()` replaced the hand-rolled substring and prefix comparisons, so every lookup had to be revisited, and both were found there.

## WebSocket implementation bugs
1 - the handshake response check never compared a full field name. The code used a comparison like this: 
`strncasecmp(name.c_str(), "Upgrade", 7)` for the headers "Upgrade", "Connection" and "Sec-WebSocket-Accept" comparing the first N bytes using `strncasecmp`, where N is the length of the expected header name. The problem is that this approach would accept headers such as "Connection-Something" as valid because the name starts with "Connection", and so on with other headers listed previously.

Example - the old client accepted this response even though the server never sent an `Upgrade` field:
```
HTTP/1.1 101 Switching Protocols\r\n
Upgrade-Extra: websocket\r\n
Connection: Upgrade\r\n
Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n\r\n
```

---

2 - the client is allowed to send a list-based field not as a comma-separated list, but as separate header fields separated by CRLF. The previous issue was that `std::unordered_map` used a "last-wins" approach, thereby discarding one of the values for that header. In such cases, user could only hope that the last header received contained the required token (`"Connection: Upgrade"`) as the header-field value.

Example - the old server refused this valid request because `keep-alive` arrived last:
```
GET /ws HTTP/1.1\r\n
Host: 127.0.0.1:8080\r\n
Upgrade: websocket\r\n
Connection: Upgrade\r\n
Connection: keep-alive\r\n
Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==\r\n
Sec-WebSocket-Version: 13\r\n\r\n
```

### DIFF
--- a/docs/networking/advanced-networking.md
+++ b/docs/networking/advanced-networking.md
@@ -74,7 +74,7 @@ server.use(custom_cors);
 server.get("/test-cors", [](const glz::request& req, glz::response& res) {
     res.json({
         {"message", "CORS test endpoint"},
-        {"origin", req.headers.count("Origin") ? req.headers.at("Origin") : "none"},
+        {"origin", std::string{req.headers.first_value("Origin").value_or("none")}},
         {"method", glz::to_string(req.method)},
         {"timestamp", std::time(nullptr)}
     });
@@ -207,12 +207,12 @@ secure_ws->on_validate([](const glz::request& req) -> bool {
         return false;
     }
     
-    return validate_jwt_token(auth_header->second);
+    return validate_jwt_token(auth_header->value);
 });
 
 secure_ws->on_open([](auto conn, const glz::request& req) {
     // Store user info from validated token
-    auto user_data = extract_user_from_token(req.headers.at("Authorization"));
+    auto user_data = extract_user_from_token(*req.headers.first_value("Authorization"));
     conn->set_user_data(std::make_shared<UserData>(user_data));
     
     conn->send_text("Authenticated successfully");
@@ -384,7 +384,7 @@ User new_user{0, "John Doe", "john@example.com"};
 auto create_response = client.post_json("https://api.example.com/users", new_user);
 
 // POST with custom headers
-std::unordered_map<std::string, std::string> headers = {
+glz::http_headers headers = {
     {"Authorization", "Bearer " + token},
     {"Content-Type", "application/json"}
 };
@@ -444,7 +444,7 @@ auto timing_middleware = [](const glz::request& req, glz::response& res) {
 auto response_time_middleware = [](const glz::request& req, glz::response& res) {
     auto start_header = res.response_headers.find("X-Request-Start");
     if (start_header != res.response_headers.end()) {
-        auto start_ms = std::stoll(start_header->second);
+        auto start_ms = std::stoll(start_header->value);
         auto now_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
             std::chrono::high_resolution_clock::now().time_since_epoch()).count();
         
--- a/docs/networking/advanced-networking.md
+++ b/docs/networking/advanced-networking.md
@@ -202,12 +202,12 @@ auto secure_ws = std::make_shared<glz::websocket_server>();
 
 // Validate connections before upgrade
 secure_ws->on_validate([](const glz::request& req) -> bool {
-    auto auth_header = req.headers.find("Authorization");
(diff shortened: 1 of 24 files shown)


stephenberry (commented):
Thanks for this. The two WebSocket findings are real and would justify the port on their own:

- the prefix-matched field names in the client handshake, where `strncasecmp(name.c_str(), "Upgrade", 7)` happily accepts `Upgrade-Extra`;
- `Connection` split across two fields being dropped by the map's last-wins. That form is legal (RFC 9110 §5.3) and proxies do emit it, so the server was refusing valid upgrades with no workaround available to users.

The multi-value support is the bigger win: a handler currently cannot emit two `Set-Cookie` fields at all. Forcing `Content-Length`/`Transfer-Encoding` through `set()` inside `add_header` is the right call, and the CL.CL rejection, the duplicate `Host` check and the CRLF drops at both the setter and the serializer all look preserved.

A few things before this goes in.

### 1. `post_json`/`put_json` now silently discard a caller-supplied Content-Type

```cpp
auto merged_headers = headers;
merged_headers.set("Content-Type", "application/json");
```

`set()` matches case-insensitively, so a caller passing `application/vnd.api+json`, or adding a `charset`, loses it. The old `merged_headers["content-type"] = ...` was also wrong (it emitted *two* Content-Type fields, since `"Content-Type"` and `"content-type"` were distinct map keys), so this is not a regression, but set-if-absent is the semantic we want:

```cpp
if (!merged_headers.contains("Content-Type")) {
   merged_headers.add("Content-Type", "application/json");
}
```

Three call sites: `post_json`, `put_json`, `post_json_async`.

### 2. No end-to-end test for repeated `Set-Cookie`

`http_headers_test` covers repeats at the container level, but nothing asserts that two `response::add_header("Set-Cookie", ...)` calls both survive `send_response_with_conn`'s serializer and reach the wire. That path is the whole point of `add_header`, and it is the one thing users will reach for first. A raw-socket assertion in `http_server_api_tests`, alongside the existing keep-alive ones, would cover it.

### 3. Header names are no longer lowercased, and that is a silent behavior change

Requests now keep the case as received, and `response::header()` emits what the handler passed, which is why the test moved from `x-echo:` to `X-Echo:`. Lookups stay case-insensitive so `find("content-type")` is unaffected, but user code that iterates headers and compares names with `==` breaks quietly rather than failing to compile. Worth a note in `http-rest-support.md` next to the `request`/`response` structs so it carries through to the release notes.

### 4. Doc examples should prefer the view accessors over `find()` + `->value`

`first_value()`, `contains_token()` and `values()` are the surface worth documenting. `find()` returning an iterator into the container pins `http_header::value` down as a concrete `std::string` harder than it needs to be, and it encourages patterns like `std::stoll(start_header->value)` that depend on that. The docs are mostly converted already, but `advanced-networking.md` (lines 210, 447), `http-router.md` (line 358) and `http-examples.md` (line 817) still show `auth_header->value`.

### 5. Follow-up, not a blocker

`build_http_request_bytes` writes `Host` and `Connection: keep-alive` unconditionally and then appends the caller's headers, so a caller passing either gets a duplicate on the wire. That was equally true with the map, so it is out of scope here, but a container that keeps repeats makes it easier to hit.

### 6. Nit

`reject_unwritable_field` returning `true` for "rejected" while also performing the `user_headers_set` bookkeeping as a side effect reads a little oddly at the call sites. Not important.

stephenberry:
I had AI do a pass and fix some things. I'm now going to do a manual pass over the code as soon as I have some dedicated time. But, I think this is just abut ready to be merged.

stephenberry:
@annihilatorq, we're both now stacking PRs on top of this. I think this is good to merge and we can start merging other improvements on top of it before we cut a release. Do you think merging this PR is a good next step?

annihilatorq:
> @annihilatorq, we're both now stacking PRs on top of this. I think this is good to merge and we can start merging other improvements on top of it before we cut a release. Do you think merging this PR is a good next step?

Yep, looks good to me!

PR #2729, opened 2026-07-30
fix out-of-bounds read in repe zero-copy read_params

uwezkhan:
`registry::call(std::span<const char>, std::string&)` parses the request in place, so `state.in.body` views the caller's span with nothing guaranteed after it, yet the `state_view` overload of `read_params` reads that view with the registry's `Opts`, whose `null_terminated` is on by default. That is the mode in which the JSON reader drops its end checks and scans for a `'\0'` sentinel. Before: a 51 byte buffer holding a well formed message whose body is `1` gives an ASAN heap-buffer-overflow read of size 1, zero bytes past the allocation; after: the scan stops at the end of the span.

The `message` overload beside it is unaffected because its body is a `std::string`, which does supply that sentinel, so the view overload is where the buffer shape is known and where the bound belongs rather than in the callers that own the span. Reading the body as unterminated input trades the sentinel scan for per-byte end checks. It also moves one edge case: a whitespace-only body now reads as an empty value instead of a number failure, which is what `glz::read` already returns for the same bytes with `null_terminated` off.

### DIFF
--- a/include/glaze/rpc/repe/repe.hpp
+++ b/include/glaze/rpc/repe/repe.hpp
@@ -554,7 +554,17 @@ namespace glz::repe
       }
       auto start = b;
 
-      glz::parse<Opts.format>::template op<Opts>(std::forward<Value>(value), ctx, b, e);
+      // The body is a view into the request span, so nothing is guaranteed to follow it.
+      // A null_terminated read drops its end checks and scans for the '\0' sentinel that
+      // only a std::string body would supply, so read the view as unterminated input and
+      // clear the end_reached a value finishing at the span end reports, matching what
+      // finalize_read_context does for glz::read.
+      static constexpr auto BodyOpts = opt_false<Opts, &opts::null_terminated>;
+
+      glz::parse<Opts.format>::template op<BodyOpts>(std::forward<Value>(value), ctx, b, e);
+      if (ctx.error == error_code::end_reached && ctx.depth == 0) {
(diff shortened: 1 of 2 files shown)


stephenberry:
Closing in favor of #2732, which has merged. Thank you for this — the diagnosis was right and it is what the fix was built on.

Your `exact_buffer` helper and both tests (`scalar_body_ending_at_buffer_end`, `object_body_ending_at_buffer_end`) are in `main` as written.

Two things pushed the fix wider than one call site, which is why it landed as a separate PR rather than as a merge of this one:

- The same defect was reachable through two more entry points that this PR does not touch: the JSON RPC `call(std::string_view)` overload reached `glz::read_json` and `glz::validate_json`, both of which hardcode default options and so read past an exactly sized request independently of `read_params`. #2732 makes the bound a property of the registry — it reads with `null_terminated` off regardless of the user's options — so all three are covered by one rule instead of three fixes.
- Turning `null_terminated` off surfaces a second defect that bites any version of this fix, including this one as written. `read_params` returned bytes consumed and every caller read `0` as failure, but without a terminator a variant alternative that resolves at the end of the buffer rewinds its iterator, so a completed read can report zero bytes. The registry took that for an error and sent no response at all to a well formed non-notify request. `read_params` returns `bool` now.

Neither is a criticism of the change here — the second one is only visible once you make the first one, which is exactly what this PR did. Appreciate the report and the tests.

PR #2733, opened 2026-07-31
Validate UTF-8 on read

### LINKED ISSUE #2720: Glaze seems to skip UTF8-validation.
So, it seems that running this code:
```
#include <glaze/glaze.hpp>
#include <iostream>
#include <string>
#include <vector>
#include <iomanip>

void printBytes(const std::string& bytes) {
	for (char c: bytes) {
		std::cout << std::hex << std::setw(2) << std::setfill('0') << (static_cast<int>(c) & 0xFF) << " ";
	}
	std::cout << std::dec << std::endl;
}

void runCase(const std::string& label, const std::string& bytes, bool expectValid) {
	std::vector<std::string> valueNew{};
	std::cout << "Case: " << label << std::endl;
	std::cout << "Bytes: ";
	printBytes(bytes);
	auto errorCode	= glz::read_json(valueNew, bytes);
	bool didSucceed = (errorCode == 0);
	std::cout << (didSucceed ? "Glaze ACCEPTED this input." : "Glaze REJECTED this input.") << std::endl;
	std::cout << "Expected valid: " << std::boolalpha << expectValid << " | Actually accepted: " << didSucceed;
	std::cout << (didSucceed == expectValid ? " -> MATCH" : " -> MISMATCH") << std::endl;
	std::cout << std::endl;
}

int main() {
	std::string loneContinuation{ "[\"a" };
	loneContinuation += static_cast<char>(0x80);
	loneContinuation += "b\"]";

	std::string overlongEncoding{ "[\"a" };
	overlongEncoding += static_cast<char>(0xC0);
	overlongEncoding += static_cast<char>(0x80);
	overlongEncoding += "b\"]";

	std::string truncatedMultibyte{ "[\"a" };
	truncatedMultibyte += static_cast<char>(0xE2);
	truncatedMultibyte += static_cast<char>(0x82);
	truncatedMultibyte += "b\"]";

	std::string invalidStartByte{ "[\"a" };
	invalidStartByte += static_cast<char>(0xFF);
	invalidStartByte += "b\"]";

	std::string surrogateHalfEncoded{ "[\"a" };
	surrogateHalfEncoded += static_cast<char>(0xED);
	surrogateHalfEncoded += static_cast<char>(0xA0);
	surrogateHalfEncoded += static_cast<char>(0x80);
	surrogateHalfEncoded += "b\"]";

	std::string validControl{ "[\"a" };
	validControl += static_cast<char>(0xE2);
	validControl += static_cast<char>(0x82);
	validControl += static_cast<char>(0xAC);
	validControl += "b\"]";

	runCase("Lone continuation byte (0x80)", loneContinuation, false);
	runCase("Overlong encoding of NUL (0xC0 0x80)", overlongEncoding, false);
	runCase("Truncated 3-byte sequence (0xE2 0x82, missing 3rd byte)", truncatedMultibyte, false);
	runCase("Invalid start byte (0xFF)", invalidStartByte, false);
	runCase("Encoded UTF-16 surrogate half (0xED 0xA0 0x80)", surrogateHalfEncoded, false);
	runCase("Valid 3-byte Euro sign (0xE2 0x82 0xAC) - sanity check", validControl, true);

	return 0;
}
```
Results in this output:
```
Case: Lone continuation byte (0x80)
Bytes: 5b 22 61 80 62 22 5d
Glaze ACCEPTED this input.
Expected valid: false | Actually accepted: true -> MISMATCH

Case: Overlong encoding of NUL (0xC0 0x80)
Bytes: 5b 22 61 c0 80 62 22 5d
Glaze ACCEPTED this input.
Expected valid: false | Actually accepted: true -> MISMATCH

Case: Truncated 3-byte sequence (0xE2 0x82, missing 3rd byte)
Bytes: 5b 22 61 e2 82 62 22 5d
Glaze ACCEPTED this input.
Expected valid: false | Actually accepted: true -> MISMATCH

Case: Invalid start byte (0xFF)
Bytes: 5b 22 61 ff 62 22 5d
Glaze ACCEPTED this input.
Expected valid: false | Actually accepted: true -> MISMATCH

Case: Encoded UTF-16 surrogate half (0xED 0xA0 0x80)
Bytes: 5b 22 61 ed a0 80 62 22 5d
Glaze ACCEPTED this input.
Expected valid: false | Actually accepted: true -> MISMATCH

Case: Valid 3-byte Euro sign (0xE2 0x82 0xAC) - sanity check
Bytes: 5b 22 61 e2 82 ac 62 22 5d
Glaze ACCEPTED this input.
Expected valid: true | Actually accepted: true -> MATCH
```
Unless, I'm missing an option or something, this seems as though it should be catching these inputs, no?

stephenberry:
Fixes #2720.

`glz::read_json` accepted malformed UTF-8 inside strings, so a successful parse gave no guarantee the resulting `std::string` held well-formed text:

```cpp
std::string s{};
// 0xC0 0x80 is an overlong encoding of NUL, illegal UTF-8
auto ec = glz::read_json(s, "\"\xC0\x80\""); // ec was success, and s held the invalid bytes
```

RFC 8259 §8.1 requires JSON text to be UTF-8, and the documentation already claimed validation was performed. Those claims are now true rather than removed.

## Behavior

Validation is unconditional on read and fails with `error_code::invalid_utf8`. It is not an option: read input is by definition someone else's data, and a correctness guarantee that must be switched on is one most callers will not have switched on.

Writing is deliberately left alone. A value obtained by reading is already validated, so a round trip needs no second pass. This does make round-tripping asymmetric: a `std::string` holding non-UTF-8 bytes writes successfully and then fails to read back.

**This is a behavioral break.** Validation covers every string the reader passes over, including keys, values you do not model, and values reached through `glz::skip` or `glz::raw_json`, so a malformed byte in a field you never look at fails the whole parse. JSONC comments are exempt, since their bytes never reach your program.

## Implementation

`include/glaze/simd/utf8_validation.hpp` implements the Lemire and Keiser nibble-lookup algorithm: three 16-entry table lookups whose AND is non-zero exactly when a byte pair is illegal, plus a saturating-subtract check for continuation obligations. The checker is written once against a small vector-operation interface and instantiated per backend — AVX-512BW (64 bytes), AVX2 (32), SSSE3 (16), AArch64 NEON (16), WASM SIMD128 (16) — with a portable C++ backend at 16/32/64 that makes a 64-byte register testable on hosts that cannot execute AVX-512 at all. Targets without a byte-granular shuffle fall back to a scalar validator.

The string scanners already load 8-byte chunks to find the closing quote, so they OR those chunks together and skip validation entirely when a string turns out to be pure ASCII. Keys and skipped strings are found with `memchr`, which proves nothing about the bytes, so those run a full pass.

**SWAR was tried first and abandoned.** It has no byte-granular table lookup, and an 8-byte window cannot see a sequence straddling its edge without carrying state that costs more than it saves. A complete SWAR validator measured 0.73×–1.28× against scalar.

## Performance

Validator throughput on 4 MB documents: 1.5–1.7 → 5.6–10.6 GB/s on non-ASCII, 16.6 → ~80 GB/s on ASCII.

End to end on `twitter.json` (22% non-ASCII, string heavy): about 5% into structs and 6% into `glz::generic`. A pure ASCII document is within noise of unvalidated parsing into structs, and about 3% into `glz::generic`.

## Tests and CI

`tests/json_test/utf8_validation_test.cpp`: 27 tests, 17,190 assertions — every Unicode code point, all 256 single bytes, every lead byte against every invalid follower, truncation at every cut point, a bad byte and a valid sequence at every offset, the full corpus embedded in an ASCII carrier at 15 offsets straddling every register and step boundary, and differential fuzz against the scalar path with fixed seeds. The `json_conformance` UTF-8 suite was rewritten for always-on behavior, and `fuzz_json_roundtrip_string` now asserts the new contract in both directions instead of filtering malformed inputs out.

`.github/workflows/simd-backends.yml` runs the suite against every backend, since a bug in one is invisible on the others. Every job asserts which backend the preprocessor actually chose before testing it, so a flag change that quietly downgrades to scalar fails loudly. AVX-512 uses the runner's own when present and Intel SDE otherwise, and fails rather than skipping when neither is available. `utf8_validation_test` was added to the armv7 QEMU list; the big-endian s390x job picks it up through `ctest`, which exercises the byteswap path in the ASCII accumulator on real hardware.

Verified with identical results on NEON64, SSSE3, AVX2, SSE2 baseline, scalar, the portable backend at all three widths, WASM SIMD128 under node, and AVX-512BW both natively and under SDE.


Curve:
I think it would make sense to make this optional - if you know that the data is valid, it would be nice to skip this validation.

stephenberry:
@Curve, I went back and forth. But, I agree that it should be a compile time option. Do you think it should be on or off by default?

Curve:
> @Curve, I went back and forth. But, I agree that it should be a compile time option. Do you think it should be on or off by default?

I think on by default is fine

PR #2743, opened 2026-08-01
Exclude msgpack_roundtrip_string from the OSS-Fuzz build

xingyaner:
Excludes fuzzing/msgpack_roundtrip_string.cpp from the OSS-Fuzz build in fuzzing/ossfuzz.sh.

This is required because clang-22 (OSS-Fuzz base image) segfaults when compiling this file with -std=c++23: a bare glz::write_msgpack (std::string, buffer) call triggers a clang bug (stack exhaustion in the partial ordering of constrained partial specializations), reproducible at -O0 through -O3. The file was introduced by commit 4cbddc3e and is the only affected fuzzer; all other fuzzers build fine, and the remaining msgpack fuzzers keep msgpack coverage.

Because the exclusion lives in the upstream fuzzing/ossfuzz.sh (the entry point already used by OSS-Fuzz), no change to the OSS-Fuzz project config is needed. Verified: OSS-Fuzz check_build (afl/address/x86_64) passes.
Note: once the clang-22 crash is fixed (upstream LLVM or a newer OSS-Fuzz base image), msgpack_roundtrip_string.cpp should be re-enabled by removing it from the exclusion.

### DIFF
--- a/fuzzing/ossfuzz.sh
+++ b/fuzzing/ossfuzz.sh
@@ -8,7 +8,7 @@ set -eux
 
 $CXX --version
 
-for SRCFILE in $(ls fuzzing/*.cpp |grep -v -E '(exhaustive|main\.cpp)'); do
+for SRCFILE in $(ls fuzzing/*.cpp |grep -v -E '(exhaustive|main\.cpp|msgpack_roundtrip_string\.cpp)'); do
     NAME=$(basename $SRCFILE .cpp)
     $CXX $CXXFLAGS -std=c++23 -g -Iinclude \
          $SRCFILE -o $OUT/$NAME \


stephenberry:
Thanks for the report — the stack trace you included is what made this findable, and it led to a cause in Glaze rather than only in clang.

`to<MSGPACK, T>` had three overlapping string specializations (`string_t`, `static_string_t`, `str_t`). Two are viable for `std::string`, which forces the compiler to partially order constrained partial specializations, and normalizing those concepts for the subsumption check is what exhausts clang 22's stack. Every other format uses a single `str_t` specialization, which is why msgpack was the only one affected.

Fixed in #2752 by collapsing the three into one. Verified with `clang 22.0.0git cb2f0d0a` in the OSS-Fuzz `base-builder` image: unmodified `fuzzing/ossfuzz.sh` builds all 22 fuzzers, including `msgpack_roundtrip_string`.

The exclusion is no longer needed, so I'd rather keep the coverage — closing this. It also turned up a separate string truncation bug in the other formats (#2753, fixed in #2754). Thanks again.

PR #2747, opened 2026-08-01
Adding cpu-backend names for querying the currently active backend.

RealTimeChris:
Pretty straight forward. It makes sense to add this I think, given that Jsonifier and simdjson both have ways to do this.


stephenberry:
Glaze will use multiple combinations of SIMD, so this wouldn't describe what's happening. A better solution was implemented here: #2747

But, thanks for bringing this idea to mind.

PR #2758, opened 2026-08-03
validate payloads in jsonb_to_json before copying them out

uwezkhan:
`jsonb_to_json` copies the payload straight into the output document for every type the spec marks as already valid JSON text: INT, FLOAT, TEXT, TEXTJ, and the FLOAT5 fallback. That claim holds for blobs Glaze wrote, since `to<JSONB, str_t>` pre-scans with `string_needs_json_escape` and downgrades TEXT to TEXTRAW, and `read_jsonb` checks it on its own path, but the converter is aimed at SQLite BLOB columns that any writer can fill. Before: a TEXT value whose payload is `x","b":"y` renders a one-member object as `{"a":"x","b":"y"}`, an INT payload of `1,"b":9` does the same with no string involved, and payloads like `a\`, a raw control byte, or a lone `\uDE00` produce output that is not JSON while the call still reports success. An object key was unconstrained too, so a blob with an array in the key slot emitted `{{}:false}`.

After: each of those arms checks its payload against the JSON grammar before copying, and the key slot has to be one of the text types. Checking rather than re-encoding keeps the blob's exact decimal representation, so `1e10` and `-0.0` still come out byte for byte, and TEXTJ still passes through unchanged. FLOAT5 is the one arm whose shape changes: the JSON5 syntax it is left with (`.5`, `5.`, `+1.5`) is parsed and re-rendered the way INT5 already is, because none of it is legal JSON as written. The tradeoff is one pass over each payload where TEXT previously cost nothing, which is the same scan the writer already performs to pick TEXT over TEXTRAW.

### DIFF
--- a/include/glaze/jsonb/jsonb_to_json.hpp
+++ b/include/glaze/jsonb/jsonb_to_json.hpp
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <charconv>
 #include <cstddef>
 #include <cstdint>
 #include <cstring>
@@ -14,7 +15,7 @@
 #include "glaze/json/write.hpp"
 #include "glaze/jsonb/header.hpp"
 #include "glaze/jsonb/text_decode.hpp"
-#include "glaze/util/fast_float.hpp"
+#include "glaze/util/parse.hpp"
 
 namespace glz
 {
@@ -22,11 +23,33 @@ namespace glz
    {
       // Emit a JSON string literal from a raw UTF-8 byte payload. Uses the JSON writer so all
       // control characters and structural JSON chars are correctly escaped.
+      //
+      // Three writer options would each let payload bytes reach the output document unescaped,
+      // and all three are inheritable, so a caller's opts would otherwise silently reopen the
+      // hole this converter exists to close:
+      //   * escape_control_characters (off by default) leaves raw control bytes in place,
+      //   * raw_string emits the payload without escaping it at all,
+      //   * unquoted drops the surrounding quotes as well.
+      // Each default is reasonable for a C++ string being serialized, because that is the
+      // program's own data. None of them is reasonable for a blob, which is someone else's.
+      // The converter promises strict JSON, so it pins all three rather than inheriting them.
+      template <auto Opts>
+      inline constexpr auto raw_string_emit_opts =
+         opt_false<opt_false<opt_true<Opts, escape_control_characters_opt_tag{}>, raw_string_opt_tag{}>,
+                   unquoted_opt_tag{}>;
+
       template <auto Opts, class B>
       inline void emit_raw_string_as_json(is_context auto& ctx, const char* data, size_t size, B& out, size_t& ix)
       {
+         // RFC 8259 section 8.1 requires JSON text to be UTF-8, and these bytes came out of a
+         // blob rather than out of the program, so they get the same treatment the JSON reader
+         // gives its input. Honors the `validate_utf8` option, which compiles the check away.
+         // No accumulator is available here: nothing has scanned this payload yet.
+         if (validate_utf8_span<Opts>(ctx, data, data + size)) [[unlikely]] {
+            return;
+         }
          const sv s{data, size};
-         to<JSON, sv>::template op<Opts>(s, ctx, out, ix);
+         to<JSON, sv>::template op<raw_string_emit_opts<Opts>>(s, ctx, out, ix);
       }
 
       // The spec marks several payloads as "already valid JSON text" so that a converter can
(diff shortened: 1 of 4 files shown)


stephenberry:
Thanks for this — the splicing hazard is real and the direction is right. I verified the cases you describe against `main`, and also that SQLite's own `json()` emits `{1:true}` and `"a"bc"` for the equivalent blobs, so this ends up stricter than the reference implementation rather than merely matching it.

I've pushed a commit to the branch rather than leaving review comments, since a few of the changes are entangled. Summary of what changed and why:

**FLOAT5 is normalized textually instead of re-rendered.** This is the one arm I'd push back on in the original. Parsing to a `double` and re-rendering abandons the exactness principle the rest of the PR is careful about — and does so where it still applies. Concretely, feeding back a blob SQLite produces from `[+1e400, .12345678901234567890123]`:

```
sqlite json():  [1e400, 0.12345678901234567890123]
PR as written:  [1e400, 0.12345678901234568]        <- 23 significant digits down to 17
```

and `+1e400` on its own was rejected with `parse_number_failure`, because `fast_float` reports `result_out_of_range` — though the payload is perfectly good JSON once the `+` is gone. There was a leak the other way too: `fast_float` accepts `nan`, `inf` and `infinity`, none of which are JSON5, so those converted to `null` instead of erroring.

SQLite normalizes textually, so the commit does the same: strip a leading `+`, then supply the digit the payload omits on whichever side of the `.` lacks one. Worth noting the omitted digit is not always trailing — `5.e3` becomes `5.0e3`, which a leading/trailing rule would miss. The result is then held to `is_json_number` before being copied, so `.` and `.e5` (not JSON5 either) are rejected rather than normalized into a number they never denoted. Output now matches `json()` byte for byte on every FLOAT5 shape SQLite produces.

**TEXTRAW and TEXT5 were still emitting non-JSON.** Both hand the payload to the JSON string writer, which leaves control characters unescaped by default — so a raw control byte was copied straight into the output document, the same defect class the PR exists to close, just via a different arm. Two further inheritable options, `raw_string` and `unquoted`, would have let a caller's opts reopen it entirely. All three are now pinned for this converter. Those defaults are reasonable for a C++ string being serialized, because that is the program's own data; they are not reasonable for a blob.

**None of the four text arms checked encoding.** RFC 8259 §8.1 requires JSON text to be UTF-8, so a payload carrying malformed UTF-8 produced output that is not JSON while the call reported success. All four now validate, honoring the inheritable `validate_utf8` option so callers whose encoding is guaranteed upstream can still opt out. Since `is_json_string_body` already visits every byte, it accumulates them for `validate_utf8_span`, which skips the pass outright for a pure ASCII payload — the case TEXT exists to make cheap.

That surfaced a decoding bug worth calling out separately: JSON5 `\xNN` denotes the code point U+00NN, but the decoder was pushing the raw byte, so `\xFF` produced a lone `0xFF` that is never valid UTF-8. It is now UTF-8 encoded, matching how SQLite reads it. This also fixes `read_jsonb`, which shares the decoder, and does change decoded values for existing TEXT5 blobs — worth a changelog line.

**One writer-side fix.** `to<JSONB, map>` and `to<JSONB, pair>` serialize the key through the generic value writer, so `std::map<int, int>` put an INT in the key slot. `from<JSONB, T>` already `static_assert`s against non-string keys, so the writer was producing blobs the reader cannot be instantiated to read, and which SQLite renders as `{1:2}`. The writer now asserts the same condition. This is a source-breaking change for code that currently compiles, so it needs a changelog note.

I deliberately did **not** extend this to reflected `glz::meta` keys. A key containing a quote or backslash is a developer error, and Glaze requires callers to supply valid keys; it isn't worth a compile-time check. One consequence to be aware of: such a key now fails conversion with `syntax_error` instead of silently emitting invalid JSON.

On the lone-surrogate question — `is_json_string_body` rejects an unpaired `\uD800`, which RFC 8259 §7's grammar permits and SQLite passes through. I kept your stricter behavior, because Glaze's own reader rejects unpaired surrogates and accepting one here would mean emitting a document `glz::read_json` refuses. I added a test so the divergence is explicit rather than incidental.

Verification: `jsonb_test` 160/160, `jsonb_sqlite_test` 36/36, plus `chrono_test` and `embedded_null_test`. Differential-tested against SQLite 3.51 on blobs SQLite itself produces. Fuzzed the central claim — success implies valid JSON — over random and structured blobs: 466,055 successful conversions, every output validated by an independent parser, zero invalid.

Two unrelated bugs found along the way, which I'll handle separately: `glz::validate_json` rejects `0e4` (valid per RFC 8259 §6; `read_json<double>("0e4")` accepts while `read_json<double>("[0e4]")` rejects), and `glz::write_json` emits invalid JSON for a `glz::meta` key containing a quote.

uwezkhan:
The FLOAT5 pushback is right. Re-rendering through a double broke the exactness rule the rest of the converter keeps, and I hadn't considered that fast_float accepting inf/nan leaked non-JSON5 spellings through. Textual normalization held to is_json_number afterward is the better shape, and the same goes for pinning the three inheritable writer options, which was a hole I missed entirely. Also agreed on unpaired surrogates: success while emitting a document glz::read_json refuses would be a strange invariant for this converter. Ran the branch here, jsonb_test and jsonb_sqlite_test both pass.

PR #2761, opened 2026-08-05
Support ERL_BINARY_EXT tag in eetf_to_json

fregate:
Add support for ERL_BINARY_EXT tag. It writes base64 string as JSON field value

### DIFF
--- a/cmake/FindErlang.cmake
+++ b/cmake/FindErlang.cmake
@@ -28,10 +28,6 @@ This will define the following variables:
 
 ``Erlang_FOUND``
   True if the system has the Erlang library.
-``Erlang_RUNTIME``
-  The path to the Erlang runtime.
-``Erlang_COMPILE``
-  The path to the Erlang compiler.
 ``Erlang_EI_PATH``
   The path to the Erlang erl_interface path.
 ``Erlang_ERTS_PATH``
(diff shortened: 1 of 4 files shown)


stephenberry (commented):
Thanks for taking this on — `ERL_BINARY_EXT` is a real gap in `eetf_to_json` and the bounds checking here is careful and matches the surrounding style. I built the branch locally against OTP 29, ran the suite (38/38 pass), and probed the new path with real `term_to_binary/1` output under ASan/UBSan. A few things I'd like to work through before merging.

## 1. Base64 as the only representation

The mechanics are right, but I don't think base64 should be what most users get. In Erlang and especially Elixir, binaries *are* the string type, so this is what a typical term produces today:

```erlang
term_to_binary(#{name => <<"Bob">>, age => 30})
```
```json
{"name":"Qm9i","age":30}
```

Every mainstream Erlang JSON library (jsx, jsone, thoas, OTP's `json`) maps `ERL_BINARY_EXT` to a JSON string. Base64 is the right answer for opaque byte blobs and the wrong one for the far more common case of a UTF-8 payload.

Related, and made worse by base64: binary map keys still fail. `#{<<"name">> => <<"Bob">>}` — the canonical Elixir map shape — returns `syntax_error` with "unsupported key type", because `eetf::is_string` in `include/glaze/eetf/types.hpp` doesn't include `ERL_BINARY_EXT`. If binaries were written as JSON strings they'd be usable as keys; as base64 they'd be unreadable keys.

### Suggested alternative

Make it an option, defaulting to the string form, with base64 opt-in. In `eetf/opts.hpp`:

```cpp
struct eetf_opts
{
   // ...
   bool binary_as_base64 = false; // write ERL_BINARY_EXT as base64 instead of a JSON string
};
```

plus a `check_binary_as_base64(auto&& Opts)` helper following the `check_*` convention in `core/opts.hpp`, so other option structs keep working. Then:

```cpp
case ERL_BINARY_EXT: {
   ++it; // skip type
   const size_t len = get32be(ctx, it, end);
   if (bool(ctx.error)) [[unlikely]] {
      return;
   }
   // bound binary length
   if (check_invalid_offset(ctx, it, end, len)) return;
   if constexpr (check_binary_as_base64(Opts)) {
      dump('"', out, ix);
      write_base64_to(ctx, reinterpret_cast<const uint8_t*>(it), len, out, ix);
      if (bool(ctx.error)) [[unlikely]] {
         return;
      }
      dump('"', out, ix);
   }
   else {
      const sv value{reinterpret_cast<const char*>(it), len};
      to<JSON, sv>::template op<Opts>(value, ctx, out, ix);
   }
   std::advance(it, len);
   break;
}
```

The string branch then reuses the same escaping path as the `ERL_STRING_EXT` case, and it opens the door to accepting `ERL_BINARY_EXT` as a map key in the `ERL_MAP_EXT` case (happy for that to be a follow-up PR rather than this one).

## 2. Smaller points on `eetf_to_json.hpp`

- `reinterpret_cast<const uint8_t*>(&*it)` dereferences a past-the-end pointer when `len == 0` — which is exactly what the new "empty binary" test exercises. It's harmless in practice and sanitizers won't flag it (no load is performed), but it is formally UB. Every other case in the file uses `reinterpret_cast<const char*>(it)` without the `&*`; worth matching.
- `write_base64_to` can set `ctx.error` through `ensure_space`, but the code writes the closing quote and advances regardless. The sketch above adds the check.
- `#include <string_view>` is added but nothing in the new code uses it (it would be needed by the `sv` branch above).
- Test-name typo: "trucated" → "truncated", including `trucated_size` / `trucated_body`.

Setting `ctx.custom_error_message = "unsupported type"` on the default case is a nice addition, and hoisting `expect_unexpected_end` out of the four test bodies is a good cleanup.

The bounds checking itself looks correct to me: `get32be` guards the 4-byte length, `check_invalid_offset` guards the payload, and a declared length of `0xFFFFFFFF` with no payload correctly yields `unexpected_end`.

## 3. The `FindErlang.cmake` rewrite should be its own PR

Commit "Change the way how to locate Erlang libraries" is unrelated to the feature, and I'd rather review it separately — it also has some issues I'd want addressed first.

**The cross-compile path it adds doesn't work.** When `ERLANG_EI_LIB`/`ERLANG_EI_INCLUDE` are set without `ERLANG_ROOT`, the `else()` branch reads `${ERLANG_ROOT}`, which is empty by construction in that branch:

```
$ cmake -DERLANG_EI_LIB=/some/lib -DERLANG_EI_INCLUDE=/some/include
-- ERLANG_ROOT="" ; ERLANG_EI_LIB="/some/lib" ; ERLANG_EI_LIB="/some/lib"
-- Found Erlang: /some/include
-- Erlang::Erlang include = [//usr/include]
```

`Erlang::Erlang` is linked into `glaze_glaze`, so every consumer would get `-I//usr/include`.

**Debug output ships to consumers.** `cmake_print_variables(ERLANG_ROOT ERLANG_EI_LIB ERLANG_EI_LIB)` prints on every configure (and lists `ERLANG_EI_LIB` twice — presumably `ERLANG_EI_INCLUDE` was intended). `cmake/install-rules.cmake` embeds this file verbatim into the installed `glazeConfig.cmake`, so downstream users see it too.

**`FIND_PACKAGE_HANDLE_STANDARD_ARGS` misuse.** With the bare-variable signature the first argument is the *failure message*, so `Erlang_EI_LIBRARY_PATH` is being consumed as a message string rather than a required var — which is why the output above reads `Found Erlang: /some/include`. Combined with the unconditional `set(Erlang_FOUND ON)` immediately before, the found-check no longer does anything. `REQUIRED_VARS` would fix this.

**Behavior regressions:**

- `$ENV{ERLANG_HOME}/bin` is dropped from the search paths, which breaks anyone relying on `ERLANG_HOME`.
- `Erlang::ERTS`, `Erlang_OTP_VERSION`, `Erlang_EI_PATH`, and `Erlang_ERTS_*` are removed while the module's own `.rst` doc block still documents them.
- The `if(Erlang_FOUND)` guard around target creation is gone, so the imported targets are now created even when discovery fails.
- The new input variables (`ERLANG_ROOT`, `ERLANG_EI_LIB`, `ERLANG_EI_INCLUDE`) aren't documented in the doc block. Note also that `ERLANG_ROOT` collides with CMake's own `<PackageName>_ROOT` convention (CMP0074); something like `Erlang_ROOT_DIR` would avoid the overlap.

CI is green, but only the default discovery path is exercised there — the cross-compile branch isn't covered by any job.

Happy to help with any of this if you'd like to split the CMake work out and iterate on the option design here.

fregate:
Thank for the detailed review. I'll fix this soon

stephenberry:
@fregate, do you think this is ready to merge? I think its good, so I'll merge it if I get your thumbs up!

PR #2772, opened 2026-08-06
Fix tests with OpenSSL 4, backwards compatible with OpenSSL 3

raviksharma:
Fix tests with OpenSSL 4, backwards compatible with OpenSSL 3

Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/glaze/+bug/2162954


stephenberry:
Thanks for this!

PR #2778, opened 2026-08-09
Stop duplicating Host and Connection headers when the caller supplies them

### LINKED ISSUE #2777: http_client sends Host and Connection twice when the caller supplies them
Found by @stephenberry during the review of #2718 

---

`build_http_request_bytes` ([http_client.hpp:262](https://github.com/stephenberry/glaze/blob/72b670e2cf4e1c91bf61e0f3014e1e05ffbeaa3f/include/glaze/net/http_client.hpp#L262)) writes `Host` and `Connection: keep-alive` unconditionally and then appends the caller's headers, so a caller passing either gets a duplicate on the wire.

Same for `Content-Length` on requests with a body, where a second field is a CL.CL desync.

annihilatorq:
Follow-up to #2718
Fixes #2777

### DIFF
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -274,17 +274,21 @@ namespace glz
          request_str.append(" ");
          request_str.append(url.path);
          request_str.append(" HTTP/1.1\r\n");
-         request_str.append("Host: ");
-         request_str.append(url.host);
-         if (!is_default_port) {
-            char
-               port_buf[8]; // a uint16_t port is at most 5 digits; pad so the sizing does not depend on itoa internals
-            auto* end = glz::to_chars(port_buf, url.port);
-            request_str.push_back(':');
-            request_str.append(port_buf, static_cast<size_t>(end - port_buf));
+         if (!headers.contains("Host")) {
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks @annihilatorq — you got here first, and the core of the fix (suppress the writer's default when the caller supplies the field) is what shipped.

I've opened #2805, which builds on this and #2804 and closes the remaining gaps. You're credited as a co-author on the commit.

Two things came out of review that #2805 handles:

- `headers.contains("Host")` matches on name alone, but the header loop drops a field whose value carries CR or LF. So a caller `Host: evil\r\nX-Injected: 1` suppresses the derived Host and is then dropped, and the request goes out with **no** Host at all — an RFC 9112 3.2 MUST violation that glaze's own server answers with 400. #2805 only treats a caller field as supplied if it is actually writable.
- `glz::http_headers` keeps repeats, so two caller `Host` fields both reach the wire. #2805 reduces Host and Connection to one field each, the way the function already does for Content-Length.

Your `count_fields` test helper was the better shape of the two PRs, and #2805 uses the same idea. Closing in favor of #2805 — thanks again for the fix.

PR #2779, opened 2026-08-10
quote integer map keys in cbor_to_json

uwezkhan:
`glz::cbor_to_json` on a CBOR map with integer keys, before:

    a1 01 02           ->  {1:2}         (returns success)
    a2 01 02 61 61 03  ->  {1:2,"a":3}   (returns success)

Neither output re-parses as JSON, because a map key has to be a string. Integer keys are the norm for COSE/CWT/WebAuthn and other RFC 8949 profiles, so this is ordinary input, and the converter reported no error on it.

Before: every map key ran through the value dispatcher, so a non-string key was written in its native form (a bare number, or `[...]`/`{...}`).

After: `cbor_to_json_key` emits a text or byte-string key as before (byte strings keep their quoted-hex form), quotes an unsigned or negative integer key as a decimal string (one of the string forms RFC 8949 6.1 permits a converter to choose, matching what `beve_to_json` does for integer object keys), unwraps a tag on a key and re-checks the tagged content (so a tag-0 datetime key still emits as a plain string), and rejects a key type that has no JSON string form. `cbor_to_json_key` checks the recursion depth limit, so a tag chain on a key cannot recurse unboundedly.

Tradeoff: a key that is an array, map, float, bool, or typed-array tag was previously written as invalid JSON with a success code, and is now a `syntax_error`. Text-keyed and byte-string-keyed maps are byte for byte unchanged, and tagged keys keep their prior output.

### DIFF
--- a/include/glaze/cbor/cbor_to_json.hpp
+++ b/include/glaze/cbor/cbor_to_json.hpp
@@ -78,6 +78,9 @@ namespace glz
       }
 
       template <auto Opts, class Buffer>
+      inline void cbor_to_json_key(auto&& ctx, auto&& it, auto&& end, Buffer& out, auto&& ix, uint32_t recursive_depth);
+
+      template <auto Opts, class Buffer>
       inline void cbor_to_json_value(auto&& ctx, auto&& it, auto&& end, Buffer& out, auto&& ix,
                                      uint32_t recursive_depth)
       {
@@ -344,7 +347,7 @@ namespace glz
                   first = false;
 
                   // Key (must be string for JSON compatibility)
-                  cbor_to_json_value<Opts>(ctx, it, end, out, ix, recursive_depth + 1);
+                  cbor_to_json_key<Opts>(ctx, it, end, out, ix, recursive_depth + 1);
                   if (bool(ctx.error)) [[unlikely]]
                      return;
 
@@ -376,7 +379,7 @@ namespace glz
                   }
 
                   // Key
-                  cbor_to_json_value<Opts>(ctx, it, end, out, ix, recursive_depth + 1);
+                  cbor_to_json_key<Opts>(ctx, it, end, out, ix, recursive_depth + 1);
                   if (bool(ctx.error)) [[unlikely]]
                      return;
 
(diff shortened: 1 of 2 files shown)


stephenberry (changes requested):
Thanks for this — the core fix is right and I want to take it. Integer-keyed maps producing invalid JSON with a success code is a real bug, and quoting them matches what `beve_to_json_value` already does for its integer key types. I built the branch and confirmed the round-trip now closes:

```
main:  cbor_to_json -> {-3:4,1:2,500:6}       read_json -> parse error
PR:    cbor_to_json -> {"-3":4,"1":2,"500":6} read_json -> round-trips exactly
```

Prettify, indefinite-length maps, and indefinite-length text keys all behave correctly, and the CBOR suite passes.

Two things to fix before merge.

### 1. Byte-string and tagged-text keys regress from valid JSON to `syntax_error`

The description says non-string keys were "previously written as invalid JSON with a success code." That holds for arrays, maps, floats, and bools, but not for byte strings or tags. The value path already emits a `bstr` as a quoted hex string, so those keys produced valid JSON on `main`. Measured against both branches:

| key | main | this PR |
|---|---|---|
| `h'01ff'` | `{"01ff":2}` (valid) | `syntax_error` |
| `0("2013-03-21T20:04:00Z")` | `{"2013-03-21T20:04:00Z":1}` (valid) | `syntax_error` |
| `[1]`, `1.5`, `true` | invalid JSON | `syntax_error` (correct) |

Tag-0 datetime keys are legitimate, and tags are pervasive in exactly the COSE/CWT profiles this PR targets, so this needs to keep working:

- Add `case major::bstr:` alongside `major::tstr` in `cbor_to_json_key` — it delegates to `cbor_to_json_value`, which already quotes the hex, so prior behavior is restored as-is.
- For tags, consume the tag argument and recurse into `cbor_to_json_key` for the tagged content. A typed-array tag as a key should still be rejected.

### 2. `cbor_to_json_key` needs the recursion-depth guard

It currently has no `recursive_depth >= max_recursive_depth_limit` check. That is safe as written — `tstr` delegates to `cbor_to_json_value`, which checks, and `uint`/`nint` are terminal — but it becomes a live problem the moment tag handling is added, since a chain like `c0 c0 c0 ...` as a key would recurse unboundedly. Please add the guard as part of the tag change.

### Minor

- `static_cast<int64_t>(~n)` wraps for `n >= 2^63`, so CBOR `-(2^63+1)` emits `"9223372036854775807"`. This is pre-existing — the same bug is in the value path — so it is out of scope here; I'll track it separately. No need to address it in this PR.
- The RFC 8949 §6.1 citation is a little strong: the RFC permits a converter to stringify non-text keys but does not prescribe the decimal form. Worth softening the wording in the description.

Quoting can also produce duplicate JSON keys (`{1:1,"1":2}` becomes `{"1":1,"1":2}`). That is inherent to the conversion and the RFC calls it out as non-reversible, so it is fine to leave.

Please add tests covering the byte-string and tagged-text key cases so they stay pinned.

uwezkhan:
Good catch on the byte-string and tag keys, the "previously invalid" claim in my description was too broad. Pushed ec41574 with both changes done the way you outlined: `bstr` sits alongside `tstr` in `cbor_to_json_key` and delegates to the value path, so `h'01ff'` keys keep their quoted-hex form, and a tag key consumes the tag argument and recurses into `cbor_to_json_key`, so a tag-0 datetime key emits as a plain string again while typed-array tags stay rejected. The depth guard is in as well; a 300-deep `c0` chain as a key now returns `exceeded_max_recursive_depth` instead of recursing.

Tests added pinning the bstr key, tag-0 datetime key, typed-array-tag rejection, and the tag chain. Also softened the RFC 8949 6.1 wording in the description, since quoted decimal is one permitted form rather than the prescribed one. Agreed the `~n` wrap on -(2^63+1) is the value path's pre-existing behavior, so I left it for the separate issue. `ctest -R ^cbor_test$` passes.

PR #2782, opened 2026-08-10
Only pool a streaming socket that reached the end of its response

stephenberry:
Found while reviewing #2718. Independent of it — this touches the streaming disconnect wrapper and `http_stream_connection`, neither of which #2718 modifies.

## The problem

`perform_stream_request` wraps the caller's disconnect handler in one that returns the socket to the pool:

```cpp
// Return the connection to the pool for reuse
if (connection->socket) {
    connection_pool->return_connection(url.host, url.port, use_https, std::move(*connection->socket));
}
```

Every way a stream can end runs through that handler. It is reached after:

- a connect/response timeout — `disconnect()` cancels pending operations but does not close the socket
- the caller calling `conn->disconnect()` part way through the body
- a mid-stream read error
- the EOF that ends a body framed by connection close

In each of those the socket is either dead or still has response bytes arriving on it. `return_connection` only checks `socket_is_open`, and none of these leave it closed, so all four get pooled.

The streaming path shares its pool with ordinary requests, so the next request to draw that socket reads the tail of the previous response as its own status line — the same desync a conflicting `Content-Length` produces, reached through the pool instead of through a header.

## The fix

A stream is reusable in exactly one case: the response has been read to the end of its framing, so the socket sits at the start of whatever the server sends next. `http_stream_connection` records that with a `response_complete` flag, and the disconnect wrapper closes the socket in every other case.

Reaching the end of the framing takes more than the terminal chunk. Two things had to be consumed or checked before the flag can be set:

**The trailer section.** The terminal chunk is not the end of the message — the trailer section and the final CRLF follow it (RFC 9112 §7.1.2). Setting the flag at the `chunk_size == 0` branch pooled the socket with those bytes still in front of the next response, which is the same desync one line further along. `consume_trailers` now reads trailer lines to the closing empty line first, and only then marks the response complete. This mirrors what the non-streaming paths already do via `async_consume_trailers`; the streaming reader was the one place that skipped it. A read failure there is not surfaced to the caller: the body is already delivered in full, so it only means the socket cannot be reused.

**`Connection: close`.** The streaming path never checked it, so a cleanly framed response from a server that is about to hang up still went into the pool. The async path checks it before `return_connection`; streaming now does too. Such a response is single-use however cleanly the body ended.

Deliberately unchanged: streaming applies no `Content-Length` framing at all. That is by design — the streaming API hands raw bytes to the caller — so it is not treated as a defect here.

## Testing

Two cases, one per direction of the rule.

`eof_framed_stream_not_pooled` covers a stream whose body is framed by connection close: the follow-up request has to open a fresh connection rather than draw the EOF'd socket.

`chunked_stream_pools_only_after_trailers` covers the reusable case. The server writes the trailer section in a separate segment, so those bytes are still on the socket when the terminal chunk is parsed rather than sitting in the client's buffer — which is what makes the bug reproducible instead of accidentally working. The follow-up request must then succeed **on the reused connection** (`accept_count == 1`), so the test pins both that a clean stream is pooled and that it is positioned correctly. It fails against the terminal-chunk-only version of the flag.

Both use `POST` for the follow-up request, not `GET`: transparent retry covers idempotent methods, so a `GET` quietly reopens on a dead socket and hides the bug — which is how the first draft of the EOF test passed against it.

Full networking suite green: `http_content_length_test` 26/26, `http_chunked_test` 59/59, `http_client_test` 39/39, `http_client_pool_test` 14/14. Clean under ASan and UBSan; TSan reports only the pre-existing asio kqueue-reactor warnings that also appear on unmodified `main`. The SSL variant syntax-checks under `-DGLZ_ENABLE_SSL`.

## Changes since the first push

- Rebased onto current `main` (the branch predated the `ssl.hpp` / `http_headers.hpp` extraction).
- Fixed the build failure that took out all twelve GCC/MSVC/asan jobs: the new test initialized `.on_disconnect` after `.on_data` and `.on_error`, but `stream_request_params_v2` declares `on_disconnect` first, and designators must appear in declaration order. Clang accepts the reordering as an extension, which is why every clang job was green. #2819 reorders the struct so the natural spelling is the correct one; whichever of the two merges second needs those designators flipped.
- Added the trailer handling, the `Connection: close` check, and the second test described above.

### DIFF
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -902,6 +902,13 @@ namespace glz
       std::shared_ptr<asio::streambuf> buffer; // Use unified streambuf for all reads
       bool is_connected{false};
       std::atomic<bool> should_stop{false};
+      // Set only once the response has been read to its framed end, leaving the
+      // socket positioned at the start of whatever the server sends next. Every
+      // other way a stream ends - a timeout, a caller-initiated disconnect part way
+      // through the body, a read error, or a body framed by connection close -
+      // leaves either unread response bytes or a dead peer behind, so the socket
+      // must be closed rather than handed to the next request.
+      std::atomic<bool> response_complete{false};
       stream_read_strategy strategy{stream_read_strategy::bulk_transfer}; // Default strategy
       std::function<bool(int)> status_is_error{}; // Evaluated before treating status as failure
       bool is_https{false}; // Track if this is an HTTPS connection
@@ -1426,9 +1433,21 @@ namespace glz
             if (user_on_disconnect) {
                user_on_disconnect();
             }
-            // Return the connection to the pool for reuse
             if (connection->socket) {
-               connection_pool->return_connection(url.host, url.port, use_https, std::move(*connection->socket));
+               // Only a socket sitting at the end of a fully-read response can serve
+               // the next request. Pooling one that timed out, errored, or was
+               // abandoned part way through the body hands the next request a socket
+               // with the tail of this response still arriving on it, which that
+               // request reads as its own status line - the same desync a conflicting
+               // Content-Length produces, reached through the pool instead. A body
+               // framed by connection close leaves the socket at EOF, which is dead
+               // rather than dangerous, but equally unusable.
+               if (connection->response_complete.load(std::memory_order_relaxed)) {
+                  connection_pool->return_connection(url.host, url.port, use_https, std::move(*connection->socket));
+               }
+               else {
+                  detail::close_socket(*connection->socket, connection_pool->graceful_ssl_shutdown());
+               }
             }
          };
 
@@ -1726,7 +1745,9 @@ namespace glz
                      connection->buffer->consume(bytes_transferred);
 
                      if (chunk_size == 0) {
-                        // Last chunk
+                        // Terminal chunk: the response is framed to its end here, so this
+                        // is the one exit that leaves the socket reusable.
+                        connection->response_complete.store(true, std::memory_order_relaxed);
                         if (on_disconnect) on_disconnect();
                         return;
                      }
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -1433,10 +1433,12 @@ namespace glz
          auto internal_on_disconnect = [this, user_on_disconnect = std::move(on_disconnect), connection, url,
                                         use_https]() {
             connection->is_connected = false;
-            // Call the user's handler if provided
-            if (user_on_disconnect) {
-               user_on_disconnect();
-            }
+            // Settle the socket before telling the caller the stream ended, matching the
+            // non-streaming paths, which return the connection and only then run the
+            // completion handler. The other order publishes "finished" while the socket is
+            // still in hand, so a caller that starts its next request from on_disconnect
+            // finds an empty pool and dials a second connection - never reusing the very
+            // socket the code below is about to make reusable.
             if (connection->socket) {
                // Only a socket sitting at the end of a fully-read response can serve
                // the next request. Pooling one that timed out, errored, or was
@@ -1452,6 +1454,12 @@ namespace glz
                else {
                   detail::close_socket(*connection->socket, connection_pool->graceful_ssl_shutdown());
                }
+            }
+            // Returning the socket leaves a null shared_ptr in the variant; every accessor
+            // guards on it, so the disconnect() that http_stream_connection runs from its
+            // destructor is a no-op rather than a dereference of a moved-from socket.
+            if (user_on_disconnect) {
+               user_on_disconnect();
             }
          };
 
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -902,13 +902,17 @@ namespace glz
       std::shared_ptr<asio::streambuf> buffer; // Use unified streambuf for all reads
       bool is_connected{false};
       std::atomic<bool> should_stop{false};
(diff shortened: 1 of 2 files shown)


wilkolbrzym-coder:
Quick review notes – PR #2782

**Blocking: CI is red** — 12 jobs failing (GCC 13/14/15/16, gcc-old-abi, gcc-x86, MSVC Debug/Release, sanitizers asan). Investigate before merge.

The core idea is correct: **`response_complete`** gates pooling so only a socket at the end of a fully-read chunked response is returned to the pool. All other paths close the socket.

However, verify that the chunk parser consumes the final CRLF and any **trailers** after `chunk_size == 0` before setting **`response_complete`**. If not, the socket still has bytes in front of the next response, which is exactly the desync this PR fixes. A test with chunked trailers would pin this.

The new **`eof_framed_stream_not_pooled`** test is well designed: POST avoids retry masking, and disabling liveness check is necessary.

Minor: **`memory_order_relaxed`** is fine for this single flag.

Overall approach LGTM once CI is investigated and trailer handling is confirmed.

stephenberry:
Thanks — the trailer question was the right one to ask, and the answer was no.

The parser consumed the `0\r\n` line and stopped there, so `response_complete` was set with the trailer section and the final CRLF still in front of the next response. That is the same desync this PR is about, just one line further along. It hides easily: when a server writes the whole response in one segment those bytes land in the connection's streambuf, which is discarded along with the connection, so the socket happens to be clean. Split the write and it breaks. Worth noting the non-streaming paths already got this right via `async_consume_trailers` — the streaming reader was the one place that skipped it.

Fixed by `consume_trailers`, which reads trailer lines to the closing empty line before marking the response complete. A read failure there is not reported to the caller: the body is already delivered in full, so it only means the socket cannot be reused.

Your suggested test is in as `chunked_stream_pools_only_after_trailers`. The server writes the trailer section in its own segment so the bytes are genuinely on the socket, and the follow-up request has to succeed *on the reused connection* — that way it pins both that a clean stream is pooled and that it is positioned correctly. It fails against the previous version of the flag.

One more gap turned up alongside it: the streaming path never checked `Connection: close`, so a cleanly framed response from a server about to hang up was still pooled. The async path checks it before `return_connection`; streaming does now too.

On CI: the twelve failures were all the same thing, and unrelated to the pooling logic. The new test initialized `.on_disconnect` after `.on_data` and `.on_error`, but `stream_request_params_v2` declares `on_disconnect` first, and designators have to appear in declaration order. GCC and MSVC enforce it; clang accepts any order as an extension, which is why every clang job was green and it reached CI at all. The branch is now rebased onto current `main` with that fixed, and #2819 reorders the struct so the natural spelling is the one that compiles.

wilkolbrzym-coder (commented):
## Review update – PR #2782

The trailer handling and `Connection: close` checks are correct now.

One thing to confirm in the final source: ensure the old `return_connection` block after `user_on_disconnect` is fully removed. The PR diff snippet shows both old and new blocks together, which would double-handle the socket if both existed.

Assuming that is a diff artifact:

- `consume_trailers` correctly consumes trailer lines to the final empty line before setting `response_complete`.
- `peer_will_close` prevents pooling a cleanly framed `Connection: close` response.
- Settling the socket before `user_on_disconnect` is the right ordering.
- The split-trailer test is strong and reproduces the exact failure mode.

LGTM once CI is green and #2819 ordering is coordinated.

stephenberry:
Confirmed as a diff artifact — there is one block, not two. `14201fc8` moved the existing block above `user_on_disconnect` rather than adding a second one, which is why a diff spanning both commits can read as if the old one survived.

The whole wrapper as it stands on the branch:

```cpp
auto internal_on_disconnect = [this, user_on_disconnect = std::move(on_disconnect), connection, url,
                               use_https]() {
   connection->is_connected = false;
   // Settle the socket before telling the caller the stream ended, matching the
   // non-streaming paths, which return the connection and only then run the
   // completion handler. [...]
   if (connection->socket) {
      // [...]
      if (connection->response_complete.load(std::memory_order_relaxed)) {
         connection_pool->return_connection(url.host, url.port, use_https, std::move(*connection->socket));
      }
      else {
         detail::close_socket(*connection->socket, connection_pool->graceful_ssl_shutdown());
      }
   }
   // Returning the socket leaves a null shared_ptr in the variant; every accessor
   // guards on it, so the disconnect() that http_stream_connection runs from its
   // destructor is a no-op rather than a dereference of a moved-from socket.
   if (user_on_disconnect) {
      user_on_disconnect();
   }
};
```

`grep` over the file finds exactly one `return_connection` in the streaming path, at the line above; the other three are the sync and async non-streaming paths.

Worth recording why that reordering is in this PR at all, since it arrived after your first review: it was not a tidy-up. The old order raced, and `chunked_stream_pools_only_after_trailers` hung for the full 300s ctest timeout on MSVC Release — the follow-up request went out while the pool was still empty, opened a connection the test server was no longer accepting, and both sides waited. Delaying the pooling past `on_disconnect` reproduced it locally on the first attempt. The test no longer depends on either side winning: the server polls for the follow-up request against a deadline instead of blocking on a read, and the client waits on the future with a deadline instead of `get()`. Forcing the socket never to pool now fails in about twelve seconds naming the cause, rather than hanging.

CI is green, including the MSVC Release job that caught this — 61 passing, with only a third-party packit job still pending.

On #2819: it reorders the callbacks in `stream_request_params_v2` so the natural spelling compiles. Both PRs are green independently. This one merges first, then #2819 rebases and flips the designators in the two `stream_request_v2` call sites here.

PR #2787, opened 2026-08-11
Replace http_headers matching_iterator with filter views

annihilatorq:
Recently, while looking through the `glz::http_headers` class, I suddenly realized that we could replace the entire `glz::http_headers::matching_iterator` code with a regular C++20 `std::views::filter`

No public `glz::http_headers` API changes.

## Pros
- Ranges returned by the `.fields()` method will be updated from `forward_range` to `bidirectional_range`, providing users with greater flexibility
- Less code, less responsibility, better maintainability

## Cons
None

### DIFF
--- a/include/glaze/net/http_headers.hpp
+++ b/include/glaze/net/http_headers.hpp
@@ -58,6 +58,102 @@ namespace glz
       using const_iterator = std::vector<value_type>::const_iterator;
       using size_type = std::vector<value_type>::size_type;
 
+     private:
+      template <bool IsConst>
+      class matching_iterator
+      {
+         using owner_type = std::conditional_t<IsConst, const http_headers*, http_headers*>;
+
+        public:
+         using iterator_concept = std::bidirectional_iterator_tag;
+         using iterator_category = std::bidirectional_iterator_tag;
+         using difference_type = ptrdiff_t;
+         using value_type = http_headers::value_type;
+         using reference = std::conditional_t<IsConst, const value_type&, value_type&>;
+         using pointer = std::conditional_t<IsConst, const value_type*, value_type*>;
+
+         matching_iterator() = default;
+         matching_iterator(const matching_iterator&) = default;
+         matching_iterator(matching_iterator&&) = default;
+         matching_iterator& operator=(const matching_iterator&) = default;
+         matching_iterator& operator=(matching_iterator&&) = default;
+         ~matching_iterator() = default;
+
+         matching_iterator(owner_type owner, size_t start_index, std::string_view key)
+            : owner_(owner), index_(start_index), key_(key)
+         {
+            seek_forward();
+         }
+
+         [[nodiscard]] reference operator*() const noexcept { return owner_->headers_[index_]; }
+         [[nodiscard]] pointer operator->() const noexcept { return std::addressof(owner_->headers_[index_]); }
+
+         matching_iterator& operator++() noexcept
+         {
+            if (index_ < owner_->headers_.size()) {
+               ++index_;
+               seek_forward();
+            }
+            return *this;
+         }
+
+         matching_iterator operator++(int) noexcept
+         {
+            auto copy = *this;
+            ++(*this);
+            return copy;
+         }
+
+         matching_iterator& operator--() noexcept
+         {
+            seek_backward();
+            return *this;
+         }
+
+         matching_iterator operator--(int) noexcept
+         {
+            auto copy = *this;
+            --(*this);
+            return copy;
+         }
+
+         [[nodiscard]] friend bool operator==(const matching_iterator& left, const matching_iterator& right) noexcept
+         {
+            return left.owner_ == right.owner_ && left.index_ == right.index_ && glz::striequal(left.key_, right.key_);
+         }
+
+        private:
+         void seek_forward() noexcept
+         {
+            while (index_ < owner_->headers_.size()) {
+               const auto& current = owner_->headers_[index_].name;
+               if (glz::striequal(current, key_)) {
+                  return;
+               }
+               ++index_;
+            }
+         }
+
+         // Precondition: a preceding match exists, as required of a bidirectional iterator
+         void seek_backward() noexcept
+         {
+            while (index_ > 0) {
+               --index_;
+               if (glz::striequal(owner_->headers_[index_].name, key_)) {
+                  return;
+               }
+            }
+         }
+
+         owner_type owner_{};
+         size_t index_{0};
+         std::string_view key_;
+      };
+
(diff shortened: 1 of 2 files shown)


stephenberry:
I tend to avoid std::ranges because they are heavy to compile and have some performance issues, but for this API they might be great. What do you think?

stephenberry:
*These thoughts are AI generated.*

Thanks for this. The bidirectional upgrade is a genuinely nice win, the diff is clean, and CI is fully green.

Before merging I'd like to float an alternative, because moving to `std::views::filter` has a few consequences worth weighing.

**1. The returned range is no longer const-iterable.** `filter_view::begin()` is non-const, so this compiles on `main` and fails on this branch:

```cpp
const auto r = headers.fields("Set-Cookie");
for (auto&& f : r) { ... }   // error: no matching begin()
```

Same for `values(name)`. Worth noting that `names()` and the no-arg `values()` stay const-iterable, since they are `transform` over `ref_view`, so `values(name)` and `values()` would end up behaving differently from each other.

**2. `range_iterator` and `const_range_iterator` are public typedefs** that shipped in v8.0.0, so removing them is a source break, however unlikely anyone is naming them.

**3. Mutation through the non-const `fields()` overload becomes UB in one case.** Per [range.filter.iterator], modifying an element so that it no longer satisfies the predicate is undefined. So normalizing header case through the mutable view:

```cpp
for (auto& f : headers.fields("Set-Cookie")) f.name = "set-cookie";
```

was well-defined with the index-based iterator and is UB with `filter_view`. Changing `.value` is unaffected.

## An alternative that keeps the win

Give the existing `matching_iterator` an `operator--`, roughly 20 lines:

```diff
-         using iterator_concept = std::forward_iterator_tag;
-         using iterator_category = std::forward_iterator_tag;
+         using iterator_concept = std::bidirectional_iterator_tag;
+         using iterator_category = std::bidirectional_iterator_tag;
```

```diff
          matching_iterator operator++(int) noexcept
          {
             auto copy = *this;
             ++(*this);
             return copy;
          }
+
+         matching_iterator& operator--() noexcept
+         {
+            seek_backward();
+            return *this;
+         }
+
+         matching_iterator operator--(int) noexcept
+         {
+            auto copy = *this;
+            --(*this);
+            return copy;
+         }
```

```diff
        private:
+         // Precondition: a preceding match exists, as required of a bidirectional iterator
+         void seek_backward() noexcept
+         {
+            while (index_ > 0) {
+               --index_;
+               if (glz::striequal(owner_->headers_[index_].name, key_)) {
+                  return;
+               }
+            }
+         }
+
          void seek_forward() noexcept
```

plus the same `forward_range` to `bidirectional_range` `static_assert` bump this PR already makes.

That yields `bidirectional_range`, so `fields(name) | std::views::reverse` works, which is the headline benefit here, while keeping const-iterability, the public typedefs, and well-defined mutation through the mutable overload. The cost is that the ~80 lines you wanted to delete stay.

I compiled this at `-Wall -Wextra` and ran it under ASan/UBSan with hardened libc++. Reverse iteration, `--` from `end()` landing on the last match, `++`/`--` round-tripping, and the empty and single-match cases all behave.

Curious what you think. If the const-iterability loss seems acceptable, and there is a reasonable case for that since it is idiomatic for std ranges and views are not really meant to live in const objects, then this PR is fine as-is and the typedef removal just wants a line in the release notes.

annihilatorq:
TL;DR - I agree. Point 3 has convinced me the most - I'll replace `views::filter` with a bidirectional `matching_iterator`. Please give me a `🚀` if you'd like me to commit those changes.

> I tend to avoid std::ranges because they are heavy to compile and have some performance issues, but for this API they might be great. What do you think?

I may have misunderstood the context a bit, but `glz::http_headers` is already built entirely around the idea of the ranges API, and this PR just adds one more adapter to the existing ones. Or do you mean that you have thoughts on redesigning the entire `glz::http_headers` API at this stage? In case you're considering a broader API redesign, here are my thoughts on ranges in general, since `views::filter` (this PR) will have minimal impact on the existing user experience.

I remember your policy regarding compilation speed. I won't claim the compile-time cost is negligible, because that needs to be benchmarked, but I'd consider it an acceptable trade-off for the quality and usability of the ranges API. Also, given that Glaze requires C++23, there is a high probability that most serious projects using Glaze are already using `std::ranges`, though this is of course not certain and depends on the internal policies of different projects.

Another point worth mentioning is that with the range adaptors, the `glz::http_headers` API will remain "modern C++" for a pretty long time and will automatically incorporate all updates and overloads from future language standards, expanding the API with zero effort on our part.


> 1. The returned range is no longer const-iterable
> 2. range_iterator and const_range_iterator are public typedefs
> 3. Mutation through the non-const fields() overload becomes UB in one case
>
> Curious what you think. If the const-iterability loss seems acceptable, and there is a reasonable case for that since it is idiomatic for std ranges and views are not really meant to live in const objects, then this PR is fine as-is and the typedef removal just wants a line in the release notes.

1. I wouldn't consider this the main issue, since ranges aren't really intended for `const` use; some adaptors have to cache state in begin(), which is why const-iteration was never part of the views contract. As for the API, I think this comes down to personal preference. I'd put it this way:
   - In favor of `non-const-iterable` is that it's idiomatic and aligns with standard C++ practices, it follows what users already expect from standard views.
   - In favor of `const-iterable` is user convenience - although `const` on a view object is shallow and gives no real protection, it allows users who habitually write `const auto` to do so without a confusing compile error.
2. This is not as serious as 1 and 3, but that's a fair point, I missed it in the Cons section.
3. In my opinion, this is the most serious problem and the reason why we should still keep `matching_iterator` and follow the approach you suggested. I wasn't aware of this UB, and after reading a bit about the topic, it turns out that's indeed the case.

stephenberry:
> I may have misunderstood the context a bit, but glz::http_headers is already built entirely around the idea of the ranges API, and this PR just adds one more adapter to the existing ones. Or do you mean that you have thoughts on redesigning the entire glz::http_headers API at this stage? In case you're considering a broader API redesign, here are my thoughts on ranges in general, since views::filter (this PR) will have minimal impact on the existing user experience.

I think the ranges API is good, and valuable for this higher level interface. Core code can avoid it. I just don't want to move APIs to ranges without considering whether the higher level abstraction is worth the cost. But, I like the design of `http_headers` and think it's good.

stephenberry:
@annihilatorq, is this ready to merge? If we get this merged I think I can cut a new release.

annihilatorq:
> @annihilatorq, is this ready to merge? If we get this merged I think I can cut a new release.

Yep, everything is ready

PR #2789, opened 2026-08-11
reject bare CR/LF and non-token header fields in try_parse_request

uwezkhan:
try_parse_request splits header field lines on CRLF and stores each name and value as-is, so a bare LF or CR embedded in a field line survives into the stored value. A request whose value is `1\nContent-Length: 40` reaches glaze as a single X-Test field, the content-length lookup finds nothing, and the trailing bytes get reparsed as a second request. A front-end proxy that treats a bare LF as a line break instead reads a separate Content-Length and frames those bytes as the body, so the two sides disagree on where the message ends (CL.0 smuggling).

The fix runs each field name through glz::valid_header_name and each value through glz::valid_header_value inside the parse loop, answering 400 and closing on a violation. Those predicates already back the WebSocket handshake path and were written as the shared source of field validity, so this brings the server parser in line with the rest of the net stack instead of adding a separate check. Before, malformed field bytes were accepted and mis-framed; after, they are refused before body framing, while well-formed headers with spaces, colons, and obs-text in the value are unchanged.

### DIFF
--- a/include/glaze/net/http.hpp
+++ b/include/glaze/net/http.hpp
@@ -263,6 +263,20 @@ namespace glz
       return striequal(name, "content-length") || striequal(name, "transfer-encoding");
    }
 
+   // RFC 9110, Section 5.6.2:
+   // token = 1*tchar
+   // tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*"
+   //       / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+   //       / DIGIT / ALPHA
+   // The single tchar predicate for the net stack: valid_header_name and the
+   // request-line method check both build on it.
+   [[nodiscard]] inline bool is_tchar(char ch) noexcept
+   {
+      return (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') || (ch >= '0' && ch <= '9') || ch == '!' ||
+             ch == '#' || ch == '$' || ch == '%' || ch == '&' || ch == '\'' || ch == '*' || ch == '+' || ch == '-' ||
+             ch == '.' || ch == '^' || ch == '_' || ch == '`' || ch == '|' || ch == '~';
+   }
+
    // RFC 7230 3.2.6: a header field-name is one or more token characters (tchar).
    // A name that is empty or carries any other byte (CR/LF, space, colon, a
    // control char, ...) cannot be written as a well-formed field. This is the
@@ -272,11 +286,8 @@ namespace glz
       if (name.empty()) {
          return false;
       }
-      for (const unsigned char c : name) {
-         const bool tchar = (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '!' ||
-                            c == '#' || c == '$' || c == '%' || c == '&' || c == '\'' || c == '*' || c == '+' ||
-                            c == '-' || c == '.' || c == '^' || c == '_' || c == '`' || c == '|' || c == '~';
-         if (!tchar) {
+      for (const char c : name) {
+         if (!is_tchar(c)) {
             return false;
          }
       }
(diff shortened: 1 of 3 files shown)


stephenberry:
Thanks for this. I applied it on top of `main` and verified it locally (Debug build with sanitizers).

**The desync is real and reproducible.** Reverting just the `http_server.hpp` hunk while keeping the new tests fails all three, and not only on the missing 400 — `smuggled_hits` goes non-zero, so the smuggled `GET /smuggled` genuinely reaches a route handler:

```
FAILED "bare LF inside a header value is rejected, not desynced"  Expected 400 for bare LF in header value
FAILED "bare LF inside a header value is rejected, not desynced"  Smuggled request reached a route handler
FAILED "bare CR inside a header value is rejected, not desynced"  ...
FAILED "bare LF inside a header name is rejected, not desynced"   ...
tests: 7 (4 passed, 3 failed)
```

With the hunk in place `http_header_strictness_test` is 7/7, and there is no over-rejection elsewhere: `http_smuggling`, `http_headers`, `http_response_splitting`, `http_content_length`, `http_router`, `cors`, `http_chunked`, `http_server_post`, `http_client`, `http_server_api`, `websocket_close`, `websocket_client`, and `rest` are all green.

The approach is the right one. Reusing `glz::valid_header_name` / `glz::valid_header_value` rather than adding a parallel check is what those predicates exist for, and running the value check after the OWS trim is safe since SP/HTAB are legal field-value bytes anyway. `obs-text` stays accepted and `_` is a tchar, so Latin-1 values and `X_Forwarded_For`-style names are unaffected. With the request-line already strictly validated, this closes the last bare-CR/LF ingress in the request head.

### One change before merge

Please drop the now-redundant field-name whitespace guard a few lines above:

```cpp
if (!name_sv.empty() && (name_sv.back() == ' ' || name_sv.back() == '\t')) {
   result.status = parse_status::error;
   send_error_response_with_close(conn, 400, "Bad Request");
   return result;
}
```

`valid_header_name` already rejects SP and HTAB (neither is a tchar), so this block is fully subsumed and both paths produce the same 400-and-close. It is an extra branch on the parse hot path, and leaving it in means two adjacent comments explaining overlapping rules, which is exactly the kind of thing that rots. Fold the smuggling rationale from its comment (the obfuscated `Transfer-Encoding : chunked` storing under a mangled key) into your new comment so the reasoning is preserved in one place. The existing "space before colon is rejected, not desynced" test should keep passing untouched — that is the check that the removal is safe.

### Optional, welcome in this PR

`glz::detail::is_tchar` in `http_server.hpp` duplicates the tchar set that `valid_header_name` documents itself as "the single source of truth for field-name validity across the net stack." Since `http_server.hpp` includes `http.hpp` and not the other way around, consolidating means moving the per-character predicate into `http.hpp` next to `valid_header_name`, having `valid_header_name` call it, and pointing `parse_request_line`'s method check at the same one. Happy to take that separately if you would rather keep this PR to the fix.

### Follow-ups, not for this PR

- The client's response header parsing (three sites in `http_client.hpp`) uses the same `find(':')` + trim pattern with no validation, so a bare LF in a response field value has the same split-disagreement property in the other direction. The same two predicates would close it.
- A field line with no colon at all is silently dropped in `try_parse_request` (the `if (colon_pos != npos)` has no `else`). RFC 9112 §5 makes such a line malformed; it should be a 400 rather than ignored.

One test nit you can take or leave: `response.find("400")` could in principle match body bytes, and `starts_with("HTTP/1.1 400")` would be tighter. The surrounding tests use the same idiom, so it is consistent as written.

uwezkhan:
Done. Dropped the whitespace guard and folded the obfuscated "Transfer-Encoding : chunked" rationale into the validation comment. Before, the same input hit two adjacent checks with separate explanations; after, the token check is the single rejection point and the hot path has one less branch. The existing "space before colon" test passes untouched, which confirms the removal changed no behavior.

Took the is_tchar consolidation here too since it is the same cleanup: the predicate now lives in http.hpp next to valid_header_name, valid_header_name loops over it, and the request-line method check picks it up from there, so http_server.hpp no longer carries its own copy of the tchar set.

Reran locally: http_header_strictness is 7/7, and http_smuggling, http_headers, http_response_splitting, http_content_length, http_router, cors, http_chunked, http_server_post, http_server_api, the request-line parser tests, websocket_close, and websocket_client are all green.

Agreed on both follow-ups; the three client-side response header sites and the colon-less field line are the same pattern and worth a separate PR. Left the find("400") idiom as is to stay consistent with the rest of the file.

PR #2798, opened 2026-08-13
Include unordered_map meta header in json/schema.hpp

### LINKED ISSUE #2764: glaze/json/schema.hpp doesn't include glaze/api/std/unordered_map.hpp
The `glaze/json/schema.hpp` header doesn't include the `glaze/api/std/unordered_map.hpp` header, leading to differences between compilers. It does include `map`, so I suppose it was forgotten.

https://github.com/stephenberry/glaze/blob/dda11044fb0cf4ec6b3bced58fcd8136a987e06b/include/glaze/json/schema.hpp#L10

alisonpetersonhq:
Fixes #2764

`glaze/json/schema.hpp` already pulls in `glaze/api/std/map.hpp` for consistent type naming across compilers, but was missing the matching header for `std::unordered_map`.

Adds the include and a JSON Schema regression test mirroring the existing `std::map` case.

### DIFF
--- a/include/glaze/json/schema.hpp
+++ b/include/glaze/json/schema.hpp
@@ -8,6 +8,7 @@
 #include "glaze/api/std/functional.hpp"
 #include "glaze/api/std/list.hpp"
 #include "glaze/api/std/map.hpp"
+#include "glaze/api/std/unordered_map.hpp"
 #include "glaze/api/std/optional.hpp"
 #include "glaze/api/std/shared_ptr.hpp"
 #include "glaze/api/std/string.hpp"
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for opening this PR. The reason why I haven't immediately added this, is that I want to ensure that users in general aren't paying higher compilation costs for features they don't need. Including headers for when a name might be needed by some users makes everyone who is using json/shema.hpp pay needlessly.

Now, there are two viable solutions. I have a PR I just added that groups a bunch of the naming headers together for the standard library, so you only pay for the specific names when you need them. But, I'm also considered a more efficient and general solution that uses C++20 concepts to avoid needing to bring in any headers and works generically for third party containers that match the same API.

PR #2801, opened 2026-08-14
feat(json): support std::chrono::utc_clock time points (#2657)

### LINKED ISSUE #2657: Support std::chrono::utc_time / utc_clock
Raised in #2640 (the reporter noted `std::chrono::utc_time` is missing and offered to implement it) and confirmed during review of the per-field chrono wrappers.

### Current state

There is no `utc_clock` support anywhere in Glaze. The chrono serializers and the new per-field wrappers gate on `is_system_time_point` (i.e. `clock == system_clock`), so a `std::chrono::utc_time` member hard-errors via `static_assert`. Because `date_format` already describes `system_clock` times as "UTC wall-clock", a user could reasonably assume `utc_clock` is supported when it is not.

### Proposed

- Serialize `std::chrono::utc_time<Duration>` (default ISO 8601 string, matching the `system_clock` representation) and parse it back.
- Allow `glz::date_format` / `glz::epoch_count` on `utc_time` members.
- Mind the semantic difference: `utc_clock` accounts for leap seconds whereas `system_clock` (Unix time) does not, so `23:59:60` is representable for `utc_time`. Decide and document how leap seconds are handled on read/write (the current ISO path rejects `:60`).

### Notes

This is the primary feature follow-up from #2640; the wrappers PR scopes it out explicitly. Until it lands, the docs should state that `utc_clock` time points are out of scope and fail to compile.

alisonpetersonhq:
## Summary

Adds JSON serialization support for `std::chrono::utc_clock` time points (`utc_time`), addressing #2657.

UTC timestamps use the same ISO 8601 wire format as `system_clock` time points, converting via `std::chrono::clock_cast` on read and write.

Fixes #2657

## Test plan

- [x] `tests/chrono_test/chrono_test.cpp`: `utc_time_json_roundtrip` (C++20 `__cpp_lib_chrono`)

### DIFF
--- a/include/glaze/core/chrono.hpp
+++ b/include/glaze/core/chrono.hpp
@@ -35,6 +35,13 @@ namespace glz
    concept is_system_time_point =
       is_time_point<T> && std::is_same_v<typename std::remove_cvref_t<T>::clock, std::chrono::system_clock>;
 
+#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201907L
+   // Concept for utc_clock time_points (serialize as ISO 8601 string via clock_cast)
+   template <class T>
+   concept is_utc_time_point =
+      is_time_point<T> && std::is_same_v<typename std::remove_cvref_t<T>::clock, std::chrono::utc_clock>;
+#endif
+
    // Concept for steady_clock time_points (serialize as numeric count)
    template <class T>
    concept is_steady_time_point =
--- a/include/glaze/json/read.hpp
+++ b/include/glaze/json/read.hpp
@@ -5209,6 +5209,29 @@ namespace glz
       }
    };
 
+#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201907L
+   // utc_clock::time_point: parse from ISO 8601 string via clock_cast from sys_time.
+   template <is_utc_time_point T>
+      requires(not custom_read<T>)
+   struct from<JSON, T>
+   {
+      template <auto Opts, class It0, class It1>
+      static void op(auto&& value, is_context auto&& ctx, It0&& it, It1 end) noexcept
+      {
+         std::string_view str;
+         parse_transient_string_view<Opts>(str, ctx, it, end);
+         if (bool(ctx.error)) [[unlikely]]
+            return;
+
+         std::chrono::system_clock::time_point sys{};
+         chrono_detail::parse_iso8601(str, sys, ctx.error);
+         if (bool(ctx.error)) [[unlikely]]
+            return;
+         value = std::chrono::clock_cast<std::chrono::utc_clock>(sys);
+      }
+   };
+#endif
+
    // year_month_day: parse from "YYYY-MM-DD" JSON string
    template <is_year_month_day T>
       requires(not custom_read<T>)
--- a/include/glaze/json/write.hpp
+++ b/include/glaze/json/write.hpp
@@ -2594,6 +2594,25 @@ namespace glz
       }
    };
 
+#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201907L
+   // utc_clock::time_point: serialize as ISO 8601 string (UTC), same wire form as sys_time.
+   template <is_utc_time_point T>
+      requires(not custom_write<T>)
+   struct to<JSON, T>
+   {
+      template <auto Opts, class B>
+      static void op(auto&& value, is_context auto&& ctx, B&& b, auto& ix) noexcept
+      {
+         using Period = typename std::remove_cvref_t<T>::duration::period;
+         if (!ensure_space(ctx, b, ix + chrono_detail::iso_time_point_max_size<Period>)) [[unlikely]] {
+            return;
+         }
+         const auto sys = std::chrono::clock_cast<std::chrono::system_clock>(value);
+         chrono_detail::write_iso_time_point<not check_unquoted(Opts)>(sys, ctx, b, ix);
+      }
+   };
+#endif
+
    // year_month_day: serialize as "YYYY-MM-DD" JSON string
    template <is_year_month_day T>
       requires(not custom_write<T>)
--- a/tests/chrono_test/chrono_test.cpp
+++ b/tests/chrono_test/chrono_test.cpp
@@ -1748,4 +1748,18 @@ suite date_format_tests = [] {
    };
 };
 
+#if defined(__cpp_lib_chrono) && __cpp_lib_chrono >= 201907L
+suite utc_clock_json_tests = [] {
+   "utc_time_json_roundtrip"_test = [] {
+      const std::chrono::utc_time<std::chrono::seconds> t{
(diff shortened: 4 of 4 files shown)


stephenberry (changes requested):
Thanks for taking this on — `utc_clock` support is genuinely wanted (#2657). The concept and the `__cpp_lib_chrono >= 201907L` gate are the right shape: I confirmed Apple libc++ 21 reports `__cpp_lib_chrono = 201611` and has no `std::chrono::clock_cast`, so the gate correctly closes there.

However, the two serializers have blocking defects, and nothing has compiled them yet.

## CI has not validated this PR

The only green checks are the five Packit `rpm-build:fedora-rawhide-*` jobs. `util/glaze.spec` builds with `glaze_DEVELOPER_MODE=OFF`, so no tests are built and `%ctest` runs zero cases; a header-only install never instantiates the new templates. There is no GitHub Actions run on the head commit. The green ticks say nothing about the new code.

## 1. Buffer overflow in the writer (blocking)

In `include/glaze/json/write.hpp`, space is reserved from `T::duration::period`, but the bytes are written from the `clock_cast` result, whose duration is `std::common_type_t<T::duration, system_clock::duration>` — always as fine or finer:

```cpp
using Period = typename std::remove_cvref_t<T>::duration::period;   // seconds
if (!ensure_space(ctx, b, ix + chrono_detail::iso_time_point_max_size<Period>)) ...
const auto sys = std::chrono::clock_cast<std::chrono::system_clock>(value);  // microseconds/nanoseconds
chrono_detail::write_iso_time_point<...>(sys, ctx, b, ix);          // writes at the *cast's* precision
```

For `utc_time<seconds>` — the type used in the test in this PR — that reserves 22 bytes and writes 29 (libc++, µs) or 31 (libstdc++, ns). Driving Glaze's own writer with the two types under ASan:

```
reserved (from T's period)     = 22
written  (from cast's period)  = 29
==ERROR: AddressSanitizer: stack-buffer-overflow
    #0 glz::chrono_detail::write_digits<6ul, ...>
    #1 glz::chrono_detail::write_iso_time_point<true, time_point<system_clock, duration<long long, ratio<1, 1000000>>>, ...>
```

For a bounded buffer (`std::array`, `std::span`) `ensure_space` is the only guard, so this is an out-of-bounds write rather than merely an under-reservation.

The same root cause corrupts the wire format: `utc_time<seconds>` would emit `...T12:00:00.000000Z` where `sys_time<seconds>` emits `...T12:00:00Z`, and `utc_days` would lose the date-only `"YYYY-MM-DD"` form that `sys_days` gets.

## 2. The reader does not compile for coarse precisions (blocking)

In `include/glaze/json/read.hpp`:

```cpp
std::chrono::system_clock::time_point sys{};
chrono_detail::parse_iso8601(str, sys, ctx.error);
value = std::chrono::clock_cast<std::chrono::utc_clock>(sys);
```

This assigns a `utc_time<system_clock::duration>` to `utc_time<Duration>`. `time_point`'s converting constructor requires `is_convertible_v<Duration2, Duration>`, which is false whenever `Duration` is coarser than the system duration. The exact analogue:

```
common_type period num=1 den=1000000
implicitly convertible to seconds?                                 0
sys_time<seconds> assignable from system_clock::time_point?        0
```

`utc_time<seconds>` is precisely that case, so `utc_time_json_roundtrip` as written cannot build. The intermediate also drops the bare `"YYYY-MM-DD"` read path that `sys_days` accepts.

## 3. Suggested shape: one generic delegation, not per-format code

Both defects disappear if the conversion is expressed once as a delegation to the existing `system_clock` serializers, placed in `core/chrono.hpp` beside the `is_count_time_point` generic rather than hand-rolled per format:

```cpp
using Duration = typename std::remove_cvref_t<T>::duration;
const std::chrono::sys_time<Duration> sys =
   std::chrono::time_point_cast<Duration>(std::chrono::clock_cast<std::chrono::system_clock>(value));
to<Format, std::chrono::sys_time<Duration>>::template op<Opts>(sys, ctx, args...);
```

Constrained with `requires write_supported<std::chrono::sys_time<Duration>, Format>` (and the `read_supported` mirror), one small generic covers JSON, YAML, TOML, CBOR and MsgPack at once, inherits the `sys_days` handling and the exact size accounting, and keeps `write_supported` / `read_supported` reporting `false` for BEVE and CSV instead of hard-erroring inside the op.

## 4. `noexcept` over a potentially throwing call

`clock_cast` between `sys_time` and `utc_time` consults leap-second data through the tzdb; on libstdc++ and MSVC that path can throw `std::runtime_error` when the tz database is unavailable. Inside a `noexcept` op that is `std::terminate`, in a header-only library that also supports `-fno-exceptions`. This needs a deliberate decision — at minimum a guarded call that maps failure to an `error_code`.

## 5. Leap-second semantics are undecided and undocumented

#2657 asks explicitly for a decision here. A `utc_time` at `23:59:60` casts to the following second in sys time and does not come back, so the advertised round-trip does not hold for the one case `utc_clock` exists to represent. Defining the wire form as Unix time is a reasonable call (interop favors it), but it should be stated in the docs and pinned by a test rather than left implicit.

## 6. Scope left open

- `glz::date_format` and `glz::epoch_count` still reject `utc_time` (`json/chrono_format.hpp` static asserts). That is the second half of #2657.
- `docs/chrono.md` is untouched: the format-support matrix needs a `utc_clock::time_point` row, the leap-second note needs updating, and the "out of scope / fails to compile" framing needs removing.

## 7. Test coverage

`utc_clock::now()` is non-deterministic, and a symmetric round-trip cannot detect a wrong offset — encode the same instant incorrectly on both sides and the test still passes. Once the implementation is fixed, the test should:

- use fixed instants and assert the exact emitted string, including that it matches the `sys_time` form for the same instant;
- cover `seconds`, `milliseconds`, `nanoseconds` and `utc_days`;
- write into a `std::array` buffer, which would have caught the overflow above;
- exercise the parse-error path on malformed input;
- pin the chosen leap-second behavior.

Happy to see this land once the writer and reader are reworked along the lines of §3 and the leap-second semantics are settled.

wilkolbrzym-coder (changes requested):
One addition to Stephen’s review.

His points cover the writer overflow, reader compile failure, generic delegation, noexcept, leap-second semantics, scope, and tests.

Missing from his list:

1. The test calls `std::chrono::utc_clock::now()`, which can throw if tz data is unavailable. Fixed-instant tests would avoid nondeterminism and that failure.

2. No CI or test covers `-fno-exceptions`. `clock_cast` may throw while the new ops are `noexcept`; this currently ends in `std::terminate`.

3. `check_unquoted(Opts)` changes the output, but the reservation `iso_time_point_max_size<Period>` does not account for omitted quotes. For bounded buffers this is another potential under-reservation.

4. Only JSON is implemented and tested. If the fix should be generic delegation, at least one other format such as YAML or CBOR should be covered now.

These do not change the verdict; they reinforce the existing request for changes.

cc @stephenberry @alisonpetersonhq

PR #2803, opened 2026-08-14
bound dynamic eigen matrix reads to the matrix storage

uwezkhan:
The dynamic Eigen matrix readers for JSON and BEVE take the [rows,cols] header off the wire and never check it against the buffer or the matrix's own storage. The JSON reader parses the extents array and then reads `*it` to look for the ',' that precedes the data without confirming it is still inside the buffer, so a non-null-terminated input that ends right after the header (for example the six bytes `[[2,3]` read with null_terminated off) reads one byte past the end. Both readers also size the data span from `extents[0] * extents[1]` instead of the count the matrix actually holds, which is equal only for a plain dynamic matrix and diverges for a fixed-max-storage one.

Fix checks the parse error before resizing, guards the post-header dereference against the end on non-null-terminated reads, and sizes the span from `value.size()`, the shape `from<CBOR, T>` already uses one specialization over. Well-formed input behaves the same, since `value.size()` equals the wire product after a successful resize; the tradeoff is a couple of branches on the read path that were previously skipped. Confirmed under AddressSanitizer (heap-buffer-overflow at eigen.hpp:330 before, clean after) with a regression test added to eigen_test.

### DIFF
--- a/include/glaze/ext/eigen.hpp
+++ b/include/glaze/ext/eigen.hpp
@@ -24,6 +24,44 @@ static_assert(false, "Eigen must be included to use glaze/ext/eigen.hpp");
 
 namespace glz
 {
+   namespace eigen_detail
+   {
+      // Validates the [rows, cols] header of a dynamic matrix against what the matrix can
+      // actually store, before it is handed to resize().
+      //
+      // A negative extent is not a dimension, and Eigen's resize() has no defined behavior for
+      // one. Worse, two negatives multiply back to a positive size(), so a header like [-2, -3]
+      // would otherwise produce a 6-element data span over a matrix that holds nothing.
+      //
+      // A dynamic matrix with a fixed maximum (Matrix<T, -1, -1, 0, MaxRows, MaxCols>) carries
+      // MaxRows * MaxCols scalars of inline storage, and its resize() only records the new
+      // extents in that fixed buffer -- it cannot grow. size() therefore reports the wire
+      // product rather than the capacity, so nothing downstream would catch a header that
+      // exceeds the buffer. Reject it here, mirroring the fixed-extent check in from<CBOR, T>.
+      template <class T>
+      [[nodiscard]] GLZ_ALWAYS_INLINE bool invalid_extents(const std::array<Eigen::Index, 2>& extents,
+                                                           is_context auto& ctx) noexcept
+      {
+         if (extents[0] < 0 || extents[1] < 0) [[unlikely]] {
+            ctx.error = error_code::syntax_error;
+            return true;
+         }
+         if constexpr (T::MaxRowsAtCompileTime >= 0) {
+            if (extents[0] > T::MaxRowsAtCompileTime) [[unlikely]] {
+               ctx.error = error_code::exceeded_static_array_size;
+               return true;
+            }
+         }
+         if constexpr (T::MaxColsAtCompileTime >= 0) {
+            if (extents[1] > T::MaxColsAtCompileTime) [[unlikely]] {
+               ctx.error = error_code::exceeded_static_array_size;
+               return true;
+            }
+         }
+         return false;
+      }
+   }
+
    template <matrix_t T>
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this — I reproduced all three issues under ASan before reviewing, and they're real:

- The `*it` probe for the `,` overflows exactly as described (`heap-buffer-overflow`, `eigen.hpp:330`) on `[[2,3]` with `null_terminated` off.
- The BEVE truncated header genuinely resizes to stack garbage; my repro threw `std::bad_alloc` out of `read_beve`.
- Your third check also fixes a **second** out-of-bounds read that the description doesn't mention: `match<']'>(ctx, it)` is unguarded in the same way, so `[[1,1],[7]` (well-formed data, buffer ends before the closing bracket) reads past the end at `eigen.hpp:342`. Worth calling out, since it had no test.

I've pushed a commit to the branch rather than leaving these as review notes — details below.

## The fixed-max-storage rationale isn't right

> sizes the data span from `extents[0] * extents[1]` instead of the count the matrix actually holds, which is equal only for a plain dynamic matrix and diverges for a fixed-max-storage one

`value.size()` doesn't diverge there. For `Matrix<T, -1, -1, 0, MaxRows, MaxCols>`, `PlainObjectBase::resize` only records `m_rows`/`m_cols` into the fixed `DenseStorage` — it can't grow — so `size()` returns the wire product, exactly what it replaced. The hunk is a no-op.

The underlying overflow is real, though. `Matrix<double,-1,-1,0,2,2>` fed `[[4,4],[1..16]]`, built with `-DEIGEN_NO_DEBUG` (i.e. any release build, where the `eigen_assert` that catches the oversized resize is compiled out):

```
before: stack-buffer-overflow, WRITE of size 8, read.hpp:2352
after:  stack-buffer-overflow, WRITE of size 8, read.hpp:2393
```

Identical on both sides. So rather than just correcting the wording, I backed the claim: `eigen_detail::invalid_extents<T>` now validates the header against the matrix's storage *before* the resize, shared by the JSON and BEVE readers, mirroring the fixed-extent check `from<CBOR, T>` already performs. The `if constexpr` branches disappear entirely for plain `MatrixXd`, so the common path is unchanged.

That helper also rejects negative extents, which turned out to be **required** rather than a bonus: two negatives multiply back to a positive `size()`, so `[[-2,-3]]` slips past a max-only check and still produces a 6-element span over a matrix holding nothing.

With the extents validated, sizing from `value.size()` is finally correct — but because of the check above it, not on its own. I reworded both comments accordingly.

## Tests

Added three, each verified to crash or throw against `main` in isolation:

- buffer ending after a well-formed data array, before `]` (the second OOB above)
- fixed-maximum matrix overflowed by an oversized header, over both JSON and BEVE
- negative extents

I deliberately avoided byte-patching a written BEVE buffer to forge the oversized header — that breaks on the s390x job (BEVE is little-endian on the wire and glaze byteswaps on big-endian hosts) and on armv7 (`Eigen::Index` is 4 bytes, shifting every offset). Writing a 4×4 matrix and reading it into the 2×2-bounded type produces the same header portably.

Ran the suite under ASan+UBSan with `-fno-sanitize-recover=all` in both an assertions-on build and an `EIGEN_NO_DEBUG -DNDEBUG` build. The release config is the one that matters here — `eigen_assert` was masking this case, so an assertions-on run alone proves nothing about it.

## Still open (follow-up, not this PR)

A header declaring huge extents drives an unbounded allocation before any data is read, and `std::bad_alloc` escapes a `glz::read_*` call that otherwise reports errors by return code. 54 bytes of BEVE with the extents patched to `0x0FFFFFFF`, or JSON `[[100000000,100000000],[1]]`, both attempt an enormous allocation. Glaze's own BEVE readers guard this with `if (uint64_t(end - it) < n)` (`beve/read.hpp:1433` and ~10 other sites) and `exceeds_capacity` (`core/read.hpp:27`); the eigen readers use neither. The declared element count can't exceed the bytes remaining, so it should be a cheap fix — happy to take it separately if you'd like to pick it up.

PR #2804, opened 2026-08-14
fix(net): avoid duplicate Host/Connection when caller supplies them (#2777)

### LINKED ISSUE #2777: http_client sends Host and Connection twice when the caller supplies them

alisonpetersonhq:
﻿## Summary
Avoids duplicate `Host` and `Connection` headers when the caller supplies them (#2777).

`build_http_request_bytes` previously always wrote default `Host` / `Connection: keep-alive`, then appended caller headers. Callers that set either field ended up with two on the wire.

## Change
- Detect caller `Host` / `Connection` (case-insensitive, ignoring CRLF-rejected fields)
- Emit writer defaults only when absent
- Caller values are written once in the header loop (body framing still owned by the writer)

## Test
Adds `build_http_request_bytes prefers caller Host and Connection` to `http_response_splitting_test`.

Closes #2777

### DIFF
--- a/include/glaze/net/http_client.hpp
+++ b/include/glaze/net/http_client.hpp
@@ -268,23 +268,43 @@ namespace glz
       {
          const bool is_default_port = (!use_https && url.port == 80) || (use_https && url.port == 443);
 
+         // Prefer a single Host / Connection on the wire. If the caller already
+         // supplied either, skip our defaults so we do not emit duplicates (#2777).
+         bool caller_has_host = false;
+         bool caller_has_connection = false;
+         for (const auto& [name, value] : headers) {
+            if (header_field_has_crlf(name, value)) [[unlikely]] {
+               continue;
+            }
+            if (glz::striequal(name, "host")) {
+               caller_has_host = true;
+            }
+            else if (glz::striequal(name, "connection")) {
+               caller_has_connection = true;
+            }
+         }
+
          std::string request_str;
          request_str.reserve(512 + body.size());
          request_str.append(method);
          request_str.append(" ");
          request_str.append(url.path);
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks @alisonpetersonhq. I've opened #2805, which builds on this and #2778 (which reached the same core fix a few days earlier). You're both credited as co-authors on the commit.

The CRLF-aware pre-scan here was the right call, and #2805 keeps it: skipping fields that `header_field_has_crlf` will drop is what stops an injected `Host` from suppressing the derived one and putting an HTTP/1.1 request with no Host at all on the wire. Test placement in `http_response_splitting_test` was right too — that's where #2805 put them.

#2805 adds on top:

- Caller repeats are reduced to one field. `glz::http_headers` keeps repeats, so two caller `Host` fields both reached the wire. RFC 9112 3.2 admits exactly one, and glaze's own server answers a repeat with 400.
- A caller `Host` is written in the Host slot rather than in the header loop, so it stays the first field after the request-line (RFC 9112 5).
- Test field counting matches whole field lines rather than substrings, so a `find("Host:")` can't be satisfied by an `X-Forwarded-Host:` added later.

Closing in favor of #2805 — thanks for the fix.

PR #2807, opened 2026-08-16
escape control characters in beve, cbor, bson, and eetf converters

uwezkhan:
`beve_to_json`, `cbor_to_json`, `bson_to_json`, and `eetf_to_json` emit decoded strings through `to<JSON, string_view>` using the caller's inherited options. `escape_control_characters` is off by default, so a control byte (0x00 to 0x1F) in a decoded value was written without an escape, and a control byte is legal inside a binary-format string. The converter reported success and produced a document Glaze's own reader rejects.

What came out was worse than "unescaped". Strings long enough for the SWAR or SIMD path get `char_escape_table[c]` memcpy'd, and that entry is **zero** for a control character with no short escape, so one byte became two NULs and the string changed length. Short strings on the scalar tail passed the byte through raw. The two paths disagreed.

A pre-existing eetf test was pinning exactly this. `"eetf_to_json check big-endian size"` converts a 1000-byte binary of NULs and asserted success; on `main` it produces 1995 bytes of which 1993 are NUL.

## The approach

Refuse, rather than either escaping by default or continuing to emit bytes that will not re-parse.

A control character with no two-character JSON escape now fails with **`error_code::invalid_control_character`**. The ones that do have an escape (newline, tab, backspace, form feed, carriage return) are unaffected, so ordinary text never trips it.

Escaping was the obvious fix and it is the wrong default. It does not merely make the output parse, it makes it acceptable *everywhere*: a NUL that most strict parsers would have rejected becomes an escape every parser takes, and lands in whatever reads the JSON back, where an embedded null truncates C APIs and splits log lines. That is a decision for the caller, not for the converter.

Not escaping is not a policy either. It yields no error code and nothing to hook a decision onto, just malformed bytes and the hope that something strict is downstream.

So the two are separated. **Encoding** is `escape_control_characters`, inherited rather than pinned, and it is how a caller opts into carrying such bytes across:

```c++
struct escaping : glz::opts {
   bool escape_control_characters = true;
};

glz::beve_to_json<escaping{}>(beve, json);   // control bytes escaped, round-trips
```

That case is real. A control character is legal in a length-prefixed BEVE, CBOR, or BSON string; a `std::string` holding one round-trips through `write_beve`/`read_beve`, and escaping is the only way the JSON form can represent it. It is now an explicit choice.

`raw_string` and `unquoted` stay pinned off. Both produce output that is not JSON regardless of what the caller wants.

`jsonb_to_json` keeps escaping and keeps validating UTF-8. A JSONB blob is the storage form of a JSON document that could itself have carried an escape, so a control character there is expected rather than anomalous.

## Cost

The check rides a branch the writer already takes. Only a quote, a backslash, or a control character reaches the escape path at all, so nothing is added to the scan. Measured on an M-series mac, clang -O3, `beve_to_json` of a clean 1500 byte payload:

| | ns/op |
|---|---|
| `origin/main` | 121 |
| this branch, default | 122 |
| this branch, `escape_control_characters` | 147 |

The default also beats the escaping path, because it no longer pays the 6n worst-case reservation that escaping forces.

Setting the error cannot return early out of `write_escape`: the SIMD loops locate the byte and rely on the lambda advancing `c`, so an early return makes the scan find the same byte forever. It flags and lets the write finish; the output is abandoned.

## Bounded output buffers

Under `escape_control_characters` the writer reserves `10 + 6n` for a string. A resizable buffer only over-allocates against that ceiling, but `JSONBuffer` is unconstrained: for a `std::array` or `std::span`, `ensure_space` reports `buffer_overflow` instead of growing, so the ceiling rejects output that fits. 216 bytes of escaped `beve_to_json` output did not fit in a 512 byte array.

6 bytes per character is reached only by a string of nothing but control characters, so bounded buffers measure the string instead, bracketed between two O(1) bounds and paid only when the answer is in doubt: the ceiling fitting means yes, the floor of one byte per character not fitting means no. `escaped_string_size` reuses the writer's own 8-byte block scan rather than counting byte at a time, which cost about as much as the escaping write it sizes. A test pins it against the writer's actual output for every byte value at offsets landing in both the block scan and the scalar tail.

Scoped to `escape_control_characters`. Plain `write_json` keeps the reservation it has today, for resizable and bounded buffers alike.

## Atom object keys in eetf

Separate bug, no option involved. `eetf_to_json` dumped the atoms `true` and `false` as bare literals in key position, so a map keyed by the atom `true` converted to `{true:1}` and reported success. A JSON object key must be a string; `term_to_json_value` takes a `Key` flag and keeps the atom quoted there. Values are unchanged.

## UTF-8

Not checked by beve/cbor/bson/eetf. `validate_utf8` is a read-side option (every other consumer is in `read.hpp`, `parse.hpp`, or `lazy.hpp`), it is on by default there, and unlike the escaping decision it costs a real pass: ~8.5 ns per string plus ~0.015 ns/byte ASCII, ~0.1 ns/byte multibyte. `jsonb_to_json` is the exception and keeps the check it has on `main`.

## Docs

`docs/binary.md` gains an "Untrusted Strings" section covering what is pinned, what the default refuses, and the tradeoff `escape_control_characters` buys. `docs/options.md` documents the new `reject_control_characters` option and updates `escape_control_characters` and `validate_utf8`.

### DIFF
--- a/docs/EETF/erlang-external-term-format.md
+++ b/docs/EETF/erlang-external-term-format.md
@@ -19,7 +19,7 @@ std::string json{};
 auto ec = glz::eetf_to_json(term, json);
 ```
 
-An Erlang binary holds arbitrary bytes by definition, and `binary_as_base64` is off by default, so a binary is emitted as a JSON string. A control character (0x00 to 0x1F) among those bytes has no two-character JSON escape, so the conversion fails with `error_code::invalid_control_character` rather than produce output that will not re-parse.
+`binary_as_base64` is off by default, so an Erlang binary is written as a JSON string. Binaries hold arbitrary bytes, and a control character (0x00–1F) among them has no short JSON escape, so the conversion fails with `error_code::invalid_control_character`.
 
 Two ways across:
 
@@ -36,6 +36,6 @@ struct escaping : glz::eetf::eetf_opts {
 glz::eetf_to_json<escaping{}>(term, json);
 ```
 
-See [Untrusted Strings](../binary.md#untrusted-strings) for what that trade buys and costs.
+See [String Escaping](../binary.md#string-escaping).
 
-Only strings, atoms, and binaries are accepted as object keys, since a JSON key must be a string. The atoms `true` and `false` emit as bare JSON literals in value position, and stay quoted in key position.
+Only strings, atoms, and binaries work as object keys, since a JSON key must be a string. The atoms `true` and `false` are written as bare JSON literals, except as a key, where they stay quoted.
--- a/docs/EETF/erlang-external-term-format.md
+++ b/docs/EETF/erlang-external-term-format.md
@@ -10,3 +10,32 @@ If you are not using CMake the macro `GLZ_ENABLE_EETF` must be added to utilize 
 
 ## [See Unit Tests](https://github.com/stephenberry/glaze/blob/main/tests/eetf_test/eetf_test.cpp)
 
+## EETF to JSON Conversion
+
+`glz::eetf_to_json` converts a buffer of Erlang terms directly to a buffer of JSON.
+
+```c++
+std::string json{};
+auto ec = glz::eetf_to_json(term, json);
+```
+
+An Erlang binary holds arbitrary bytes by definition, and `binary_as_base64` is off by default, so a binary is emitted as a JSON string. A control character (0x00 to 0x1F) among those bytes has no two-character JSON escape, so the conversion fails with `error_code::invalid_control_character` rather than produce output that will not re-parse.
+
+Two ways across:
+
+```c++
+// binary_as_base64 is a member of eetf_opts, so set it inline.
+// Usually the right answer for a binary: no raw bytes end up in a JSON string.
+glz::eetf_to_json<glz::eetf::eetf_opts{.binary_as_base64 = true}>(term, json);
+
+// escape_control_characters is inheritable, so add it to a custom opts struct.
+// Keeps the bytes as JSON string content, escaped.
+struct escaping : glz::eetf::eetf_opts {
+   bool escape_control_characters = true;
+};
+glz::eetf_to_json<escaping{}>(term, json);
+```
+
+See [Untrusted Strings](../binary.md#untrusted-strings) for what that trade buys and costs.
+
+Only strings, atoms, and binaries are accepted as object keys, since a JSON key must be a string. The atoms `true` and `false` emit as bare JSON literals in value position, and stay quoted in key position.
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -480,32 +480,42 @@ Writing is Version 2 only. Version 2 output is not decodable as a variant by a G
 
 `glaze/binary/beve_to_json.hpp` provides `glz::beve_to_json`, which directly converts a buffer of BEVE data to a buffer of JSON data.
 
-### Untrusted Strings
+### String Escaping
 
-A converter turns a document you received into JSON text. Two of its string options are pinned, because inheriting them would let a caller produce output that is not JSON at all: `raw_string` (emit the payload unescaped) and `unquoted` (drop the quotes). This applies to `glz::cbor_to_json`, `glz::bson_to_json`, and `glz::eetf_to_json` as well.
-
-**Control characters are refused by default.** A BEVE string may hold any byte, including a control character (0x00–0x1F). Most have a two-character JSON escape — `\n`, `\t`, `\b`, `\f`, `\r` — and those pass through normally. The rest cannot be written without `\uXXXX`, so rather than emit bytes that will not re-parse, the conversion fails with `error_code::invalid_control_character`.
+A BEVE string can hold any byte, including control characters (0x00–1F). JSON cannot. `\n`, `\t`, `\b`, `\f` and `\r` are escaped normally, but the rest have no short escape, and by default the conversion fails rather than write something that will not parse:
 
 ```c++
 // a BEVE string holding a 0x01 byte
-auto ec = glz::beve_to_json(beve, json);   // ec == invalid_control_character
+std::string json{};
+auto ec = glz::beve_to_json(beve, json);
+// ec.ec == glz::error_code::invalid_control_character
 ```
 
-**`escape_control_characters` opts into carrying them across.** The byte is written as `\uXXXX`, the document round-trips, and a value that legitimately contains control bytes survives:
+To keep those bytes, turn on `escape_control_characters`. They are written as `\uXXXX` and the document round-trips:
 
 ```c++
 struct escaping : glz::opts {
    bool escape_control_characters = true;
 };
 
+std::string json{};
 auto ec = glz::beve_to_json<escaping{}>(beve, json);   // {"k":"a\u0001b"}
 ```
 
-Worth knowing what that buys and costs. A control character is legal inside a length-prefixed BEVE, CBOR, or BSON string, so `std::string("a\0b", 3)` round-trips through `write_beve`/`read_beve`, and escaping is the only way the JSON form can represent it. But escaping also means a NUL in the source becomes a NUL in whatever reads the JSON back, where an embedded null can truncate a C API or split a log line. The default refuses so the choice is yours to make explicitly.
+Which to pick:
 
-For Erlang binaries, which hold arbitrary bytes by definition, `binary_as_base64` is usually the better answer than escaping: it carries any payload across without putting raw bytes in a JSON string at all.
+- **Your data legitimately contains control bytes.** Use `escape_control_characters`. A `std::string` holding a `\0` round-trips through BEVE, and escaping is the only way JSON can carry it.
+- **The bytes are arbitrary,** as in an Erlang binary. Use `binary_as_base64` instead, so no raw bytes end up in a JSON string at all.
+- **You want to know when they appear.** Keep the default and check for `invalid_control_character`.
 
-**UTF-8 is not checked.** That is the reader's job, it is on by default there via `validate_utf8`, and unlike the escaping decision it costs a real pass over every string. `glz::jsonb_to_json` is the exception on both counts: it validates UTF-8 and escapes control characters by default, because a JSONB blob is the storage form of a JSON document that could itself have carried `\uXXXX`.
+One thing to weigh when escaping: `\u0000` is valid JSON that every parser accepts, so a NUL in your data reaches whatever reads the JSON back. Embedded nulls can truncate C strings and split log lines.
+
+The same applies to `glz::cbor_to_json`, `glz::bson_to_json`, and `glz::eetf_to_json`. `glz::jsonb_to_json` always escapes, because a JSONB blob stores a JSON document that may have contained `\uXXXX` to begin with.
+
+**Other notes**
+
+- `raw_string` and `unquoted` are ignored by these converters. Both would produce output that is not JSON.
+- UTF-8 is not checked, except by `glz::jsonb_to_json`. Reading the JSON checks it, and `validate_utf8` is on by default there.
 
 ### Function Pointers
 
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -480,6 +480,10 @@ Writing is Version 2 only. Version 2 output is not decodable as a variant by a G
 
 `glaze/binary/beve_to_json.hpp` provides `glz::beve_to_json`, which directly converts a buffer of BEVE data to a buffer of JSON data.
 
+### String Escaping
+
+A BEVE string may hold any byte, including a control character (0x00–0x1F), which JSON may not. The converter escapes those as `\uXXXX` so its output re-parses, and it does so regardless of the options you pass: `escape_control_characters`, `raw_string`, and `unquoted` are inheritable, so honoring them here would hand you a switch that turns the converter's output into something Glaze's own JSON reader rejects. `glz::cbor_to_json`, `glz::bson_to_json`, `glz::eetf_to_json`, and `glz::jsonb_to_json` behave the same way.
+
 ### Function Pointers
 
 Objects that expose function pointers (both member and non-member) through `glz::meta` are skipped by the BEVE writer by default. This mirrors JSON/TOML behaviour and avoids emitting unusable callable placeholders in binary payloads.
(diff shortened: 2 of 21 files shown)


fregate:
Hi! After https://github.com/stephenberry/glaze/pull/2761 need rebase and add it to binary too for eetf

stephenberry:
Rebased onto current `main` and pushed to this branch (clean rebase, no conflicts). Changes on top of the original commit:

- **`ERL_BINARY_EXT` now covered**, per @fregate's note. `#2761` landed after this branch forked, and its non-base64 path was the one string emit still inheriting the caller's options. Since `binary_as_base64` defaults to off, that is the default path for a payload that holds arbitrary bytes by definition.
- **Updated a test that pinned the old behavior.** `"eetf_to_json binary key map explicit convert"` asserted a raw `0x01` byte in the output, so it would have failed on merge. It now expects the escaped form. Added `"eetf_to_json binary escapes control characters"`, which also round-trips through `read_json`, plus an explicit `glaze/json/read.hpp` include the file had been getting transitively.
- **Deduplicated the helper.** `jsonb_to_json` now uses the shared `detail::raw_string_emit_opts` instead of keeping its own identical copy, so the pinning is defined once for all five converters. PR description updated to match.

Verified locally with `glaze_EETF_FORMAT=ON`: eetf, beve, cbor, bson, and jsonb suites all pass.

Two follow-ups I left out of scope: invalid UTF-8 is still unhandled in these four converters (`jsonb`'s `emit_raw_string_as_json` validates it, the others only pin the options), and `escape_control_characters` moves the writer's reservation from `10 + 2*n` to `10 + 6*n`, which is a real memory bump on large payloads.

uwezkhan:
Checked the rebased branch and it looks right to me. The ERL_BINARY_EXT path was the one emit I'd missed since the branch forked before #2761, and moving jsonb onto the shared helper beats keeping two copies of the same pinning. Nothing outstanding on my side; agreed the UTF-8 validation gap and the larger writer reservation are better handled separately.

wilkolbrzym-coder:
Quick review notes – PR #2807

**No merge-blocking bugs found.**

The fix is correct: **`detail::raw_string_emit_opts<Opts>`** pins **`escape_control_characters=true`**, **`raw_string=false`**, and **`unquoted=false`** at the emit site, so untrusted binary payloads always produce strict JSON regardless of caller options.

Coverage is complete:
- **BEVE**: value, map key, string-array element
- **CBOR**: definite and indefinite text strings
- **BSON**: single **`emit_json_string`** choke point
- **EETF**: **`ERL_STRING_EXT`**, atoms, and **`ERL_BINARY_EXT`** (default non-base64 path)
- **jsonb_to_json** now uses the shared helper

Tests are solid: each format round-trips through **`read_json`**, proving output is valid JSON. All 62 checks pass.

Minor observations:
- Invalid UTF-8 remains unvalidated in beve/cbor/bson/eetf (jsonb validates). Pre-existing, not introduced here — worth a follow-up.
- Forcing **`escape_control_characters`** increases reservation to **`10 + 6*n`** on every untrusted string emit. Acceptable for correctness, but large payloads may see extra allocation.
- The helper is in **`detail`**; consider a short note for future converters.

Overall LGTM.

stephenberry:
*(Trimmed to its bounded-buffer content. The rest described an earlier design where the converters escaped by default, which was later reversed; see the newer comment and the description.)*

Under `escape_control_characters` the JSON string writer reserves `10 + 6n`. Resizable buffers only over-allocate against that ceiling, but `JSONBuffer` is unconstrained: for a `std::array` or `std::span`, `ensure_space` reports `buffer_overflow` instead of growing, so the ceiling rejects output that fits. 216 bytes of escaped `beve_to_json` output would not go into a 512 byte array.

Only a string of nothing but control characters actually reaches 6 bytes per character, so bounded buffers measure the string instead, bracketed between the ceiling and a one-byte-per-character floor so the pass is paid only when the answer is in doubt. `escaped_string_size` uses the writer's own 8-byte block scan rather than counting byte at a time, which cost about as much as the escaping write it sizes.

That same 216 bytes now converts into a `std::array<char, 512>` at ~58 ns/op, against ~56 ns/op into a `std::string`. Scoped to `escape_control_characters`, so plain `write_json` keeps the reservation it has today for both resizable and bounded buffers.

wilkolbrzym-coder:
Review notes – PR #2807 {final}
No merge-blocking bugs found.

The shared emit policy is clean:

- `untrusted_string_emit_opts` pins `escape_control_characters=true`, `raw_string=false`, `unquoted=false`
- `emit_untrusted_string` adds UTF-8 validation
- `emit_untrusted_bytes` intentionally skips it

That split correctly maps to format guarantees.

`escaped_string_size` is the riskiest piece, but the SWAR scan matches the writer and the all-byte/all-length test gives strong confidence. The bounded-buffer reservation logic is also correct and keeps resizable buffers unchanged.

EETF atom-key fix is correct and closes a real invalid-JSON path.

Open question, not a blocker: byte-defined payloads (`ERL_BINARY_EXT`, `ERL_ATOM_EXT`, `ERL_STRING_EXT`) can still produce non-UTF-8 JSON when `binary_as_base64` is off. That is documented, but worth confirming it is the intended converter contract.

Default `validate_utf8=true` adds a measurable scan to the four converters that previously ignored it. Correctness justifies it, but it should be called out in release notes.

Verdict: LGTM once CI is green.

uwezkhan:
Pulled the latest head and the cbor, beve, bson, and jsonb suites all pass here too. The rename is right, raw_string_emit_opts did read backwards for something that turns raw_string off. Measuring the string only for bounded buffers keeps the std::string path unchanged, which is the tradeoff I'd want, and the UTF-8 split between text-promising and byte-defined payloads matches what the formats actually guarantee.

stephenberry:
Changed the approach here, and rewrote the description to match. Summarising why, since the earlier commits went the other way.

Escaping control characters makes the output parse, but it also makes it acceptable *everywhere*. A NUL that most strict parsers would have rejected becomes an escape every parser takes, and lands in whatever reads the JSON back, where an embedded null truncates C APIs and splits log lines. Pinning the escape on made that the default for every converted value, and that is the caller's decision, not the converter's.

Reverting to no escaping is not the answer either. That path yields no error and nothing to act on, just malformed bytes and the hope that something strict is downstream. It is also not a faithful passthrough: on the SWAR and SIMD paths the writer memcpys `char_escape_table[c]`, which is zero for a control character with no short escape, so one byte becomes two NULs and the string changes length. Short strings on the scalar tail keep the raw byte. The two paths disagree.

Worth seeing how far that reaches: `"eetf_to_json check big-endian size"`, a test that predates this branch, converts a 1000-byte binary of NULs and asserts success. On `main` that produces 1995 bytes of which 1993 are NUL. A length-decode test was quietly pinning a kilobyte of nulls turning into invalid JSON.

So encoding and policy are now separate knobs:

- **Default: refuse.** A control character with no two-character escape fails with `error_code::invalid_control_character`. Newline, tab, backspace, form feed and carriage return have escapes and are unaffected, so ordinary text never trips it.
- **`escape_control_characters`: carry it across.** Inherited rather than pinned. Legitimate case, since a control byte is legal in a length-prefixed BEVE/CBOR/BSON string and escaping is the only way JSON can represent it. Now an explicit choice.
- **`raw_string` and `unquoted` stay pinned off.** Both produce output that is not JSON regardless of what the caller asks for.
- **`jsonb_to_json` is the exception** and keeps escaping plus its UTF-8 check. A JSONB blob is the storage form of a JSON document that could itself have carried an escape.

The check is free in practice. Only a quote, a backslash, or a control character reaches the escape path at all, so nothing is added to the scan: `beve_to_json` of a clean 1500 byte payload is 122 ns/op against 121 on `main`, and the default now beats the escaping path outright because it no longer pays the 6n reservation.

Also dropped the UTF-8 validation I had added to beve/cbor/bson/eetf. `validate_utf8` is a read-side option, on by default there, and unlike this it costs a real pass, roughly 8.5 ns per string. Wrong trade for a write path. `jsonb_to_json` keeps the check it already had.

Unrelated bug fixed in passing: eetf dumped the atoms `true` and `false` bare in key position, so a map keyed by the atom `true` converted to `{true:1}` and reported success.

Full suite green, plus ASan/UBSan on the converter suites and eetf with `glaze_EETF_FORMAT=ON`.

wilkolbrzym-coder (commented):
## Deep review – PR #2807

The shift from forced escaping to a **default-refuse** policy is the correct architectural decision. It eliminates silent NUL propagation while preserving an explicit opt-in path via `escape_control_characters`. The bounded-buffer measurement logic and EETF atom-key fix are both well executed.

That said, two correctness gaps remain within the same trust boundary this PR is hardening. They are not theoretical.

---

### 1. Invalid UTF-8 still produces invalid JSON with a success return

The converter now rejects control characters, but **does not validate UTF-8** for BEVE, CBOR, BSON, or EETF. A lone continuation byte (or a latin1 atom) is emitted as-is, producing a document that Glaze’s own reader rejects.

```cpp
std::map<std::string, std::string> v{{"k", std::string("a\x80z")}};
auto beve = glz::write_beve(v);
std::string json{};
auto ec = glz::beve_to_json(beve, json);
// ec == none, but json contains raw 0x80
// glz::read_json(json) fails with invalid_utf8
```

This is the **same failure mode** as the original control-character bug: converter reports success, output does not re-parse. I understand `validate_utf8` is intentionally a read-side check and was omitted to avoid a real scan cost, but that omission must be tracked as a follow-up with a release-note warning. The same applies to Erlang byte-defined payloads when `binary_as_base64` is off.

---

### 2. Short-escape control characters lack converter test coverage

The default-refuse policy should reject only control characters **without a two-character JSON escape**. Characters like `\n`, `\t`, `\b`, `\f`, `\r` must continue to pass through normally.

The raw-tail path has:

```cpp
if constexpr (check_reject_control_characters(Opts)) {
    if (uint8_t(value) < 0x20) [[unlikely]] {
        ctx.error = error_code::invalid_control_character;
    }
}
```

If `\n` or `\t` ever reaches this branch, it will be incorrectly rejected. I do not see a test pinning this for any converter. Add explicit coverage:

```cpp
std::map<std::string, std::string> v{{"k", "line1\nline2\tend"}};
// expect all four converters to succeed
```

---

### 3. Interaction between `escape_control_characters` and `reject_control_characters` needs explicit precedence

`untrusted_string_emit_opts` sets the internal `reject_control_characters` bit **unconditionally**, while `escape_control_characters` remains inherited:

```cpp
template <auto Opts>
inline constexpr auto untrusted_string_emit_opts =
   glz::reject_control_characters<
       opt_false<opt_false<Opts, raw_string_opt_tag{}>, unquoted_opt_tag{}>>();
```

If a caller enables `escape_control_characters`, both flags are now set. The existing tests appear to pass, so escape likely takes precedence, but that is not documented. A future refactor could easily reverse the order and make escaped output fail with `invalid_control_character`. Add either a static assertion or a dedicated test that verifies `escape_control_characters` wins.

---

### What is solid

- `raw_string` and `unquoted` pinned off – correct.
- New `invalid_control_character` error code and its documentation – clear.
- `escaped_string_size` bounded-buffer logic – sound, and the exhaustive byte/length test is excellent.
- EETF atom-key quoting – correct and well tested.
- CI is essentially green; only one RPM job still in progress.

---

### Verdict

**No merge-blocking bug found**, but issues 2 and 3 require confirmation before I can approve this in good conscience. Issue 1 must be captured as a follow-up. The design is strong, and the new policy is a significant improvement.

LGTM after the above are addressed.

stephenberry:
Thanks for the read. Two of the three points don't hold, and I'd like the reasoning on the thread for anyone reading it later.

**2. Short-escape control characters.** The snippet is quoted without its enclosing branch. In the scalar tail the reject check is the `else` of

```cpp
if (const auto escaped = char_escape_table[uint8_t(*c)]; escaped) { ... }
```

so `\n \t \b \f \r` are consumed before it can be reached, and in `write_escape` the condition is explicitly `char_escape_table[uint8_t(*c)] == 0`. There is no input that reaches the reject carrying a short-escapable byte. Confirmed on both the scalar tail and the SIMD/SWAR path: `{"k": "line1\nline2\tend"}` converts with `ec == none` and round-trips. A test is still worth having, since the guard depends on a lookup two branches up, so I pushed one (f75bb20) covering all four converters. Each case exercises both writer paths: a short value on the scalar tail, and a long one that puts the run into the block scan, which rejects at a separate site. Inverting either guard to a bare `< 0x20` fails them.

**3. Precedence.** It isn't incidental. Every site is `if constexpr (check_escape_control_characters(Opts)) { escape } else { reject }`, so escaping wins by construction. It's documented at the flag in `opts.hpp` ("Only consulted when `escape_control_characters` is off"), and all four converters already have a test that sets both and asserts the escaped output plus a `read_json` round-trip. The proposed `static_assert` would break the feature: enabling `escape_control_characters` on a converter is exactly the case where both bits are set on purpose.

**1. UTF-8.** This one reproduces, but it isn't a gap this PR opened. `write_json` on the same `std::string` emits the same bytes with the same success return:

```
beve_to_json: ec == none, read_json -> invalid_utf8
write_json:   ec == none, read_json -> invalid_utf8
```

No writer in Glaze validates UTF-8; readers do, on by default. The comparison to the control-character bug doesn't fit either. That bug reshaped the payload, writing one byte as two NULs, with the SIMD and scalar paths disagreeing about it. Invalid UTF-8 passes through byte-exact. Whether the converter trust boundary should be stricter than `write_json` is a fair question, and the PR description covers why that isn't part of this change, along with the measured cost.

stephenberry:
The read side always needs to be guarded, but we don't automatically guard writing, because it isn't a security issue, it is a developer problem. We don't want to make developers always pay a performance cost for checking valid outputs. Checking what a developer's code produces is up to them to decide whether there is any danger and they need checks, or if checking can just be performed on the read. Write errors tend to be much more rare because the developer's own program is generating the output. And, the developer can do their own validation for their specific needs.

wilkolbrzym-coder (commented):
Thanks for the follow-up. The added tests cover both the scalar tail and block-scan paths, and the precedence is now explicit. No remaining concerns from my side. Once CI is fully green, this is ready to merge.

uwezkhan:
Thanks for carrying this one through to merge, the refuse-by-default split ended up in a better place than where I started it.

PR #2809, opened 2026-08-17
Refactor FindErlang CMake module

fregate:
Refactor FindErlang CMake module

Remove some issues, more clearer behavior and result. Add inplace documentation for developers.

**The reason for this PR**: During real-world use of this module in embedded systems, several issues were identified. These issues led to a redesign of the module into its current form, which provides a more flexible and predictable build process for both conventional systems and embedded systems using cross-compilation.

### DIFF
--- a/cmake/FindErlang.cmake
+++ b/cmake/FindErlang.cmake
@@ -1,162 +1,257 @@
 #[=======================================================================[.rst:
 FindErlang
--------
+----------
 
-CMake Find module that locates the Erlang/OTP runtime and erlang_interface (EI) library for building NIFs
-and EI-based projects.
+Finds the Erlang/OTP runtime and the ``erl_interface`` (EI) library, for
+building NIFs and EI-based programs.
 
-Input Variables (Hints)
-^^^^^^^^^^^^^^^^^^^^^^^
-┌───────────────────┬─────────────────────────────────────────────────────────────────────────┐
-│Variable           │Description                                                              │
-├───────────────────┼─────────────────────────────────────────────────────────────────────────┤
-│ERLANG_ROOT        │Root path of Erlang installation. If set, overrides erl auto-detection.  │
-├───────────────────┼─────────────────────────────────────────────────────────────────────────┤
-│ERLANG_EI_LIB      │Direct path to libei.a (erlang_interface library).                       │
-├───────────────────┼─────────────────────────────────────────────────────────────────────────┤
-│ERLANG_EI_INCLUDE  │Direct path to erlang_interface include directory.                       │
-├───────────────────┼─────────────────────────────────────────────────────────────────────────┤
-│$ENV{ERLANG_HOME}  │Environment variable for Erlang install root.                            │
-└───────────────────┴─────────────────────────────────────────────────────────────────────────┘
+Two discovery modes are supported:
 
-If ERLANG_EI_LIB and ERLANG_EI_INCLUDE are both provided, auto-detection via erl is skipped entirely.
-Otherwise, the module searches for the erl executable and queries it for paths.
+* **Auto-detection** (default): locate the ``erl`` executable and query it for
+  the OTP layout. Requires an ``erl`` that can run on the build host.
+* **Explicit paths**: supply ``ERLANG_EI_LIB`` and ``ERLANG_EI_INCLUDE`` and
+  ``erl`` is never invoked. This is the supported path for cross-compiling to
+  an embedded target, where the target's ``erl`` cannot run on the host.
 
-Search Paths for erl
-^^^^^^^^^^^^^^^^^^^^
-$ENV{ERLANG_HOME}/bin, ${ERLANG_ROOT}/bin, /opt/bin, /sw/bin, /usr/bin, /usr/local/bin, /opt/local/bin
+Hint Variables
+^^^^^^^^^^^^^^
 
+``Erlang_ROOT``
+  Root of the Erlang installation. Honored automatically by CMake's ``find_``
+  commands (policy ``CMP0074``); this is the preferred spelling.
+``ERLANG_ROOT``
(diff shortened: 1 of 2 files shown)


stephenberry:
Thanks for this — the cross-compile motivation is right, and `ERLANG_EI_LIB`/`ERLANG_EI_INCLUDE` is the correct escape hatch for it. Nice catch on `${Erlang_RUNTIME}` too: `main` was querying bare `erl` from `PATH` even after `find_program` had located a specific one.

I pushed a follow-up commit (a850765) fixing four things I hit while reviewing.

**1. `Erlang::Erlang` include directory.** `<root>/usr/include` → `<root>/include`. OTP installs the NIF headers to `usr/include` — see [`erts/emulator/Makefile.in`](https://github.com/erlang/otp/blob/master/erts/emulator/Makefile.in) (`RELEASE_INCLUDES`). There's no `<root>/include` on a stock install, so `#include <erl_nif.h>` through that target fails, and linking it directly trips CMake's "imported target includes non-existent path" at generate time.

Worth knowing why CI didn't catch it: glaze routes the target through the `glaze_glaze` INTERFACE library, and that indirection launders CMake's non-existent-path check. glaze only ever includes `<ei.h>`, so the dead `-I` never mattered to `eetf_test`. The Erlang jobs really did run and pass — they just can't see this one.

**2. The two `message(FATAL_ERROR)` calls.** A find module has to set `Erlang_FOUND=FALSE` and return; failing the build is `REQUIRED`'s job, via FPHSA. This bites harder than usual because `install-rules.cmake` embeds this file verbatim into `glazeConfig.cmake`, where `Erlang_FIND_REQUIRED` is never set — so a downstream `find_package(glaze)` on a machine without Erlang aborted instead of degrading. (The `libei.a` check also sat outside `if(Erlang_FOUND)`, firing after FPHSA had already said not-found.)

**3. `ERLANG_EI_LIB`.** Documented as the path to `libei.a`, but the code appends `/libei.a`, so following the docs gives `.../libei.a/libei.a`. Both spellings are accepted now.

**4. OTP root derivation.** `${ERLANG_EI_INCLUDE}/../..` is off by one — EI headers are at `<root>/lib/erl_interface-x.y.z/include`, so it lands on `<root>/lib`. On a hand-assembled sysroot there's no reliable relationship anyway, so it now comes from `ERLANG_ROOT`/`Erlang_ROOT` when given, and is left unset otherwise.

Along the way: `find_path`/`find_library` replace the layout assumptions, which is what makes the `FATAL_ERROR`s unnecessary and lets a shared `libei` work; `UNKNOWN IMPORTED` drops the non-namespaced `erlang_ei` target that was leaking into consumers' global namespace; `execute_process` got `RESULT_VARIABLE`/`ERROR_QUIET`/`OUTPUT_STRIP_TRAILING_WHITESPACE`; FPHSA uses `REQUIRED_VARS`/`VERSION_VAR`; the success diagnostics moved behind `CMAKE_VERBOSE_MAKEFILE` to match how you'd already guarded the input ones; `execute_erlang` → `_erlang_query` so `glazeConfig.cmake` doesn't export a generic name.

NIF headers are optional now, so `Erlang::Erlang` just carries no include directory when `erl_nif.h` isn't found and an EI-only cross build still links. `ERLANG_INCLUDE` overrides the location.

I kept your removal of `Erlang::ERTS`, `Erlang_COMPILE`, `Erlang_EI_PATH` and `Erlang_ERTS_*` — `liberts.a` isn't shipped as a linkable library, so `Erlang::ERTS` never resolved on a stock install — but noted it in the module docs since it's a downstream break.

Verified against a real OTP 29 install (`eetf_test` passes, macOS/Homebrew layout, which no CI job covers), plus a synthetic OTP tree covering auto-detection, both `ERLANG_EI_LIB` spellings, shared `libei`, `Erlang_ROOT` via `CMP0074`, and optional/`REQUIRED`/embedded behavior with `erl` unreachable.

Happy to restore any of the removed variables, or add a different hint if the embedded layouts you're targeting need something other than `ERLANG_INCLUDE`.

fregate:
I've checked this in real application and it works. So, for me this is ok. Merge?

fregate:
Hi @stephenberry! Have you something to add to this PR? Merge?

stephenberry:
Pushed one more commit (2b8e587). Two fixes, neither from your work — both reproduce identically on `main`:

**Missing Erlang surfaced too late.** `glazeConfig.cmake` never checked `Erlang_FOUND`, so a consumer without Erlang got `-- Configuring done` followed by a generate-time `The link interface of target "glaze::glaze" contains: Erlang::EI but the target was not found`. It now sets `glaze_FOUND FALSE` with a `glaze_NOT_FOUND_MESSAGE` naming `ERLANG_EI_LIB`/`ERLANG_EI_INCLUDE`, so `REQUIRED` fails at the `find_package` with a usable reason and a plain `find_package(glaze)` degrades.

**FPHSA name-mismatch warning.** Because `install-rules.cmake` embeds the module verbatim, `find_package_handle_standard_args(Erlang)` runs while `CMAKE_FIND_PACKAGE_NAME` is `glaze`; FPHSA reads that as a likely typo and warned on every consumer's configure. Marked deliberate with `FPHSA_NAME_MISMATCHED`.

Verified against OTP 29: in-tree `eetf_test`, and an install plus a downstream `find_package(glaze)` with Erlang present, with it absent under both `REQUIRED` and plain `find_package`, and reached through `ERLANG_EI_LIB`/`ERLANG_EI_INCLUDE` with `erl` unreachable.

Thanks for this, and for validating it against a real application — the cross-compile path is a real improvement. Merging.

fregate:
Thanks to you for your deep dive into this PR! (I guess from my initial commit there is only my name as author 🙂 )

stephenberry:
Thanks for bringing up these improvements. Iteration is expected.

PR #2811, opened 2026-08-18
Support glz::custom for every format that can express it

### LINKED ISSUE #2786: glz::custom<READ, WRITE> for TOML?
Any plans to add this feature for TOML parsing? :)

stephenberry:
Closes #2786.

`glz::custom` only worked for JSON and BEVE. The generic `from`/`to` specializations in `core/custom.hpp` were already format independent, but they called `skip_array<Opts>` — a **format-specific** helper that only those two formats define — so the zero-argument setter form failed to compile everywhere else. Several format headers also never included `core/custom.hpp` and worked only when a JSON header happened to be included alongside.

## The fix

Discard the incoming value with `skip_value<Format>` instead. Every format defines it, and it is what a no-input handler actually wants. This also lets such a field round-trip a getter that writes a non-array value, which previously failed to re-read **even in JSON**.

Three formats needed more, so the read and write dispatch is split into `detail::custom_read` / `detail::custom_write` and each gets a thin entry point rather than restating ~150 lines:

- **MSGPACK** and **BSON** hand their readers a tag the caller already consumed, which the generic entry point cannot pass along. BSON additionally needs `to<BSON, custom_t>::type_code` (the element type byte written before the key), resolved from the getter's type via the existing `resolve_write_type_t`.
- **EETF** closes `write.hpp` with an unconstrained `to<EETF, T>` sink that has no `op`, so unsupported types fail to compile. Fixing `Format` makes it more specialized than `to<Format, custom_t>`, so it shadowed the generic one.

The `core/custom.hpp` include is now explicit in `bson`/`csv`/`eetf`/`jsonb`/`msgpack`/`yaml.hpp`. MSGPACK, BSON, JSONB, and YAML were only reaching it by accident through `core/seek.hpp`; a refactor dropping that edge would have silently reintroduced this bug for all four.

## Status by format

| Format | Before | After |
|---|---|---|
| JSON, BEVE | worked | worked |
| CBOR, JSONB, YAML | compile error | works |
| TOML | compile error | works |
| MSGPACK | read did not compile | works |
| BSON | read + write did not compile | works |
| EETF | write did not compile | works |
| CSV | not supported | still not supported |

**CSV is out of reach, and not because of this bug**: its columnar layout requires every struct field to be a container of row values — a plain `int` field already fails there. Supporting `glz::custom` would mean teaching the column writer/reader to unwrap transparent wrappers and materialize a getter result once, across rowwise × colwise × read × write. That is a feature, not a fix. `csv.hpp` still includes `core/custom.hpp` with a comment explaining why, so an attempt fails inside the CSV writer naming the real constraint rather than on an "undefined template" that reads like a missing include.

## Two TOML bugs found along the way

- A `glz::custom` getter over a nullable emitted `key = ` with nothing after it — invalid TOML that fails to re-read with `unexpected_end`. The TOML writer's null check keyed off the *member* type, which for a `custom_t` field is never `null_t`. Now uses the existing `custom_getter_returns_nullable` / `is_custom_field_null` helpers that the JSON, BEVE, and CBOR writers already use, in both writer paths.
- `toml::skip_enclosed` recursed one frame per level whenever the bracket type alternated, so skipping an unknown key like `z = [{[{...` **segfaulted**. Nesting is now tracked with an explicit stack bounded by `max_recursive_depth_limit`. Worth noting: at `-O2` clang happened to turn the old recursion into a loop, so only `-O0`/`-O1`/`-Og` crash — Release CI would not have caught it.

  This does mean skipped values are now capped at 256 nesting levels even for *same-type* nesting, which was previously unbounded. That matches `enter_depth` and the existing `recursion_depth_tests`, but it is a behavior change on input that used to be accepted.

## Tests

`tests/custom_formats/` has one translation unit per format, each including **exactly one** format header. That isolation is the point — it is what makes this class of bug detectable at all. Against the previous headers, **6 of 8 fail to compile** (CBOR, MSGPACK, BSON, JSONB, TOML, YAML); JSON and BEVE compile, as expected.

`tests/toml_test/custom_test.cpp` covers the TOML-specific surface: nullable getters in both writer paths, `std::function` handlers, `glz::custom` as the whole `meta::value`, nested objects through a `[table]` header, and the `glz::context&` error-reporting form.

EETF is wired in under `glaze_EETF_FORMAT`. That flag is off in my local config, so I built and ran that TU by hand against Erlang's headers to confirm it passes rather than assuming it.

`ctest`: 120/120 pass.


wilkolbrzym-coder:
Quick review notes – PR #2811

**No merge-blocking bugs found.**

The refactor is clean: **`detail::custom_read` / `detail::custom_write`** with callbacks correctly removes the **`skip_array<Opts>`** dependency and routes through **`skip_value<Format>`**. Format-specific entry points for **MSGPACK**, **BSON**, and **EETF** are minimal and well-commented.

TOML fixes are correct:
- Nullable custom getters now drop the key in both writer paths, avoiding **`key =`** with no value.
- **`skip_enclosed`** now uses an explicit stack bounded by **`max_recursive_depth_limit`**, fixing the stack overflow on alternating brackets. The 256-level cap is intentional, but should be noted as a behavior change.

Tests are strong: one TU per format catches include-order regressions; 6 of 8 formats failed before this patch.

Minor observations:
- Rerun the 3 cancelled CI jobs before merge.
- Confirm **`static constexpr uint8_t type_code`** is fine for all supported C++ standards.
- In **`skip_enclosed`**, verify a mismatched closing bracket reports **`syntax_error`** instead of falling through.
- Mention the 256-level skipped-value cap in release notes.

Overall LGTM.

PR #2815, opened 2026-08-19
Support arrays of class elements in volatile structs (experimental)

stephenberry:
> **Experimental — not necessarily intended for merge.** Opening this for review and discussion; I'm not yet sure this belongs in Glaze.

## The problem

`volatile T*` is not a `std::input_iterator` when `T` is a class type. `std::indirectly_readable` requires `T` be constructible from `volatile T&`, and an implicit copy constructor takes `const T&`, to which no volatile lvalue binds.

That means `glz::volatile_array<some_class, N>` (and a C array of class type inside a `volatile` struct) fails every `range`-based serialization path, and `std::span` cannot be formed over the storage either. This is the natural shape of a register block built from N identical sub-blocks:

```c++
struct channel_t
{
   uint32_t control{};
   uint32_t status{};
};

struct block_t
{
   uint32_t enable{};
   glz::volatile_array<channel_t, 4> channels{};
};
```

Scalar elements are unaffected: they copy through a built-in conversion, so those arrays stay iterable.

## The change

- `is_volatile_array` moves from `hardware/volatile_array.hpp` to `concepts/container_concepts.hpp` so the serialization concepts can see it without pulling in the hardware headers.
- New `volatile_indexed_array` concept (`is_volatile_array && !range && !meta_value_t`) selects only the class-element case.
- Index-based array readers/writers for JSON and BEVE: `read/write_json_indexed_array` and `read/write_beve_indexed_array`. Elements are visited with `value[i]`, keeping every access volatile-qualified.
- The `V (&value)[N]` C-array specializations gain an `if constexpr (requires { std::span{value, N}; })` fallback to the same index-based path.
- A `seek_op` specialization so JSON pointers (`glz::get`, `/channels/0/control`) reach into these arrays.
- Docs section in `docs/volatile-support.md`.

Fixed-extent semantics match `std::array`: a short JSON array leaves trailing elements untouched, and more input elements than `N` is `exceeded_static_array_size`.

## Tests

New suites in `beve_test` and `json_test` covering class elements, two-level nesting, C arrays of class type (with explicit `glz::meta`), JSON-pointer access, prettify, and the short/overlong array cases.

Two encoding tests pin the wire format:
- A `volatile_array<class, N>` produces byte-identical output to the equivalent `std::array<class, N>`.
- A `volatile_array<scalar, N>` still encodes as a BEVE *typed* array, not a generic one. Routing scalars through the index path would have silently changed the format.

## Open questions

- Is this worth the added surface area in the hot serialization headers, given how narrow the use case is?
- Pure reflection still cannot handle an aggregate containing a C array (the member-count probe counts the array's elements individually), so those need an explicit `glz::meta`. Documented rather than fixed here.

Local: `beve_test` 456/456 and `json_test` 729/729 pass.

### DIFF
--- a/docs/volatile-support.md
+++ b/docs/volatile-support.md
@@ -39,3 +39,53 @@ expect(obj.d == 9.9);
 expect(obj.e == 12);
 ```
 
+
+## Arrays of class elements
+
+The array elements may themselves be class types, which is the usual shape of a register block built from N identical sub-blocks:
+
+```c++
+struct channel_t
+{
+   uint32_t control{};
+   uint32_t status{};
+};
+
+struct block_t
+{
+   uint32_t enable{};
+   glz::volatile_array<channel_t, 4> channels{};
+};
+```
+
+```c++
+volatile block_t block{};
+block.enable = 1;
+block.channels[0].control = 7;
+
+std::string s{};
+glz::write_json(block, s);
+// {"enable":1,"channels":[{"control":7,"status":0}, ...]}
+
+expect(!glz::read_json(block, s));
+expect(glz::get<volatile uint32_t>(block, "/channels/0/control") == uint32_t(7));
+```
+
+These arrays are traversed by index rather than by iterator: `volatile_array<T, N>::begin()` returns a `volatile T*`, which is not a `std::input_iterator` when `T` is a class type, because `std::indirectly_readable` requires `T` be constructible from `volatile T&` and an implicit copy constructor takes `const T&`. Scalar elements copy through a built-in conversion instead, so they keep taking the ordinary iterable path and their encoding is unchanged.
+
+A plain C array of class type inside a `volatile` struct works the same way:
+
+```c++
+struct block_t
+{
+   channel_t channels[4]{};
+};
+
+template <>
+struct glz::meta<block_t>
+{
+   static constexpr auto value = object(&block_t::channels);
+};
+```
+
+Note the explicit `glz::meta`: pure reflection cannot handle an aggregate containing a C array, with or without `volatile`, because the member-count probe counts the array's elements individually. Use `glz::volatile_array` (or `std::array`) to keep reflection working.
--- a/include/glaze/beve/read.hpp
+++ b/include/glaze/beve/read.hpp
@@ -1772,6 +1772,71 @@ namespace glz
       }
    };
 
+   // Reads a BEVE array by indexing rather than by iterating.
+   //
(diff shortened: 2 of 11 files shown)


wilkolbrzym-coder (commented):
Review notes – PR #2815

Verdict: not mergeable in current state.

Blockers

1. CI red

- gcc / build-gcc16-reflection (Release)
- p2996-reflection / P2996 Reflection Tests

This PR touches shared concept and reflection paths. The P2996 failures must be investigated before merge.

2. C-array fallback condition is too weak

Current code:

if constexpr (requires { std::span{value, N}; }) {
    parse<JSON>::op<Opts>(std::span{value, N}, ctx, it, end);
} else {
    read_json_indexed_array<Opts, N>(value, ctx, it, end);
}

This only proves std::span is constructible. It does not prove the span is a Glaze range. A volatile class-type array can be span-constructible while volatile T* is not a valid std::input_iterator. That can select the broken normal path.

Safer check:

using span_t = decltype(std::span{value, N});
if constexpr (range<std::decay_t<span_t>>) {
    parse<JSON>::op<Opts>(std::span{value, N}, ctx, it, end);
} else {
    read_json_indexed_array<Opts, N>(value, ctx, it, end);
}

Same pattern appears in:

- include/glaze/json/read.hpp
- include/glaze/json/write.hpp
- include/glaze/beve/read.hpp
- include/glaze/beve/write.hpp

3. BEVE indexed reader ignores the encoded count against N

read_beve_indexed_array decodes n, then loops over N:

const size_t n = int_from_compressed(ctx, it, end);

for (size_t i = 0; i < N; ++i) {
    from<BEVE, val_t>::template op<Opts>(value[i], ctx, it, end);
}

No explicit n == N, n > N, or n < N handling.

JSON rejects too many elements with exceeded_static_array_size. BEVE does not.

If BEVE fixed-extent semantics should match std::array, either:

- validate n == N, or
- use n to control parsing and N only as an upper bound.

Current behavior is inconsistent between formats.

Non-blocking

4. ctx.depth may leak on early returns in read_json_indexed_array and write_json_indexed_array.

5. val_t removes volatile before dispatch. Current tests pass, but a copy-constrained class test would be useful.

What looks good

- Clean volatile_indexed_array concept.
- is_volatile_array moved into container_concepts.hpp.
- JSON pointer access through seek_op.
- Strong tests for volatile/non-volatile, nesting, C arrays, short/overlong JSON, BEVE wire parity, scalar volatile_array path.

cc @stephenberry

PR #2816, opened 2026-08-19
reject indefinite marker in cbor skip_argument

uwezkhan:
`skip_value<CBOR>::skip_argument` reads the argument for the uint, nint, and tag major types by switching on the 5-bit additional info. It has a `case info::indefinite` (additional info 31) that returns 0 without an error, on the assumption the caller intercepts it. Only the byte/text-string, array, and map skippers do that intercept, so on major type 0, 1, or 6 the marker reaches `skip_argument` and is accepted. A skipped map value of the single byte `0x1F` (uint with additional info 31), reachable through `read<CBOR>` with `error_on_unknown_keys = false`, and at default options through a `glz::skip{}` member (`from<CBOR, glz::skip>`) or a const member when `error_on_const_read = false`, therefore reports success on input that is not well-formed per RFC 8949 Appendix C.

Before, the skip path accepted these items while the reader's own `decode_arg` already rejected the identical bytes with `syntax_error`, so whether a value was well-formed depended on whether its key happened to be a struct member. After, `skip_argument` returns `syntax_error` for additional info 31, matching `decode_arg`. The tradeoff is that the indefinite marker on a non-container type now fails instead of being consumed as a no-op; genuine indefinite-length strings, arrays, and maps are untouched, since their skippers never route additional info 31 through `skip_argument`.

### DIFF
--- a/include/glaze/cbor/cbor_to_json.hpp
+++ b/include/glaze/cbor/cbor_to_json.hpp
@@ -321,9 +321,12 @@ namespace glz
                ctx.depth += check_indentation_width(Opts);
             }
 
+            // A prettified map only breaks the line before its '}' when it holds at least one
+            // pair, so an empty map stays "{}" whether its length was definite or indefinite.
+            bool wrote_pair = false;
+
             if (additional_info == info::indefinite) {
                // Indefinite-length map
-               bool first = true;
                while (true) {
                   if (it >= end) [[unlikely]] {
                      ctx.error = error_code::unexpected_end;
@@ -337,14 +340,14 @@ namespace glz
                      break;
                   }
 
-                  if (!first) {
+                  if (wrote_pair) {
                      dump(',', out, ix);
                   }
                   if constexpr (Opts.prettify) {
                      dump('\n', out, ix);
                      dumpn(check_indentation_char(Opts), ctx.depth, out, ix);
                   }
-                  first = false;
+                  wrote_pair = true;
 
                   // Key (must be string for JSON compatibility)
                   cbor_to_json_key<Opts>(ctx, it, end, out, ix, recursive_depth + 1);
@@ -368,6 +371,8 @@ namespace glz
                const uint64_t count = cbor_to_json_decode_arg(ctx, it, end, additional_info);
                if (bool(ctx.error)) [[unlikely]]
                   return;
+
+               wrote_pair = count > 0;
 
                for (uint64_t i = 0; i < count; ++i) {
                   if (i > 0) {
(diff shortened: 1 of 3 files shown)


wilkolbrzym-coder:
**LGTM**

This aligns the skip path with `decode_arg` and removes a real validation inconsistency: the same malformed CBOR item should never be accepted just because its key is unknown.

The fix is minimal and correctly scoped — genuine indefinite byte/text strings, arrays, and maps are intercepted before `skip_argument`, so only invalid uses of `additional_info == 31` on uint/nint/tag now fail.

Only minor suggestion: add a coverage case for an indefinite byte string (`0x5F ... 0xFF`) and/or standalone `0xFF` if that reaches the same branch.

uwezkhan:
Added an indefinite byte string (0x5F ... 0xFF) to the accept case. A standalone 0xFF doesn't reach skip_argument, it goes through the major type 7 arm which already fails it as an unexpected break, so I left that one out.

wilkolbrzym-coder:
> Added an indefinite byte string (0x5F ... 0xFF) to the accept case. A standalone 0xFF doesn't reach skip_argument, it goes through the major type 7 arm which already fails it as an unexpected break, so I left that one out.

👍

stephenberry:
Verified locally, and the divergence is real. `skip_argument` was the only place in the CBOR code that accepted additional info 31 on major types 0, 1, and 6; both `cbor_detail::decode_arg` and `cbor_to_json_decode_arg` already reject those bytes through their `default:` arm, so whether an item was well-formed depended on whether it happened to land on a struct member. Every caller that legitimately carries the indefinite marker (byte/text string, array, map) intercepts it before reaching `skip_argument`, and the writer never emits it, so nothing well-formed changes behavior.

One note: the reach is a bit wider than the description says. `skip_value<CBOR>` is also entered at default options via `from<CBOR, glz::skip>` and via const members when `error_on_const_read = false`, so `A1 61 6D 1F` against a struct with a `glz::skip{}` member was accepted by a plain `read_cbor` too.

I pushed one commit on top. While checking the sibling `cbor_to_json` path I found the prettified map writer deciding emptiness from the map's additional info, which does not answer that question: an indefinite map ends at a break code, and a definite one can carry a zero count in a following byte. Both `BF FF` and `B8 00` prettified as `{\n}`, and a nested empty map shifted the indentation of the map holding it. It now tracks whether a pair was written. Tests cover the empty and non-empty cases in both encodings.

Thanks for the fix.

uwezkhan:
Good point on the reach. I'd only traced the unknown-key path, but `from<CBOR, glz::skip>` and the const-member branch call `skip_value<CBOR>` with the same opts, so a plain `read_cbor` hits it too. Updated the description to say so. The prettified empty map fix looks right as well, additional info alone can't distinguish `BF FF` or `B8 00` from a populated map.

PR #2817, opened 2026-08-20
Fix count_members for non-default constructible fields

makaimann:
Hey `glaze` folks, thanks for all your work on this software!

I've been experimenting with `glaze` and noticed that the C++23 (i.e., non P2996 reflection) `count_members` makes some assumptions about the fields in the aggregate and can return incorrect results.

## Example

Using this minimal example:
```cpp
struct ctor_only_member
{
   ctor_only_member(int v) : v{v} {}
   int v;
};

template <>
struct glz::meta<ctor_only_member>
{
   static constexpr auto value = glz::object(&ctor_only_member::v);
};

struct non_default_constructible_second
{
   double a;
   ctor_only_member b;
};
```

### Before Fix
`glz::detail::count_members<non_default_constructible_second>` evaluates to 0.

### After Fix
`glz::detail::count_members<non_default_constructible_second>` evaluates to 2.

## Summary

The pro for making this change is resolving this unexpected behavior. The con is that we need to do some search which could increase compile times.

This PR first adds a test that fails, and then implements a fix. The approach I used was to do a galloping search up to the `max_pure_reflection_count` to find an upper bound and then do a binary search to find the number of members.

I'd definitely appreciate a maintainers' opinion on whether this is a good solution or not. It made sense to me, but if there's a better way in C++23, then great!

### DIFF
--- a/include/glaze/reflection/to_tuple.hpp
+++ b/include/glaze/reflection/to_tuple.hpp
@@ -92,19 +92,92 @@ namespace glz
          [[maybe_unused]] constexpr operator std::string_view() const { return {}; }
       };
 
-      template <class T, class... Args>
-         requires(std::is_aggregate_v<std::remove_cvref_t<T>>)
-      inline constexpr auto count_members = [] {
-         using V = std::remove_cvref_t<T>;
-         if constexpr (requires { V{Args{}..., any_t{}}; }) {
-            return count_members<V, Args..., any_t>;
-         }
-         else {
-            return sizeof...(Args);
-         }
-      }();
+      // can V be brace-initialized from exactly N any_t values?
+      template <class V, size_t N>
+      inline constexpr bool initializable_with_n = []<size_t... I>(std::index_sequence<I...>) {
+       return requires { V{(void(I), any_t{})...}; };
+      }(std::make_index_sequence<N>{});
 
       inline constexpr size_t max_pure_reflection_count = 128;
+
+      // Lowest n for which initializable_with_n is true
+      template<class V, size_t K, size_t Max>
+      consteval size_t count_low()
+      {
+       if constexpr (K > Max) {
+           return Max;
+       }
+       else if constexpr (initializable_with_n<V, K>)
+       {
+           return K;
+       }
+       else
+       {
+           return count_low<V, K+1, Max>();
+       }
+      }
+
+      // Double from a knonw-true low until initializable_with_n is false: an exclusive upper bound
+      template<class V, size_t Low, size_t Max>
+      consteval size_t count_high()
+      {
+          if constexpr (Low * 2 > Max)
+          {
+              return Max;
+          }
+          else if constexpr (initializable_with_n<V, Low * 2>)
+          {
+              return count_high<V, Low * 2, Max>();
+          }
+          else
+          {
+              return Low * 2;
+          }
+      }
+
+      // Largest true in [Low, High): Low known true, High known false
+      template<class V, size_t Low, size_t High>
+      consteval size_t count_between()
+      {
+          if constexpr (High <= Low + 1)
+          {
+              return Low;
+          }
+          else
+          {
+              constexpr size_t Mid = Low + (High - Low) / 2;
+              if constexpr (initializable_with_n<V, Mid>)
+              {
+                  return count_between<V, Mid, High>();
+              }
+              else
+              {
+                  return count_between<V, Low, Mid>();
+              }
+          }
+      }
+
+      template <class V>
+      consteval size_t count_members_impl()
+      {
+       constexpr size_t low = count_low<V, 0, max_pure_reflection_count>();
+       if constexpr (low == 0)
+       {
+           // Every member is default-constructible. If false for n=1 that means the struct has no members.
+           if constexpr (!initializable_with_n<V, 1>) return 0;
+           constexpr size_t high = count_high<V, 1, max_pure_reflection_count>();
+           return count_between<V, 1, high>();
+       }
+       else
+       {
+           constexpr size_t high = count_high<V, low, max_pure_reflection_count>();
+           return count_between<V, low, high>();
+       }
+      }
+
+      template <class T>
+        requires(std::is_aggregate_v<std::remove_cvref_t<T>>)
+       inline constexpr auto count_members = count_members_impl<std::remove_cvref_t<T>>();
 #endif
    }
 
--- a/include/glaze/reflection/to_tuple.hpp
+++ b/include/glaze/reflection/to_tuple.hpp
@@ -173,6 +173,9 @@ namespace glz
       template <class V>
       consteval size_t count_members_impl()
       {
+       static_assert(!initializable_with_n<V, max_pure_reflection_count + 1>,
+                     "glaze: this type has more members than pure reflection supports "
+                     "(max_pure_reflection_count); provide a glz::meta<T> specialization.");
        constexpr size_t low = count_low<V, 0, max_pure_reflection_count>();
        if constexpr (low == 0)
        {
--- a/include/glaze/reflection/to_tuple.hpp
+++ b/include/glaze/reflection/to_tuple.hpp
@@ -173,6 +173,8 @@ namespace glz
       template <class V>
       consteval size_t count_members_impl()
       {
+       // Note: this assertion will not catch every violation.
+       // Example: struct with max_pure_reflection_count + 2 fields with the last one non-default constructible
        static_assert(!initializable_with_n<V, max_pure_reflection_count + 1>,
                      "glaze: this type has more members than pure reflection supports "
                      "(max_pure_reflection_count); provide a glz::meta<T> specialization.");
--- a/include/glaze/reflection/to_tuple.hpp
+++ b/include/glaze/reflection/to_tuple.hpp
@@ -99,6 +99,7 @@ namespace glz
       }(std::make_index_sequence<N>{});
 
       inline constexpr size_t max_pure_reflection_count = 128;
+      inline constexpr size_t no_proven_upper_bound = max_pure_reflection_count + 1;
 
       // Lowest n for which initializable_with_n is true
       template<class V, size_t K, size_t Max>
@@ -123,7 +124,7 @@ namespace glz
       {
           if constexpr (Low * 2 > Max)
           {
-              return Max;
+              return no_proven_upper_bound;
           }
           else if constexpr (initializable_with_n<V, Low * 2>)
           {
@@ -157,6 +158,18 @@ namespace glz
           }
       }
 
+      // Largest true value in [Low, Max], given Low is true.
+      template<class V, size_t Low, size_t High, size_t Max>
+      consteval size_t resolve_count()
+      {
+          if constexpr (High > Max && initializable_with_n<V, Max>) {
+              return Max;
+          }
+          else {
+              return count_between<V, Low, High>();
+          }
+      }
+
       template <class V>
       consteval size_t count_members_impl()
       {
@@ -166,12 +179,12 @@ namespace glz
            // Every member is default-constructible. If false for n=1 that means the struct has no members.
            if constexpr (!initializable_with_n<V, 1>) return 0;
            constexpr size_t high = count_high<V, 1, max_pure_reflection_count>();
-           return count_between<V, 1, high>();
+           return resolve_count<V, 1, high, max_pure_reflection_count>();
        }
        else
        {
            constexpr size_t high = count_high<V, low, max_pure_reflection_count>();
-           return count_between<V, low, high>();
+           return resolve_count<V, low, high, max_pure_reflection_count>();
        }
       }
 
--- a/tests/json_reflection_test/json_reflection_test.cpp
+++ b/tests/json_reflection_test/json_reflection_test.cpp
@@ -1382,4 +1382,51 @@ suite rename_tests = [] {
    };
 };
 
+struct ctor_only_member
+{
+   ctor_only_member(int v) : v{v} {}
+   int v;
+};
+
+template <>
+struct glz::meta<ctor_only_member>
+{
+   static constexpr auto value = glz::object(&ctor_only_member::v);
+};
+
+struct non_default_constructible_second
+{
+   double a;
+   ctor_only_member b;
+};
+static_assert(glz::detail::count_members<non_default_constructible_second> == 2,
+              "a non-default-constructible member at position >=2 must not zero the count");
+
+// An unbraced sub-object can absorb extra initializers
+struct elision_inner
+{
+   int a, b;
+};
+
+struct elision_outer
+{
+   elision_inner i;
+   int c;
+};
+static_assert(glz::detail::count_members<elision_outer> == 2,
+              "brace elision must not overcount");
+
+suite non_default_constructible_member_tests = [] {
+   "non-default-constructible member at position 2"_test = [] {
+      std::string buffer{};
+      expect(not glz::write_json(non_default_constructible_second{2.0, ctor_only_member{1}}, buffer));
+      expect(buffer == R"({"a":2,"b":{"v":1}})") << buffer;
+
+      non_default_constructible_second obj{0.0, ctor_only_member{0}};
+      expect(not glz::read_json(obj, buffer));
+      expect(obj.a == 2.0);
+      expect(obj.b.v == 1);
+   };
+};
+
 int main() { return 0; }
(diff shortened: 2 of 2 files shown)


wilkolbrzym-coder:
I’ll take a closer look later today around 16:00–18:00 UTC.

wilkolbrzym-coder:
Before review, could you confirm that the new static_asserts and the non_default_constructible_member_tests pass in a clean C++23 non-P2996 configuration? That’s the exact path this PR targets, so local verification there is the most useful signal.

makaimann:
Cool, no rush! The test I added passes for me locally and failed before the change. I just built with default options and didn’t specifically check it was configured for the non-P2996 path (and I’m away from my laptop currently). But since I only changed code in that path and that made tests pass, I’m quite confident.

wilkolbrzym-coder:
> Cool, no rush! The test I added passes for me locally and failed before the change. I just built with default options and didn’t specifically check it was configured for the non-P2996 path (and I’m away from my laptop currently). But since I only changed code in that path and that made tests pass, I’m quite confident.

Thanks — but “quite confident” isn’t enough here. The whole point of the PR is the C++23 non-P2996 path, so please verify locally with `-std=c++23` and without the P2996 reflection path enabled, and confirm that the new test passes in that exact configuration.

If you can’t do that before the review, that’s fine — I’ll verify it myself later today around 16:00–18:00 UTC. But we should not base the review on a default build that may have taken a different path.

makaimann:
Fair enough! I just confirmed locally that it passes with `-std=c++23`.

wilkolbrzym-coder:
> Fair enough! I just confirmed locally that it passes with `-std=c++23`.

Thanks for confirming. I managed to start a bit earlier, so I’ll do a quick logic pass first, then try some aggressive edge-case tests before writing the review. Expect my notes between 16:00 and 17:00 UTC.

wilkolbrzym-coder:
I’m still running additional targeted tests. There is at least one confirmed issue, so this PR will not pass review in its current state. Full details once testing finishes.

wilkolbrzym-coder (changes requested):
## Review – PR #2817

**Verdict: Request changes. This PR cannot be merged in its current state.**

The new galloping + binary search fixes the original `non_default_constructible_second` case, but it introduces correctness bugs at the algorithm boundaries and mis-handles array members.

I confirmed these with a dedicated repro test file. Results:

| Case | Expected | Actual |
|---|---:|---:|
| Aggregate with exactly 128 default-constructible members | 128 | **127** |
| Aggregate with `int arr[3]` as first member | 2 | **4** |
| Aggregate with `int arr[3]` as second member | 2 | **4** |
| Aggregate with 129 default-constructible members | 129 | **127** |

### 1. Off-by-one at `max_pure_reflection_count`

`count_high()` returns `Max` whenever `Low * 2 > Max`, but it does **not prove** that `initializable_with_n<V, Max>` is false. `count_between()` requires `High` to be a false upper bound.

For a type with exactly 128 members, `Max` is still true, so the search returns 127 instead of 128.

```cpp
struct big_128 { /* 128 int members */ };
static_assert(glz::detail::count_members<big_128> == 128); // fails: 127
```

### 2. Array members are counted as flattened initializer slots

`initializable_with_n` tests how many brace initializers the aggregate can accept. An array member absorbs N initializers, so:

```cpp
struct array_first {
   int arr[3];
   int b;
};
```

is counted as `4`, not `2`, because `V{any_t, any_t, any_t, any_t}` is valid. The same happens when the array is the second member.

This violates `count_members` semantics: an array is one aggregate member, not N members.

### 3. Beyond-limit case is silently wrong

For 129 members, the result is `127`, combining the boundary bug with incorrect capping. There is no diagnostic that the type exceeds `max_pure_reflection_count`.

### Required changes

- Ensure `count_high()` returns a genuinely false upper bound, or handles `Max + 1`.
- Explicitly detect and report types exceeding `max_pure_reflection_count`.
- Handle array members as single members in the C++23 counting path, or add a clear constraint if arrays are unsupported.

The original reported case passes, but the algorithm is not generally correct yet.

Full repro test file attached.

cc @stephenberry — review complete; this should not be merged as-is.

[count_members_edge_test.cpp](https://github.com/user-attachments/files/31270641/count_members_edge_test.cpp)

annihilatorq:
Hey, thanks for the PR!
The approach itself looks right to me, hope you'll get it over the line. I'd just wait for Stephen's take before reworking anything - he has the final word here, and compile-time reflection is one of the most sensitive areas of Glaze. Would be a shame to spend time on changes he might want done differently.

To be clear, I haven't reviewed the diff or checked the other findings - the only thing I looked into is the array flattening, and it's not new, the old counter also counted through brace-init, so `int arr[3]` was 3 slots before this PR too. I wouldn't consider this a blocker. If a second pair of eyes would help at some point, feel free to ping me.

wilkolbrzym-coder:
> Hey, thanks for the PR! The approach itself looks right to me, hope you'll get it over the line. I'd just wait for Stephen's take before reworking anything - he has the final word here, and compile-time reflection is one of the most sensitive areas of Glaze. Would be a shame to spend time on changes he might want done differently.
> 
> To be clear, I haven't reviewed the diff or checked the other findings - the only thing I looked into is the array flattening, and it's not new, the old counter also counted through brace-init, so `int arr[3]` was 3 slots before this PR too. I wouldn't consider this a blocker. If a second pair of eyes would help at some point, feel free to ping me.

@annihilatorq Thanks for the context. I agree array flattening may predate this PR; I’m not treating it as the main blocker.

The hard blocker is the boundary result:

- `count_members<big_128>` returns **127**
- `count_members<big_129>` returns **127**

The 128-member case is an off-by-one in the new `count_high` / `count_between` contract. `count_high` can return `Max` as if it were a proven false upper bound, while it is actually true.

That is introduced by this implementation, not inherited from the old counter. So I think it needs fixing before merge.

If Stephen wants the 129 case treated as a soft limit, that is a separate design decision. But 128 members should count as 128.

@makaimann I can open a minimal failing test against your branch if that helps.

makaimann:
Thanks for the great review! I pushed up some new tests and fixes for the off-by-one issue.

> If Stephen wants the 129 case treated as a soft limit, that is a separate design decision. But 128 members should count as 128.

I'm definitely interested in this design decision. I made an attempt to detect the `>max_pure_reflection_count` case but I couldn't think of anything better than 4f26fecb43f9cb45a6ad0ead49977e047798a33b. See the comment in that commit for the limitations.

### Arrays
I didn't address the C-style arrays issue. Partly because it was pre-existing in the repo, but I also just couldn't come up with a solution that worked. I'm open to ideas there though!

wilkolbrzym-coder:
I’ll re-check the updated branch later today, around 16:00–18:00 UTC, possibly earlier if I get to it. I have two possible approaches in mind, but I need to test them before proposing anything.

stephenberry:
I've just had a chance to glance at this. I'm very cautious about adding complexity to the C++23 reflection logic, because I don't want to slow down compile time and because now we already have C++26 reflection support that solves this problem in Glaze. So, this isn't a generally useful improvement and I'd rather encourage C++26 adoption over a complex C++23 implementation.

@makaimann, I appreciate your work on this, but could you share a use case for why this feature would be necessary for your work to succeed? Or, are you thinking of this as a more general improvement?

The danger I see is that if we start accepting constructors with C++23 pure reflection then we need to start handling all the corner cases of constructors in C++ and this seems like there will be even more surprises than the simple rule that structs with constructors won't work in C++23 reflection.

wilkolbrzym-coder:
@stephenberry I agree with the caution on expanding C++23 reflection complexity. But I’d frame the issue slightly differently: this is not about supporting constructors broadly. The failing type is still an aggregate and satisfies `std::is_aggregate_v`; it only has a member that isn’t default-constructible. Counting it as zero fields is surprising for a documented aggregate path.

I’ll have concrete compile-time numbers and updated boundary-test results by 18:00 UTC.

My current position: C++26 reflection is the right long-term answer, but that doesn’t mean the C++23 path should silently produce wrong results for an aggregate that is otherwise legal. The fix should be narrower than the current galloping/binary search if it stays in C++23 — specifically avoid broad constructor handling and avoid plausible-but-wrong results at the 128/129 boundary.

wilkolbrzym-coder:
I’m starting the verification pass on the updated branch now. I’ll measure compile-time impact against `main` and re-run the boundary cases, then post the results.

wilkolbrzym-coder:
I’m extending the review pass. The updated branch needs a few more targeted compile-time checks than expected, so I’ll post the final review around 20:00 UTC.

wilkolbrzym-coder (commented):
## Review update – PR #2817

I completed the full verification pass. The results below compare the current PR state with the sentinel-based fix applied.

---

### Compile-time impact vs `main`

20 runs, 2 warm-ups, `clang++ -fsyntax-only`, median reported.

| Members | main before fix | PR before fix | main after fix | PR after fix |
|---:|---:|---:|---:|---:|
| 0 | 0.407 s | 0.413 s | 0.408 s | 0.392 s |
| 1 | 0.400 s | 0.395 s | 0.393 s | 0.391 s |
| 2 | 0.397 s | 0.401 s | 0.401 s | 0.391 s |
| 8 | 0.396 s | 0.392 s | 0.397 s | 0.401 s |
| 64 | 0.444 s | 0.412 s | 0.425 s | 0.401 s |
| 128 | 0.489 s | 0.413 s | 0.484 s | 0.401 s |

Compile time after the fix remains comparable. Differences on small structs are within noise, and the 128-member case is still clearly faster on this setup.

---

### Boundary and over-limit results

| Case | Before fix | After fix |
|---|---|---|
| `big_128` | correct: 128 | correct: 128 |
| `big_129` | limit `static_assert` | limit `static_assert` |
| `silent_over_limit` | **bug: silently returned 128** | **correct: new `static_assert`** |
| `silent_over_limit_second` | returned 130 | limit `static_assert` |
| `very_deep_over_limit` | no guaranteed detection | limit `static_assert` |
| `array_first` | flattened count: 4 | unchanged: 4 |

The critical change is `silent_over_limit`.

Without the fix, `count_members` returned a plausible-but-wrong `128`.  
With the fix, `count_low` returns the sentinel `129`, and `count_members_impl` terminates with a clear diagnostic:

```
glaze: unable to count members for this aggregate within
max_pure_reflection_count; provide a glz::meta<T> specialization.
```

---

### Required fix

```cpp
template<class V, size_t K, size_t Max>
consteval size_t count_low()
{
   if constexpr (K > Max) {
       return no_proven_upper_bound; // was: return Max
   }
   else if constexpr (initializable_with_n<V, K>) {
       return K;
   }
   else {
       return count_low<V, K + 1, Max>();
   }
}
```

```cpp
constexpr size_t low = count_low<V, 0, max_pure_reflection_count>();

if constexpr (low == no_proven_upper_bound) {
   static_assert(false,
                 "glaze: unable to count members for this aggregate within "
                 "max_pure_reflection_count; provide a glz::meta<T> specialization.");
}
```

---

### Full repro, measurement script, and updated implementation

The complete updated `to_tuple.hpp` is available here:

https://gist.github.com/wilkolbrzym-coder/507450b884f59406d854def8dc4ccea0

The measurement script, positive boundary tests, and negative diagnostic cases are attached directly to this review.

---

### Verdict

With the sentinel fix, I no longer see a correctness bug in the C++23 counting path.

The remaining decision is architectural: whether this added C++23 complexity is justified when C++26 reflection already solves the problem.

I’ll defer that to @stephenberry.

If the PR is kept, the sentinel fix and the diagnostic cases should be included.

cc @stephenberry 

[count_members_compile_time.sh](https://github.com/user-attachments/files/31317612/count_members_compile_time.sh)
[count_members_edge_test.cpp](https://github.com/user-attachments/files/31317609/count_members_edge_test.cpp)
[silent_over_limit.cpp](https://github.com/user-attachments/files/31317610/silent_over_limit.cpp)
[silent_over_limit_second.cpp](https://github.com/user-attachments/files/31317613/silent_over_limit_second.cpp)
[very_deep_over_limit.cpp](https://github.com/user-attachments/files/31317611/very_deep_over_limit.cpp)

wilkolbrzym-coder:
Why the 128-member case got faster: the old implementation did a linear template recursion, one `initializable_with_n` probe per member. The new galloping + binary search reduces that to a small logarithmic number of probes. For small aggregates the difference is noise; at the upper boundary it becomes measurable.

makaimann:
Sorry for the delayed response -- had a busy last 24 hours.

Thanks all for the discussion, and for the explanation @stephenberry. @wilkolbrzym-coder is correct that this is still about aggregate structs. That being said, I'll explain my overarching goal in case there's a better approach for that.

I'm hoping to use `glaze` in a codebase that I can't upgrade to C++26 yet. It has some types that I would like to have json serialization/deserialization/schema generation for. Some of those types have fields of a type that has no default constructor. For those field types,  I can use `glz::meta`, but my goal is for it to "compose" automatically if used as a field in an aggregate struct. Without this fix, the json write could just return `{}`.

I'm not sure that explanation made a lot of sense,  so I'll give a short example using the types introduced in the tests of this PR to illustrate:

```cpp
struct ctor_only_member {
    ctor_only_member(int v) : v{v} {}
    int v;
};

template<>
struct glz::meta<ctor_only_member>
{
    static constexpr auto value = glz::object(&ctor_only_member::v);
}

struct non_default_constructible_second {
    double a;
    ctor_only_member b;
}

suite non_default_constructible_member_tests = [] {
    "non-default-constructible member at position 2 writes as {}"_test = [] {
        std::string buffer{};
        expect(not glz::write_json(non_default_constructible_second{2.0, ctor_only_member{1}}, buffer));
        expect(buffer != "{}") << buffer; // <-- This fails before the count_members fix.
    };
};
```

I can add that test to the PR if people think that would help explain the purpose. Let me know if I can clarify further!

It’s not my preference, but we could also wait to integrate glaze until we can update to C++26 of course.

wilkolbrzym-coder:
@makaimann Please update the test to assert the exact JSON:

expect(buffer == R"({"a":2,"b":{"v":1}})") << buffer;

Then run that test locally with -std=c++23 together with the updated count_members changes, and confirm:

- the new test passes with the sentinel fix,
- the existing reflection tests still pass,
- the boundary cases from the review pass.

A schema-generation test can be a follow-up if that is part of your goal.

The final decision remains with @stephenberry.

makaimann:
Sounds good, just added the test! For some reason I get a 404 error if I click on the `silent_over_limit.cpp` link above. Could you paste the code? I'd like to confirm that still behaves as expected.

wilkolbrzym-coder:
> Sounds good, just added the test! For some reason I get a 404 error if I click on the `silent_over_limit.cpp` link above. Could you paste the code? I'd like to confirm that still behaves as expected.

The `silent_over_limit.cpp` file is the one that should fail with the sentinel fix applied.

```cpp
#include "glaze/reflection/to_tuple.hpp"

struct constructor_only_member
{
   constructor_only_member(int value) : value{value} {}
   int value;
};

#define MEMBERS_8(prefix) int prefix##0, prefix##1, prefix##2, prefix##3, prefix##4, prefix##5, prefix##6, prefix##7;
#define MEMBERS_64(prefix) \
   MEMBERS_8(prefix##a) MEMBERS_8(prefix##b) MEMBERS_8(prefix##c) MEMBERS_8(prefix##d) \
   MEMBERS_8(prefix##e) MEMBERS_8(prefix##f) MEMBERS_8(prefix##g) MEMBERS_8(prefix##h)

struct silent_over_limit
{
   int extra;
   MEMBERS_64(x)
   MEMBERS_64(y)
   constructor_only_member last;
};

static_assert(glz::detail::count_members<silent_over_limit> == 128);
```

Expected with the fix: compilation fails with:

```text
glaze: unable to count members for this aggregate within
max_pure_reflection_count; provide a glz::meta<T> specialization.
```

If it compiles and `count_members` returns `128`, the sentinel change is not present in `count_low`.

---

One more thing: I see you are working over the weekend. Take care of yourself — burnout is real. Even Stephen usually skips weekends, though I think he showed up briefly not long ago.

I personally avoid burnout by not specializing in only one thing ;)

wilkolbrzym-coder:
Also: I may be less available tomorrow - I’m making a full attempt at the ARC-AGI 3 benchmark. AGI.

makaimann:
> The silent_over_limit.cpp file is the one that should fail with the sentinel fix applied.

Ahh right, thanks! I just confirmed locally that this still fails with the expected error message:
`… this type has more members than pure reflection supports…`

makaimann:
Hey @wilkolbrzym-coder and @stephenberry, I just wanted to check in on this PR. I noticed there's a couple failing tests which I can look into, but I thought I'd wait until there's a consensus decision. Let me know if there's anything I can answer or investigate more to help with the decision?

wilkolbrzym-coder:
> Hey @wilkolbrzym-coder and @stephenberry, I just wanted to check in on this PR. I noticed there's a couple failing tests which I can look into, but I thought I'd wait until there's a consensus decision. Let me know if there's anything I can answer or investigate more to help with the decision?

Let's wait for Stephen to see the general concept.

PR #2818, opened 2026-08-21
Gate glaze::glaze's MSVC-only interface flags per consumer

### LINKED ISSUE #2814: INTERFACE_LINK_OPTIONS (/LTCG, /INCREMENTAL:NO) and /Zc:preprocessor leak to consumers using a different compiler frontend than the one that built glaze
Similar in shape to #2072 (fixed in #2076): Glaze's root CMakeLists.txt bakes MSVC-specific compile and link flags into `glaze_glaze`'s `INTERFACE` properties, basing the decision on the compiler that builds glaze itself rather than the compiler of whatever target consumes it later.

### Root cause

```cmake
if (MSVC)
  string(REGEX MATCH "\/cl(.exe)?$" matched_cl ${CMAKE_CXX_COMPILER})
  if (matched_cl)
    target_compile_options(glaze_glaze INTERFACE "/Zc:preprocessor" /permissive- /Zc:lambda)
    if(PROJECT_IS_TOP_LEVEL)
      target_compile_options(glaze_glaze INTERFACE
        $<$<CONFIG:Release>:/GL>
        $<$<CONFIG:MinSizeRel>:/GL>)
      target_link_options(glaze_glaze INTERFACE
        $<$<CONFIG:Release>:/LTCG /INCREMENTAL:NO>
        $<$<CONFIG:MinSizeRel>:/LTCG /INCREMENTAL:NO>)
    endif()
  endif()
endif()
```

CMake evaluates `INTERFACE_COMPILE_OPTIONS` and `INTERFACE_LINK_OPTIONS` once at export time rather than per consumer, so this setup silently propagates flags that only make sense for glaze's own build environment onto downstream targets that may use a completely different toolchain.

### Cross-compilation risk

More generally, this pattern breaks any cross-compiling setup where the toolchain that builds glaze differs from the toolchain that builds the consuming project. Package managers that resolve each dependency's toolchain independently (vcpkg is one example, not the only one) hit this risk particularly reliably, since a host compiler and a target compiler can reasonably differ even on a single machine. The same failure mode applies to cross-compilation generally, with or without a package manager involved.

### Reproduction

Any project that consumes glaze through a build where the compiler that builds glaze differs from the compiler building the consuming project hits this. Concretely, on Windows, this reproduces when a project consumes glaze through a package manager (vcpkg, for example) while building itself with `clang++` (the LLVM toolset Visual Studio bundles) targeting the MSVC ABI and `lld-link`, rather than real `cl.exe` or `clang-cl.exe`. In this setup:

- `CMAKE_CXX_COMPILER_ID` is `Clang`.
- `CMAKE_CXX_SIMULATE_ID` is `MSVC` (so CMake's `MSVC` variable evaluates `TRUE`).
- The compiler's driver frontend is GNU-style (`-D`, `-I`, not `/D`, `/I`).
- The linker is still MSVC-compatible (`lld-link`).

A package manager builds the glaze port with its own toolchain resolution (vcpkg typically uses real `cl.exe` for this), so `matched_cl` evaluates true for glaze's own build, and glaze exports these flags as `INTERFACE`. When our project, a different, separately-configured `clang++` build, links `glaze::glaze`, it inherits `/Zc:preprocessor /permissive- /Zc:lambda` as raw compile options and `/LTCG /INCREMENTAL:NO` as raw link options. Our Clang silently accepts (or warns about) the compile-time flags, but the link-time ones break the build outright:

```
clang++: error: no such file or directory: '/LTCG'
clang++: error: no such file or directory: '/INCREMENTAL:NO'
```

Our build's linker rule wraps only the first token after `-Xlinker` when forwarding to `lld-link`, so `/LTCG` and `/INCREMENTAL:NO` fall through as bare argv tokens Clang's GNU-style driver does not recognize.

### Suggested fix

CMake 3.30 added `$<CXX_COMPILER_FRONTEND_VARIANT:MSVC>`, a generator expression that re-evaluates against each consuming target's own frontend variant at that target's own generate time, rather than freezing whatever `CMAKE_CXX_COMPILER_FRONTEND_VARIANT` happened to be when glaze itself configured. This correctly includes both real `cl.exe` and `clang-cl` (both report `MSVC` as their frontend variant) and correctly excludes a GNU-frontend `clang++` targeting the MSVC ABI:

```cmake
target_compile_options(glaze_glaze INTERFACE
  $<$<CXX_COMPILER_FRONTEND_VARIANT:MSVC>:/Zc:preprocessor /permissive- /Zc:lambda>)

target_link_options(glaze_glaze INTERFACE
  $<$<AND:$<CXX_COMPILER_FRONTEND_VARIANT:MSVC>,$<CONFIG:Release>>:/LTCG /INCREMENTAL:NO>
  $<$<AND:$<CXX_COMPILER_FRONTEND_VARIANT:MSVC>,$<CONFIG:MinSizeRel>>:/LTCG /INCREMENTAL:NO>)
```

Happy to open a PR with this change if a specific fix direction is preferred.

Yaraslaut:
Closes https://github.com/stephenberry/glaze/issues/2814 

Fix detection of clang-cl in CMake and add a CI job to test direct usage of clang-cl.

### DIFF
--- a/.github/workflows/clang-cl.yml
+++ b/.github/workflows/clang-cl.yml
@@ -49,3 +49,26 @@ jobs:
     - name: Test
       working-directory: build
       run: ctest --build-config ${{env.BUILD_TYPE}} --timeout 300
+
+  build-direct-clang-cl:
+    runs-on: windows-2025-vs2026
+    timeout-minutes: 25
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - uses: seanmiddleditch/gha-setup-ninja@v6
+
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: x64
+
+    - name: Configure CMake
+      run: cmake -B ${{github.workspace}}/build-direct-clang-cl -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_CXX_STANDARD=23
+
+    - name: Build
+      run: cmake --build build-direct-clang-cl --parallel
+
+    - name: Test
+      working-directory: build-direct-clang-cl
+      run: ctest --build-config ${{env.BUILD_TYPE}} --timeout 300
--- a/.github/workflows/clang-cl.yml
+++ b/.github/workflows/clang-cl.yml
@@ -57,18 +57,19 @@ jobs:
     steps:
     - uses: actions/checkout@v6
 
-    - uses: seanmiddleditch/gha-setup-ninja@v6
-
-    - uses: ilammy/msvc-dev-cmd@v1
-      with:
-        arch: x64
-
-    - name: Configure CMake
-      run: cmake -B ${{github.workspace}}/build-direct-clang-cl -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DCMAKE_CXX_STANDARD=23
-
-    - name: Build
-      run: cmake --build build-direct-clang-cl --parallel
-
-    - name: Test
-      working-directory: build-direct-clang-cl
-      run: ctest --build-config ${{env.BUILD_TYPE}} --timeout 300
+    # Ninja and vswhere are both preinstalled on this runner image, and vcvarsall is
+    # what hands clang-cl the MSVC headers, libraries and linker, so nothing here has
+    # to be fetched from outside the image. That keeps the job's supply chain to the
+    # runner itself rather than a third-party action fetching a toolchain at run time.
+    # The developer environment only lives as long as the step that calls vcvarsall,
+    # so configure, build and test all run inside this one step.
+    - name: Configure, build and test with clang-cl
+      shell: cmd
+      run: |
+        "%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath > "%RUNNER_TEMP%\vsinstall.txt" || exit /b 1
+        set /p VSINSTALL=<"%RUNNER_TEMP%\vsinstall.txt"
+        if not defined VSINSTALL (echo No Visual Studio installation with the C++ toolset was found & exit /b 1)
+        call "%VSINSTALL%\VC\Auxiliary\Build\vcvarsall.bat" x64 || exit /b 1
+        cmake -B build-direct-clang-cl -G Ninja -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_BUILD_TYPE=%BUILD_TYPE% -DCMAKE_CXX_STANDARD=23 || exit /b 1
+        cmake --build build-direct-clang-cl --parallel || exit /b 1
+        ctest --test-dir build-direct-clang-cl --build-config %BUILD_TYPE% --timeout 300 || exit /b 1
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,9 +31,13 @@ if (MSVC)
   # consumer only gets these flags if *it* is actually compiling with real MSVC. (Note
   # $<CXX_COMPILER_FRONTEND_VARIANT:MSVC> would not work here: it evaluates to true for
   # clang-cl too, since matching MSVC's driver interface is the entire premise of clang-cl.)
+  # The compile options narrow further to CXX, for the same reason the else() branch below
+  # guards on COMPILE_LANGUAGE: a consumer that compiles C, ISPC or CUDA alongside its C++
+  # should not have /Zc:lambda handed to those compilers. Link options have no per-language
+  # form, so they stay on CXX_COMPILER_ID alone.
   target_compile_options(glaze_glaze INTERFACE
     # for a C++ standards compliant preprocessor, not needed for clang-cl
-    $<$<CXX_COMPILER_ID:MSVC>:/Zc:preprocessor /permissive- /Zc:lambda>)
+    $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/Zc:preprocessor /permissive- /Zc:lambda>)
 
   # Whole-program optimization changes codegen quality, never correctness, and the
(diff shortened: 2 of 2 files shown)


stephenberry:
Thanks for taking this on, and for adding the direct clang-cl job — that Ninja path is welcome coverage.

A few things before this goes in:

1. I don't think clang-cl was actually being misdetected. The regex `\/cl(.exe)?$` requires a `/` immediately before `cl`, so `.../clang-cl.exe` never matched it — our `-T ClangCL` CI confirms the path CMake sees is `C:/Program Files/Microsoft Visual Studio/18/Enterprise/VC/Tools/Llvm/x64/bin/clang-cl.exe`. The `CMAKE_CXX_COMPILER_ID` check is still an improvement (the path regex misses real `cl.exe` behind a compiler launcher, and the unquoted `${CMAKE_CXX_COMPILER}` expansion is fragile with spaces in paths), so I'd like to take it — could you reword the title/description as a robustness cleanup rather than a misdetection fix?

2. Could you drop `Closes #2814`? That report is about the flags being baked into `INTERFACE_COMPILE_OPTIONS` once, at glaze's own configure time, and then leaking to a consumer using a different frontend (a GNU-driver `clang++` targeting the MSVC ABI, consuming a glaze that vcpkg built with real `cl.exe`). That still happens after this change, so the issue should stay open.

3. Minor: could you pin `seanmiddleditch/gha-setup-ninja@master` to a tag or commit SHA? These are the first third-party actions in our workflows and I'd rather not track a moving branch.

For #2814 the fix is a per-consumer generator expression, e.g. `$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/Zc:preprocessor>`, which re-evaluates against the consuming target and is false for clang-cl, so current behavior is preserved. The `else()` branch has the mirror problem (`-Wno-missing-braces` is chosen by glaze's own compiler). That can be a follow-up — no need to expand this PR.

Yaraslaut:
Updated to actually fix this: the MSVC-only flags on `glaze::glaze` are now wrapped in `$<CXX_COMPILER_ID:MSVC>` generator expressions, so they're evaluated per consumer at generate time instead of being baked in once at glaze's own configure time. Verified with a separate consumer project (installed glaze, then configured a consumer with a different compiler) that a clang-cl consumer no longer inherits the MSVC-only flags, while a real-MSVC consumer still gets them.

Also added a CI job that configures/builds glaze with clang-cl directly via Ninja (`-G Ninja`, `CMAKE_CXX_COMPILER=clang-cl`), since the existing job only exercises the `-T ClangCL` VS toolset path.

stephenberry:
Pushed a commit to this branch rather than sending you back around again.

**The CI job now builds from the runner image alone.** `windows-2025-vs2026` already ships Ninja 1.13.2 and vswhere, so `gha-setup-ninja` was downloading 1.12.1 to replace something newer, and `msvc-dev-cmd` was using that same vswhere to locate `vcvarsall.bat` and call it. Inline it's four lines of batch. The cost is per-step timing in the log — the environment vcvarsall sets only survives the step that calls it, so configure/build/test share one step.

This isn't about those two actions specifically. It's that this workflow runs on pushes to `main`, where the job token is read-write, and I'd rather that job not execute third-party JavaScript pinned to a movable tag.

**Narrowed the compile options to CXX** — `$<COMPILE_LANG_AND_ID:CXX,MSVC>`. They're per-consumer now, but a consumer compiling C or ISPC alongside its C++ would still have had `/Zc:lambda` handed to those compilers, the same leak the `else()` branch already guards against. Link options have no per-language form, so they stay as you wrote them.

One more: please put `Closes #2814` back. I asked you to drop it against the first revision, which only changed the detection — this version genuinely fixes the reported failure, since a GNU-driver `clang++` consumer no longer inherits any of these flags. What's left is only the `else()` mirror, which is a different scenario and can be its own PR.

Thanks for turning this around so fast.

wilkolbrzym-coder (commented):
## Review – PR #2818

I came back to check this because I couldn’t sleep — something about this PR was passing too smoothly. It didn’t hold up.

The latest commit improves two things correctly:

- The direct `clang-cl` workflow no longer depends on runtime third-party actions.
- Compile options were narrowed to `$<COMPILE_LANG_AND_ID:CXX,MSVC>`.

But the blocking CMake issue is still there.

### Multiple flags in single generator expressions

```cmake
$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/Zc:preprocessor /permissive- /Zc:lambda>
```

and:

```cmake
$<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:Release>>:/LTCG /INCREMENTAL:NO>
```

These are not guaranteed to become separate `argv` tokens. A green CI run does not prove otherwise; it only means the current driver/response-file path happens to split them.

Split them:

```cmake
target_compile_options(glaze_glaze INTERFACE
  $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/Zc:preprocessor>
  $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/permissive->
  $<$<COMPILE_LANG_AND_ID:CXX,MSVC>:/Zc:lambda>
)
```

```cmake
target_link_options(glaze_glaze INTERFACE
  $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:Release>>:/LTCG>
  $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:Release>>:/INCREMENTAL:NO>
  $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:MinSizeRel>>:/LTCG>
  $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<CONFIG:MinSizeRel>>:/INCREMENTAL:NO>
)
```

After that, verify with `ninja -t commands` or `cmake --build ... --verbose` that every flag is a separate token.

No other blockers.

cc @stephenberry

Yaraslaut:
thanks for the feedback, updated as suggested @wilkolbrzym-coder

wilkolbrzym-coder:
> thanks for the feedback, updated as suggested @wilkolbrzym-coder 

I'll check around 15-16 utc.

wilkolbrzym-coder (approved):
## Review update – PR #2818

Confirmed fixed:

- compile flags are split into separate generator expressions
- link flags are split too
- direct clang-cl CI job uses only the runner image
- consumer-side MSVC flag leak from #2814 is addressed

Non-blocking: PR title still says “misdetected as MSVC”; current change is really consumer-option gating. The `else()` mirror case can stay as a follow-up.

CI green, no conflicts.

LGTM. Approve.

cc @stephenberry

PR #2819, opened 2026-08-21
Declare stream_request_params_v2 callbacks in the order callers use them

stephenberry:
## The problem

`stream_request_params_v2` declares its four callbacks as `on_connect`, `on_disconnect`, `on_data`, `on_error`. That is not the order a caller reaches for them: `on_data` and `on_error` are the two every streaming call supplies, and `on_connect`/`on_disconnect` are the optional extras.

Designators have to appear in declaration order. gcc and MSVC enforce the ISO rule as a hard error; clang accepts any order as an extension and stays silent. So the spelling that comes naturally,

```cpp
client.stream_request_v2({.url = url, .on_data = f, .on_error = g, .on_disconnect = h});
```

compiles locally on clang and then fails the build on gcc and MSVC. Because it is an error rather than a warning, the mistake reaches CI instead of the editor. #2782 hit exactly this: one designator out of order took out twelve jobs — every gcc, MSVC and sanitizer build — while every clang job stayed green.

## The fix

Declare the callbacks as `on_data`, `on_error`, `on_connect`, `on_disconnect`, so the obvious spelling is the correct one. The non-callback fields keep their positions.

This also lines the struct back up with two things it had drifted from:

- `perform_stream_request` has always taken the handlers in that order, and both `stream_request` and `stream_request_v2` forward them that way.
- The deprecated `stream_request_params` put `on_data` and `on_error` immediately after `url`. The v2 struct is what moved them behind the optional pair.

## Compatibility

This is a source break for designated initializers that named `on_connect` or `on_disconnect` before `on_data` or `on_error`. It breaks loudly, at compile time, and only on the compilers that enforce the rule; the fix is to reorder the designators. Field types, names, defaults and semantics are all unchanged, so clang builds are unaffected in either direction, and positional aggregate initialization of a twelve-field struct with four distinct `std::function` signatures does not silently change meaning.

The five call sites in `http_chunked_test` and the field list in `docs/networking/http-client.md` are updated. The remaining designated-initializer call sites in `http_client_test` use the deprecated v1 `stream_request_params`, which is untouched.

## Testing

`http_chunked_test` 59/59, `http_client_test` 39/39, `http_content_length_test` 24/24, `http_client_pool_test` 14/14. Every TU in the repo that includes `http_client.hpp` syntax-checks clean under `-Werror=reorder-init-list` (clang's spelling of the rule gcc and MSVC enforce), with the SSL test checked under `-DGLZ_ENABLE_SSL`. A scratch check confirms the natural order now compiles under that flag and the previous order is what it rejects.

## Note on #2782

#2782 adds a `stream_request_v2` call site written in the current declaration order. Whichever of the two merges second needs its designators reordered — a four-line change in one test.

### DIFF
--- a/docs/networking/http-client.md
+++ b/docs/networking/http-client.md
@@ -254,10 +254,10 @@ struct stream_request_params_v2 {
     size_t max_buffer_size{1024 * 1024};
     std::string body;
     glz::http_headers headers;
-    http_connect_handler on_connect;
-    http_disconnect_handler on_disconnect;
     http_data_handler on_data;
     http_error_handler on_error;
+    http_connect_handler on_connect;
+    http_disconnect_handler on_disconnect;
     std::function<bool(int)> status_is_error{[](int status){ return status >= 400; }};
 };
 ```
@@ -269,10 +269,10 @@ struct stream_request_params_v2 {
 -   `max_buffer_size`: Larger buffer can decrease dropouts and increase throughput at cost of memory usage. (default is 1 MiB)
 -   `body`: The HTTP Body to send.
 -   `headers`: The HTTP headers to send.
--   `on_connect`: A callback that's called when the connection is established and the headers are received.
--   `on_disconnect`: A callback that's called when the connection is closed.
 -   `on_data`: A callback that's called when data is received.
 -   `on_error`: A callback that's called when an error occurs.
+-   `on_connect`: A callback that's called when the connection is established and the headers are received.
+-   `on_disconnect`: A callback that's called when the connection is closed.
 -   `status_is_error`: Optional predicate to decide whether a status code should trigger `on_error` (defaults to checking for codes ≥ 400).
 
 To override the default behaviour you can supply a predicate:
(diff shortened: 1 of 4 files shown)


wilkolbrzym-coder:
designated-initializer order is exactly the kind of issue that slips through clang and then explodes across the GCC/MSVC matrix.

Let me know when this is ready for review.

wilkolbrzym-coder:
## Review – PR #2819

The change is correct and fixes a real compiler portability issue.

Earlier I said I’d wait until you marked it ready for review. Since I’m stepping away shortly and CI is already green, I’m posting this now instead of delaying further.

A few observations:

1. The field reorder correctly aligns designated-initializer order with the natural call-site order. Clang’s permissiveness was masking a GCC/MSVC hard error.

2. Technically, this is still an aggregate layout/order change. Any user performing full positional aggregate initialization of `stream_request_params_v2` could be affected. In practice the four callback types have distinct `std::function` signatures, so most mistakes will fail to compile rather than compile silently. But it should still be called out in release notes as a source-order change.

3. Documentation and test call sites are updated consistently.

4. The test coverage is sufficient: all TUs that include `http_client.hpp` are syntax-checked, which catches exactly this class of ordering bug.

No blocking concerns.

Approve.

cc @stephenberry

PR #2821, opened 2026-08-21
Give every workflow an explicit token scope

stephenberry:
Step one of restricting what a CI job can do with its `GITHUB_TOKEN`.

The repository default is currently **read-write**, and only `docs.yml` declared a scope of its own. So all 30 other workflows were handing their jobs a token that can push to `main` — including jobs that run third-party actions, and including jobs triggered by pushes to `main` and `feature/*`, where the token is not downgraded the way it is for fork pull requests.

Exactly two workflows need to write anything:

| Workflow | Scope | Why |
|---|---|---|
| `clang-format.yml` | `contents: write` | commits formatting fixes back to the branch |
| `clang-tidy.yml` | `contents: write` | commits include-cleaner fixes back to the branch |
| `docs.yml` | unchanged | already declared `pages: write` + `id-token: write` |
| everything else (28) | `contents: read` | only reads the checkout |

`upload-artifact` needs no extra scope — it authenticates against the artifact backend with the runtime token, not `GITHUB_TOKEN`.

### Why per-workflow rather than just the repository setting

Flipping the repository default alone would achieve the same runtime effect in fewer keystrokes, but it puts the security posture in a settings page instead of the repository. Declaring it in each file means the workflow says what it is allowed to do, it shows up in review when a workflow is added or changed, and it survives someone flipping the setting back.

### Ordering

**This PR is a no-op by itself** — the default is still read-write, so nothing loses access. It has to land first, so that flipping the repository default to read-only afterwards doesn't break the two workflows that commit.

Once this merges: Settings → Actions → General → Workflow permissions → **Read repository contents and packages permissions**. Worth unchecking "Allow GitHub Actions to create and approve pull requests" at the same time; nothing here uses it.

CI on this PR exercises the change directly, since editing `.github/workflows/**` triggers the full suite.

### DIFF
--- a/.github/workflows/armv7.yml
+++ b/.github/workflows/armv7.yml
@@ -16,6 +16,9 @@ on:
       - 'docs/**'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 concurrency:
   # Supersede an in-flight run when a branch is pushed again, so a fixup commit
   # does not leave the previous run competing for the same runners. Pushes to main
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -16,6 +16,9 @@ on:
       - 'docs/**'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 concurrency:
   # Supersede an in-flight run when a branch is pushed again, so a fixup commit
   # does not leave the previous run competing for the same runners. Pushes to main
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -9,6 +9,9 @@ on:
       - 'docs/**'
   workflow_dispatch:
 
+permissions:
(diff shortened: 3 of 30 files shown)


CramBL:
Are you familiar with https://github.com/zizmorcore/zizmor? It's good for hardening your actions configuration

wilkolbrzym-coder (commented):
Post-merge review – PR #2821

I went through this after merge. The change is correct and closes a real token-permission gap.

No security regression found.

Two future-proofing recommendations:

1. If a workflow ever starts publishing releases, GitHub Pages, or packages, it must explicitly add the required write scope at that point.
2. A small contributor checklist item would help: every new workflow under `.github/workflows` must declare an explicit `permissions` block. Otherwise the default token can creep back in.

LGTM.

cc @stephenberry

PR #2822, opened 2026-08-24
guard yaml pair flow-mapping key peek against end of input

uwezkhan:
The YAML std::pair reader parses the flow form {key: value}. After skipping the opening { and inline whitespace, its empty-pair check `if (it != end && *it == '}')` doesn't return when it == end, so a buffer that stops right after the { falls through and peeks *it one byte past the end. Reading a std::pair<std::string, V> from the single byte "{" over a non-null-terminated buffer (null_terminated=false, a string_view or span) is an out-of-bounds read.

Add an it == end check after the empty-pair test so the reader returns unexpected_end instead. The sibling flow-mapping and map readers already loop on while (it != end) and the pair reader's own anchor branch guards it == end after the name, so this was the one flow site missing the bound. The check belongs in the reader because that is where the cursor position is known; valid input parses the same.

### DIFF
--- a/include/glaze/yaml/read.hpp
+++ b/include/glaze/yaml/read.hpp
@@ -4282,6 +4282,11 @@ namespace glz
                return; // Empty pair
             }
 
+            if (it == end) [[unlikely]] {
+               ctx.error = error_code::unexpected_end;
+               return;
+            }
+
             // Parse key
             if constexpr (str_t<first_type>) {
                // Skip anchor on key
--- a/tests/yaml_test/yaml_test.cpp
+++ b/tests/yaml_test/yaml_test.cpp
@@ -4471,6 +4471,31 @@ ship-to: *id001)";
       expect(!ec) << glz::format_error(ec, view);
       expect(glz::write_json(parsed).value_or("WRITE_ERR") == R"("x")");
    };
+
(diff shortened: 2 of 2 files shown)


wilkolbrzym-coder:
I’ll check this today around 18:00–19:00 UTC if the schedule holds. If not, tomorrow between 16:00 and 18:00 UTC.

wilkolbrzym-coder (approved):
## Review – PR #2822

The fix is minimal and correct.

The added guard:

if (it == end) [[unlikely]] {
    return;
}

is placed after the existing empty-pair check:

if (it != end && *it == '}') {
    return; // Empty pair
}

so valid `{}` and `{ }` still parse correctly.

### Edge cases covered

- Input ending exactly after `{`
- Input ending after `{` plus whitespace/tabs
- Non-null-terminated buffers (`std::string_view`, `std::span`)
- Key parsing for `std::pair<std::string, V>` and other first types, because the guard sits before key dispatch

The test is also good: it checks multiple truncated forms and asserts `ec.count <= view.size()`, which would catch any remaining overread.

No correctness issues found.

### Verdict

LGTM. Approve.

cc @uwezkhan @stephenberry

PR #2835, opened 2026-08-25
fix required keys, array sub-tables, and variant alternatives in TOML

### LINKED ISSUE #2832: `error_on_missing_keys` Missing keys are not generating an error on TOML files
[glaze_missing_key_toml.zip](https://github.com/user-attachments/files/31422710/glaze_missing_key_toml.zip)

I stumbled across a way, where the option, `error_on_missing_keys` is not generating an error.
A minimal example repository can be found attached.
The minimal example looks like:

Just to be sure I set both options: `.error_on_unknown_keys = true` and `.error_on_missing_keys = true`

The config struct:
```cpp
struct Config {
    std::string host;
    std::uint16_t port;
    std::string required_topic;
};
```

The main:
```cpp
int main(int argc, char** argv) {
    if (argc != 2) {
        std::cerr << "usage: glaze_missing_key_toml_repro <config.toml>\n";
        return 2;
    }

    std::ifstream file(argv[1]);
    if (!file) {
        std::cerr << "failed to open: " << argv[1] << "\n";
        return 2;
    }

    std::ostringstream buffer;
    buffer << file.rdbuf();
    const std::string toml = buffer.str();

    Config config{};
    constexpr glz::opts opts{
        .format = glz::TOML,
        .error_on_unknown_keys = true,
        .error_on_missing_keys = true,
    };

    const auto error = glz::read<opts>(config, toml);
    if (error) {
        std::cout << "glaze reported an error, as expected:\n";
        std::cout << glz::format_error(error, toml) << "\n";
        return 1;
    }

    std::cout << "glaze did not report an error\n";
    std::cout << "host: " << config.host << "\n";
    std::cout << "port: " << config.port << "\n";
    std::cout << "required_topic: '" << config.required_topic << "'\n";
    return 0;
}
```

with the valid TOML:
```toml
host = "localhost"
port = 1883
required_topic = "example/topic"
```

and the TOML with a missing key:
```toml
host = "localhost"
port = 1883
```

On my machine if I build it with:
```cmd
cmake -S build -B build
cmake --build build
```

And execute:
```cmd
.\build\glaze_missing_key_toml_repro.exe .\valid.toml
```

I get the expected output:
```cmd
glaze did not report an error
host: localhost
port: 1883
required_topic: 'example/topic'
```

But if i run it with the file that has a missing key, i do not get the expected error, but:
```cmd
.\build\glaze_missing_key_toml_repro.exe .\missing_required_topic.toml
glaze did not report an error
host: localhost
port: 1883
required_topic: ''
```

OS: Windows 11
Package Manager: BPM.cmake
Glaze Version: v8.1.0
Install Options: 
`-Dglaze_ENABLE_FUZZING=OFF`
`-DBUILD_TESTING=OFF`
`-Dglaze_BUILD_PERFORMANCE_TESTS=OFF`
`-Dglaze_SIMPLE_FLOAT_TEST=OFF`
`-Dglaze_BUILD_NETWORKING_TESTS=OFF`
`-Dglaze_DISABLE_SQLITE_TESTS=OFF`
`-Dglaze_BUILD_SSL_TESTS=OFF`
`-Dglaze_BUILD_WEBSOCKET_AUTOBAHN_TEST=OFF`and

stephenberry:
Fixes #2832.

`error_on_missing_keys` was never read by the TOML reader — a silent no-op at every level. `glz::read<glz::opts{.format = glz::TOML, .error_on_missing_keys = true}>` accepted a document missing a required field, whether that field was a top-level key, in a table body, in an inline table, or in an array-of-tables element.

## Why it isn't a one-liner

Every other format settles required keys where an object's braces close. TOML can't: a struct is assembled from disjoint regions of a document.

```toml
[server]
host = "h"

[logging]      # unrelated table between the two halves of [server]
level = "info"

[server.tls]   # still filling in server
enabled = true
```

`[server.tls]` never re-enters `server`'s parse — `resolve_nested` descends straight to `tls` — so a per-scope bitset would report `tls` missing on every document of this shape.

`missing_key_record` accumulates assignments across the whole read as a tree mirroring the values the document touched: one node per struct, and for an array of tables one node per array holding one per element. It is threaded through `resolve_nested`, `resolve_array_of_tables` and a new `op_tracked` on the object reader, then resolved by one type-driven walk once the document is parsed. Inline tables are self-contained by the grammar, so they are checked in place with the usual `bit_array` + `required_fields` pattern, which also covers structs reached as map values or as elements of an inline array. Nullable wrappers are looked through, so `std::optional<T>` / `std::unique_ptr<T>` members are not required but *are* checked once the document provides them.

`check_error_on_missing_keys(Opts)` joins the ~20 existing option-presence helpers in `core/opts.hpp` rather than reading the field directly. Without it this breaks `glz::toml::toml_opts` and any trimmed-down custom options struct — the docs explicitly invite those, and TOML never made its users declare this field.

Reporting reuses `error_code::missing_key` with `custom_error_message` naming the key, matching JSON/YAML.

## Two gaps this testing exposed, fixed alongside

**`[array.sub]` after `[[array]]` failed with `unknown_key`.** Valid TOML v1.0.0: the header names a sub-table of the element the most recent `[[array]]` opened. `resolve_nested` hit its "not a struct" guard on the `std::vector` and returned success *without consuming anything*, so the body under the header was then read as though it belonged to the enclosing table. It now descends into the array's last element.

**A variant with several object alternatives committed to the first.** `std::variant<point, pair>` reading `{ a = "s", b = 2 }` failed on `unknown_key` instead of trying `pair`. It now rewinds and tries each, following the policy BEVE already documents: `missing_key` is the caller's own strictness and `exceeded_max_recursive_depth` is a property of the input, so neither is retried past, and the deduced alternative's error is what gets reported when nothing fits.

## One thing worth reviewing closely

Retrying makes an ambiguous nest exponential, so it is charged against the per-read speculation budget. `charge_speculation` alone was not enough: at depth 30 an unresolvable nest still took **95 seconds**. The existing shape initializes the local `exhausted` flag to `false`, so after the global budget latches, every level still runs one retry before discovering it — two subtree parses per level, still 2^depth. Seeding it from `speculation_exhausted(ctx)` makes the cost flat: ~3 ms at any depth in a release build, ~50 ms in debug.

There is a regression test for this, deliberately calibrated: at depth 22 an unresolvable nest takes 41 s in a debug build without the fix against 0.56 s with it. Below about depth 18 the budget saturates before the nest does and both cost the same, so a shallower version would not discriminate — there is a comment saying so.

The same `exhausted = false` shape exists in the BEVE and JSON variant readers. I have not touched them here — their charges are larger per attempt, so the exposure differs — but they may deserve the same look.

## Tests

32 new tests, in a new `toml_reader_test` target: disjoint tables, dotted keys, inline and empty inline tables, arrays of tables, nullables, `requires_key`, `skip_null_members`, array sub-tables, variant alternative resolution and the speculation bound. Also verified against `glz::meta` structs, maps of structs, recursive types, `std::array` members and sequential reads on a reused context.

They are a separate translation unit because adding them to `toml_test.cpp` OOM-killed `cc1plus` on every GCC job — that file already compiles near the ceiling at 965 MB (clang), and these took it to 1241 MB. It is back to byte-identical with main; the new TU peaks at 366 MB, against 990 MB for `toml_test`, 1.76 GB for `yaml_test` and 2.57 GB for `json_test`. See the second commit for the details.

`toml_test` (389), `json_test` (725), `yaml_test` (760), `chrono_test` (126), `embedded_null_test` and `ostream_buffer_test` all unchanged and passing. clang-format clean; no warnings under `-Wall -Wextra -pedantic -Werror`.


wilkolbrzym-coder (commented):
Review pass.

The missing_key_record approach is correct for TOML’s disjoint-table model, and the separate TU is a good call.

Two areas need confirmation before approve:

1. Does parse_toml_object_members still call check_required_fields at the end for non-inline table bodies? If so, it reintroduces local required-key checking and defeats the whole-document record for normal tables.

2. read_object_alternative resets it/error/message between variant attempts, but I don’t see ctx.depth or equivalent parser state being restored. A failed first alternative that enters a nested table could leave the retry starting at the wrong nesting level.

A targeted test where a failing object alternative parses into a nested table before the next alternative succeeds would settle both.

No other blockers.

wilkolbrzym-coder (commented):
Review update – PR #2835

Both previous concerns are resolved.

- Whole-document missing-key tracking works for disjoint tables, dotted keys, and arrays of tables.
- Variant retry depth is covered by a targeted test; `ctx.depth` returns to zero.

Non-blocking:
- Failed variant parse may leave the variant in the last attempted alternative. Document this.
- `check_missing_keys_at` is O(n²) for large arrays of tables.
- `read_object_alternative` reads `Opts.error_on_unknown_keys` directly; consider the existing option-presence helper.

LGTM once CI is green.

cc @stephenberry

stephenberry:
Checked both against the code.

**`parse_toml_object_members` for normal tables** — it isn't reached for them. The function has one call site repo-wide (`read.hpp:2615`) and it passes `is_inline_table = true`; that's pre-existing, `main` has the same single `true` call site. Normal table bodies go through `op_tracked`, which threads the record node and never touches the local `fields` bitset. The trailing `check_required_fields` is only reachable when the inline-table loop falls out at `it == end`, where `unexpected_end` is already set and it early-returns.

**`ctx.depth` across variant attempts** — restored by RAII rather than by hand. `grep -n "ctx.depth" include/glaze/toml/read.hpp` finds no raw read or write anywhere in the file; every level is entered through a `depth_guard`, whose destructor decrements on the way out regardless of error. A failed alternative is back at its entry depth when `attempt` returns.

The test was a fair ask and the existing coverage didn't reach it — the closest case (`v = { s = "str" }`) fails on the first key without descending. Added in 650b4a73: the deduced alternative consumes a nested inline table, exits it, then fails on the next key, and the retry succeeds; it reads with an explicit `glz::context` so it can assert `ctx.depth == 0` afterwards.

Separately, `is_inline_table` being a runtime parameter that is only ever `true` is what makes this easy to misread. Pre-existing and out of scope here, but worth a follow-up to drop it.

stephenberry:
Follow-up on the `is_inline_table` point: opened as #2843, based on this branch since the function sits inside this PR's hunks.

Removing the dead branches exposed that a truncated inline table (`v = {`) was accepted silently — pre-existing on `main`, though this PR's trailing required-key check turned it into a `missing_key` naming a key the caller had written. Fixed there with tests.

wilkolbrzym-coder (approved):
Final review update – PR #2835

Both previous blockers are resolved.

- `parse_toml_object_members` is only called for inline tables.
- `ctx.depth` is restored by RAII; the new rewind test confirms `ctx.depth == 0`.
- Whole-document missing-key tracking works for disjoint tables, dotted keys, arrays of tables, and nullables.
- Variant retry is bounded and tested.

Remaining non-blocking notes:

- After all variant attempts fail, the variant may remain in the last attempted alternative.
- `check_missing_keys_at` is O(n²) for large arrays of tables.
- `read_object_alternative` reads `Opts.error_on_unknown_keys` directly; the existing presence helper would be more consistent.

CI is green except the s390x RPM job still in progress.

LGTM. Approve.

cc @stephenberry

PR #2841, opened 2026-08-26
Honor yaml_opts::indent_width

stephenberry:
Stacked on #2840 — it touches a helper added there. Retarget to `main` once that merges.

`yaml_opts::indent_width` never had any effect. All nine writer call sites read it from a default-constructed `yaml_opts{}` rather than from `Opts`, so it was always 2 while staying documented as configurable. Its sibling `check_flow_style(Opts)` and JSON's `check_indentation_width(Opts)` are both threaded correctly, so this was a copy-paste slip rather than a design decision.

### Changes

`check_indent_width` now takes `Opts` as a template parameter, following `csv_delimiter<Opts>()`, so the value can be validated where it is read.

Threading the option exposed a second bug: the compact sequence form only aligns at width 2. The dash takes the first column of its element's indent, so writing `- ` and then indenting the mapping's remaining keys by one more level puts them past the first key at any other width:

```yaml
# before, indent_width = 4 -- PyYAML rejects this: "mapping values are not allowed here"
items:
    - a: 1
        b: 2
```

The dash is now padded across the indent width, which is what PyYAML's emitter does:

```yaml
items:
    -   a: 1
        b: 2
```

A width below 2 leaves no room for that padding, so it is now a compile-time error rather than silently invalid output.

### Verification

- yaml_test 796, yaml_conformance 402, yaml_conformance_struct 105, chrono_test, embedded_null_test — all pass.
- Writer output at the default width is byte-identical across 5,161 corpus documents.
- Documents at widths 2, 3, 4 and 8 round-trip, both under their own options and under the default ones, and PyYAML parses each to the same structure.

### DIFF
--- a/docs/yaml.md
+++ b/docs/yaml.md
@@ -262,8 +262,20 @@ The `glz::yaml::yaml_opts` struct provides YAML-specific options:
 |--------|---------|-------------|
 | `error_on_unknown_keys` | `true` | Error on unknown YAML keys |
 | `skip_null_members` | `true` | Skip null values when writing |
-| `indent_width` | `2` | Spaces per indent level |
+| `indent_width` | `2` | Spaces per indent level when writing (minimum 2) |
 | `flow_style` | `false` | Use flow style (compact) output |
+
+A sequence dash takes the first column of its element's indent and is padded across the rest, so
+a mapping continued on following lines aligns with its first key:
+
+```yaml
+# indent_width = 4
+items:
+    -   a: 1
+        b: 2
+```
+
+Reading needs no matching option: YAML indentation is self describing.
 
 Example with flow style output:
 
--- a/include/glaze/yaml/opts.hpp
+++ b/include/glaze/yaml/opts.hpp
@@ -43,10 +43,17 @@ namespace glz::yaml
       return ret;
    }
 
-   consteval uint8_t check_indent_width(auto&& o)
+   // Takes Opts as a template parameter (like csv_delimiter) so the value can be validated: too
+   // small a width writes nested block content at or before its parent's column, which is not the
+   // same document and does not read back.
+   template <auto Opts>
+   consteval uint8_t check_indent_width()
    {
-      if constexpr (requires { o.indent_width; }) {
-         return o.indent_width;
+      if constexpr (requires { Opts.indent_width; }) {
+         static_assert(Opts.indent_width >= 2,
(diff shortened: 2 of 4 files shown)


wilkolbrzym-coder (commented):
Review notes – PR #2841

The diagnosis is correct: `yaml_opts::indent_width` was read from `yaml_opts{}` at every call site, so the configured width had no effect. This PR fixes that and also fixes a second, related YAML validity bug.

Correctness assessment

1. `check_indent_width<Opts>()` is the right shape.

It follows the existing `csv_delimiter<Opts>()` pattern and allows `static_assert` validation at the point where the option is consumed.

The `indent_width >= 2` compile-time check is justified. Width 1 would leave no room for the dash padding and would produce invalid block-sequence YAML.

2. `dump_dash_padding` correctly matches PyYAML’s emitter behavior.

For `indent_width = 4`:

```yaml
items:
    -   a: 1
        b: 2
```

This is valid and aligns continuation keys under the first key. The dash takes one column, and the padding fills the remaining columns.

3. The update threads `Opts` through the relevant writer helpers consistently.

Visible call sites now use `check_indent_width<Opts>()`, including:

- `write_yaml_string`
- `write_empty_block_line`
- `write_block_sequence`
- tagged writers
- mapping writers

This removes the copy-paste bug rather than patching only one call site.

4. `write_empty_block_inline(..., pad)` is correctly generalized.

The default `pad = 1` preserves the old `key: {}` behavior. Sequence callers pass `indent_width - 1`, matching `dump_dash_padding`.

5. Tests are strong.

Coverage includes:

- default width unchanged
- width 4 and width 3 exact output
- round trips at widths 2, 3, 4, 8
- reading without matching width
- flow style ignoring width

The byte-identical corpus check at default width is also valuable because it proves the refactor did not accidentally change existing output.

Non-blocking observations

- Several CI jobs are cancelled, not failed. This needs a clean rerun before merge.
- The PR description says this is stacked on #2840. If the base has already retargeted to `main`, make sure the helper from #2840 is actually present before CI is trusted.
- The hard `static_assert` for `indent_width < 2` is the right behavior, but it is a compile-time breaking change for users who previously set `indent_width = 1`; docs correctly state the minimum is now 2.

Verdict

No correctness bug found in the writer logic.

LGTM once CI is green and the stacked dependency is resolved.

cc @stephenberry

PR #2843, opened 2026-08-26
drop the always-true is_inline_table parameter

stephenberry:
Follow-up to #2835, and based on it — `parse_toml_object_members` sits inside that PR's hunks, so this does not apply to `main` on its own.

`parse_toml_object_members` took a `bool is_inline_table`, had exactly one call site, and that call site passed `true`. The `!is_inline_table` branches — skipping blank lines, skipping `[section]` headers, accepting a newline instead of a `,` — were unreachable. They also made the function read like a general table-body parser, which is misleading: normal table bodies go through `op_tracked`.

Renamed to `parse_inline_table_members` and dropped the parameter.

## The bug this exposed

With the dead branches gone, the loop has one well-formed exit: the `return` at `}`. Every other way out means the document ended inside the table. That path set no error, so a truncated inline table was accepted:

```cpp
struct point { int x{}; int y{}; };
struct holder { point v{}; };

holder value{};
glz::read_toml(value, "v = {");   // returned success
```

`v = {`, `v = {\n`, `v = { x = 1,` all parsed clean and left the struct default-initialized. Falling out of the loop now yields `unexpected_end`.

This is pre-existing and independent of #2835 — `main` accepts these with `error_on_missing_keys` either on or off. #2835 partly masked it: with the option on, the trailing required-key check turned the silent success into `missing_key`, which named a key the caller had in fact written.

## Tests

4 tests in `toml_reader_test`: truncation at each position, truncation reported as truncation rather than `missing_key` under `error_on_missing_keys`, and the complete and trailing-comma forms still reading. Verified they fail without the change.

`toml_reader_test` (42), `toml_test` (389) passing. clang-format clean.

### DIFF
--- a/include/glaze/toml/read.hpp
+++ b/include/glaze/toml/read.hpp
@@ -1665,7 +1665,7 @@ namespace glz
       void check_missing_keys_at(T&& value, const missing_key_record& node, Ctx& ctx);
 
       template <auto Opts, class T, class It, class End, class Ctx>
-      inline void parse_toml_object_members(T&& value, It&& it, End end, Ctx&& ctx, bool is_inline_table)
+      inline void parse_inline_table_members(T&& value, It&& it, End end, Ctx&& ctx)
       {
          using U = std::remove_cvref_t<T>;
          static constexpr auto N = reflect<U>::size;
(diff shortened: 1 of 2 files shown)


wilkolbrzym-coder (commented):
Review – PR #2843

Clean follow-up to #2835.

- Removing the always-true `is_inline_table` parameter is correct.
- The new `parse_inline_table_members` name reflects the function’s actual scope.
- Truncated inline tables now report `unexpected_end` instead of silently succeeding.
- Tests cover missing `}`, trailing comma, and strict-mode distinction between `unexpected_end` and `missing_key`.

One point to confirm:

The diff still shows `check_required_fields<Opts, U>(fields, ctx);` near the end of the function. If that call remains after the loop, strict mode could set `missing_key` before the final `unexpected_end` fallback.

The new truncation test would fail in that case, so this is likely just diff context. Please confirm that required-key checking now lives only on the `}` path.

LGTM once confirmed.

cc @stephenberry

PR #2844, opened 2026-08-26
bounds-check output writes in the binary-to-JSON converters

### LINKED ISSUE #2842: CBOR → JSON converter — safety and correctness gaps
## Issue: CBOR → JSON converter — safety and correctness gaps

**File:** `include/glaze/cbor/cbor_to_json.hpp`

### Summary

The converter is functionally useful, but several boundary-safety and buffer-management issues should be fixed before it is considered production-safe.

---

### 1. Pointer arithmetic in bounds checks — formal UB

Many checks use:

```cpp
if ((it + 2) > end)
```

A pointer may only point to an array element or one past the end. `it + 2` can already be outside the object.

Examples:

- `uint16_follows`
- `uint32_follows`
- `uint64_follows`
- `float16`
- `float32`
- `float64`

**Fix:** use `(end - it) < 2`, `(end - it) < 4`, or `(end - it) < 8`.

---

### 2. No `ensure_space` before writing to the output buffer

The converter calls low-level `dump()` repeatedly:

```cpp
dump('"', out, ix);
dump('[', out, ix);
dump(':', out, ix);
```

but never ensures the output has capacity:

```cpp
ensure_space(ctx, out, ix + needed)
```

For `std::string` initialized with size `0`, this is unsafe. The converter should either reserve a worst-case size or call `ensure_space` before writes.

---

### 3. Empty input returns success

```cpp
while (it < end) {
   detail::cbor_to_json_value<Opts>(...);
}
return {};
```

For an empty buffer, this produces no output and returns no error.

CBOR requires at least one data item. Empty input should return:

```cpp
error_code::unexpected_end
```

---

### 4. `ctx.depth` is not restored on error in prettified maps

In `major::map`:

```cpp
if constexpr (Opts.prettify) {
   ctx.depth += check_indentation_width(Opts);
}
```

If an error occurs inside parsing and the function returns early, `ctx.depth` is not decremented.

If `ctx` is ever reused, indentation state leaks.

---

### 5. Comment says `base64`, implementation writes hex

```cpp
// Byte string - encode as base64 in JSON
```

But code writes lowercase hex:

```cpp
// Simple hex encoding for now (TODO: proper base64)
```

Hex is acceptable, but the comment should match the implementation.

---

### 6. `bytes.resize(old_size + chunk_len)` can overflow on 32-bit

`chunk_len` is `uint64_t`.

```cpp
bytes.resize(old_size + chunk_len);
```

After the earlier bounds check, this is probably safe on 64-bit, but the addition may narrow or overflow on 32-bit `size_t`.

---

### 7. `typed_array::get_info(tag_num).element_size` should be checked for zero

```cpp
if (byte_len % ta_info.element_size != 0)
```

If `element_size == 0`, this is division by zero.

Though current `get_info` likely never returns a valid zero-size entry, a defensive check is appropriate.

---

### 8. `simple::undefined` is mapped to `null`

```cpp
case simple::null_value:
case simple::undefined:
   dump("null", out, ix);
```

This may be acceptable, but it should be documented, because CBOR `undefined` has no JSON equivalent.

---

### 9. `cbor_to_json_key` forwards `bstr` keys through hex string conversion

```cpp
case major::tstr:
case major::bstr:
   cbor_to_json_value<Opts>(...);
```

A CBOR byte-string key becomes a hex-encoded JSON string key.

This may be a valid converter choice, but it should be documented.

---

### Suggested priority

- Fix pointer arithmetic.
- Add output buffer safety.
- Reject empty input.
- Restore depth on all error paths.

The remaining comments are consistency/documentation issues.

cc @stephenberry

stephenberry:
Fixes #2842.

`dump()` grows a resizable buffer but does not bounds check one it cannot grow, so the binary-to-JSON converters wrote past the end of a `std::array` or `std::span` output buffer. Under ASan, `[[[]]]` converted into a 2-byte array is a stack-buffer-overflow.

All five converters now reserve before every write through shared `emit_char` / `emit_literal` / `emit_newline_indent` / `emit_hex_bytes` helpers in `glz::detail`, and report bytes written in `error_ctx::count` — a fixed-size output buffer previously had no way to learn how much of it held JSON.

### Also

- `(it + n) > end` bounds checks rewritten as `(end - it) < n`, which is the same test without the out-of-range pointer arithmetic. 40 sites across cbor, beve, msgpack, `core/reflect.hpp`, and `toml/skip.hpp`.
- `dumpn` no longer forms `&b[ix]` when there is nothing to write. A zero-width indent left `ix == size()`, out of range for a bounded buffer even though the `memset` never touched it.
- `ctx.depth` is restored on error paths, via an `indent_guard` beside the existing `depth_guard`.
- `cbor_to_json` requires exactly one CBOR data item. An empty buffer and a CBOR sequence both used to report success while producing text no JSON parser accepts.
- `beve_to_json` requires at least one value, and writes the NDJSON separator between values that were concatenated without a delimiter tag.
- float16 typed arrays (RFC 8746 tags 80/84) now convert; binary128 (83/87) is rejected before the array is opened rather than partway through it.
- `bson_to_json` moved onto the shared helpers, dropping its per-token 256-byte reservation. A 15-byte document needed a 271-byte fixed buffer and now needs 28.

### Tests

cbor 307, beve 456, bson 124, jsonb 163 under ASan/UBSan and libc++ hardening; eetf 43, json 725, yaml 796, toml 389, msgpack 70.

### DIFF
--- a/docs/binary.md
+++ b/docs/binary.md
@@ -480,6 +480,10 @@ Writing is Version 2 only. Version 2 output is not decodable as a variant by a G
 
 `glaze/binary/beve_to_json.hpp` provides `glz::beve_to_json`, which directly converts a buffer of BEVE data to a buffer of JSON data.
 
+An empty buffer holds no value, which is not a document, and fails with `error_code::unexpected_end` rather than converting to empty output. On success the returned `error_ctx::count` is the number of bytes written, which is how a fixed-size output buffer learns how much of it holds JSON.
+
+A buffer holding several values converts to one JSON document per line. See [Delimiter Format](#delimiter-format).
+
 ### String Escaping
 
 A BEVE string can hold any byte, including control characters (0x00–1F). JSON cannot. `\n`, `\t`, `\b`, `\f` and `\r` are escaped normally, but the rest have no short escape, and by default the conversion fails rather than write something that will not parse:
@@ -737,6 +741,8 @@ auto ec = glz::read_beve_delimited(messages, buffer);
 ### Delimiter Format
 
 The BEVE delimiter is a single byte: `0x06` (extensions type 6 with subtype 0). When converting delimited BEVE to JSON via `glz::beve_to_json`, each delimiter is converted to a newline character (`\n`), producing NDJSON-compatible output.
+
+Values concatenated without delimiters, which `glz::read_beve_delimited` also accepts, get the same newline between them. Two JSON documents never run together into text that is no longer JSON.
 
 ## Lazy BEVE Parsing
 
--- a/docs/cbor.md
+++ b/docs/cbor.md
@@ -72,6 +72,26 @@ auto ec = glz::cbor_to_json(cbor, json);
 
 A CBOR text string can hold control characters (0x00–1F), which JSON cannot. Those without a short escape make the conversion fail with `error_code::invalid_control_character`. To keep them, turn on `escape_control_characters`. See [String Escaping](binary.md#string-escaping).
 
+On success the returned `error_ctx::count` is the number of bytes written. A resizable buffer is resized to fit, but a fixed-size one (`std::array`, `std::span`) has no other way to learn how much of it now holds JSON. The same applies to `glz::beve_to_json`, `glz::bson_to_json`, `glz::jsonb_to_json`, and `glz::eetf_to_json`.
+
+The buffer must hold exactly one CBOR data item, which is what a CBOR document is. An empty buffer fails with `error_code::unexpected_end`, and bytes left over after the first item (a CBOR sequence, RFC 8742) fail with `error_code::syntax_error` rather than running the items together into text no JSON parser accepts.
+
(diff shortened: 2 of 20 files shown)


wilkolbrzym-coder (approved):
Review – PR #2844

LGTM. Approve.

Key fixes:

- `(end - it) < n` removes UB from all 40 bounds checks.
- Shared emit helpers add correct fixed-buffer checks.
- `dumpn` zero-length guard prevents forming `&b[ix]` at `size()`.
- `indent_guard` restores depth on error paths.
- Empty/sequence CBOR/BEVE rejected correctly.
- `error_ctx::count` for fixed buffers is the right API.

Tests are strong.

Non-blocking: `emit_hex_bytes` may overflow `size_t` on 32-bit if `2 * n` exceeds max; pre-existing class.

cc @stephenberry

PR #2853, opened 2026-09-06
Write variants by their declared shape in MessagePack and CBOR

### LINKED ISSUE #2851: Writing `glz::generic` types to MessagePack adds type name strings to output.
Firstly, let me thank you for all of the work you've put into this library!

I am trying to save a `glz::generic_u64` (this also happens with the other `glz::generic` types) to MessagePack. However, the output it produces has pairs for each value, where the first value is a type name string. For example, if I have the following JSON:
```json
{
    "a": [ 1, 2, 3 ],
    "b": {
        "i": 64,
        "ii": 6.28,
        "iii": null,
        "iv": 1.618033
    },
    "c": "this is some text."
}
```

When I save it to MessagePack (and then convert it back to JSON), this is what it produces:
```json
[
  "glz::ordered_small_map<struct glz::generic_json<2,struct glz::ordered_small_map> >",
  {
    "a": [
      "std::vector<glz::generic>",
      [
        [ "uint64_t", 1 ],
        [ "uint64_t", 2 ],
        [ "uint64_t", 3 ]
      ]
    ],
    "b": [
      "glz::ordered_small_map<struct glz::generic_json<2,struct glz::ordered_small_map> >",
      {
        "i": [ "uint64_t", 64 ],
        "ii": [ "double", 6.28 ],
        "iii": [ "nullptr", null ],
        "iv": [ "double", 1.618033 ]
      }
    ],
    "c": [ "std::string", "this is some text." ]
  }
]
```

This only happens with MessagePack, and not other binary formats (CBOR, JSONB, BEVE, et cetera).

Here's a simple test case to replicate:
```cpp
const std::string jsonString = R"json(
    {
        "a": [ 1, 2, 3 ],
        "b": {
            "i": 64,
            "ii": 6.28,
            "iii": null,
            "iv": 1.618033
        },
        "c": "this is some text."
    }
)json";

glz::generic_u64 json;
std::ignore = glz::read_json(json, jsonString);

std::vector<std::uint8_t> jsonBuffer{ };
std::ignore = glz::write<glz::opts{ .format = glz::JSON }>(json, jsonBuffer);

std::vector<std::uint8_t> msgpackBuffer{ };
std::ignore = glz::write<glz::opts{ .format = glz::MSGPACK }>(json, msgpackBuffer);

std::vector<std::uint8_t> cborBuffer{ };
std::ignore = glz::write<glz::opts{ .format = glz::CBOR }>(json, cborBuffer);

std::vector<std::uint8_t> jsonbBuffer{ };
std::ignore = glz::write<glz::opts{ .format = glz::JSONB }>(json, jsonbBuffer);

std::vector<std::uint8_t> beveBuffer{ };
std::ignore = glz::write<glz::opts{ .format = glz::BEVE }>(json, beveBuffer);

std::println("   JSON: {} bytes", jsonBuffer.size());
std::println("MSGPACK: {} bytes", msgpackBuffer.size());
std::println("   CBOR: {} bytes", cborBuffer.size());
std::println("  JSONB: {} bytes", jsonbBuffer.size());
std::println("   BEVE: {} bytes", beveBuffer.size());
```

On MSVC, this will output the following:
```
   JSON: 86 bytes
MSGPACK: 338 bytes
   CBOR: 86 bytes
  JSONB: 89 bytes
   BEVE: 99 bytes
```

And on GCC, it will output the following:
```
   JSON: 86 bytes
MSGPACK: 305 bytes
   CBOR: 86 bytes
  JSONB: 89 bytes
   BEVE: 99 bytes
```

The output for MessagePack is several times larger than all of the others due to these type names. Also, since they are compiler-specific type names, the output sizes will vary for each compiler (for example, MSVC prepends `struct` to some template typename parameters, whereäs GCC does not).

Let me know if you need any more information.

As an aside, is it possible to serialise binary formats without this extra type information when writing a `glz::generic` object (and not having to write a custom (de)serialiser)? For example, CBOR outputs still produce these type–value pairs, but the first member is the variant index instead of a type name.  
For my use cases, I'd much rather the output be identical to the input JSON when converted. I've messed around with a few of the fields in `glz::opts` and some of the extension options (such as `structs_as_arrays` and `write_type_info`), but nothing I tried worked.

stephenberry:
Fixes #2851.

`glz::generic` written to MessagePack came out several times larger than the equivalent JSON, with a compiler-specific type name attached to every element. The cause was general rather than specific to `glz::generic`: MessagePack and CBOR ignored `glz::meta` and wrapped every variant in a discriminator of their own invention — the id in MessagePack, the index in CBOR. Without a `glz::meta::ids` declaration `ids_v` falls back to `glz::name_v`, which for a type with no `meta::name` is the compiler's own spelling of it, so a buffer written by MSVC could not be read by a GCC build.

## What changed

MessagePack and CBOR now write a variant in the shape its `glz::meta` declares, as JSON and BEVE do:

| `glz::meta` | Shape |
|---|---|
| `tag` alone | `{ tag : id, ...members }` |
| `tag` and `content` | `{ tag : id, content : value }` |
| neither | the active alternative's own value, bare |

Nothing is invented for a variant the author did not describe. Reading an undeclared variant tries the alternatives in declaration order against the format's own self-describing type information.

`glz::generic` is a variant of exactly the JSON value categories that declares no `tag`, so it needs no special case: it falls out of that rule as plain MessagePack and CBOR that any implementation can consume. The issue's example goes from 303 bytes to 63 and round-trips exactly. The four `generic_json` specializations the first draft of this PR added are gone, along with the `assign_number` helper they shared.

MessagePack structs are now keyed maps. A plain aggregate went through `to_tie` to the tuple writer before, so the same struct was a map when it had a `glz::meta` and a positional array when it did not, and an internally tagged variant over plain aggregates had no keys for its discriminator to merge into. `structs_as_arrays` still selects the positional form.

Untagged resolution is speculative -- an alternative is parsed to find out whether it fits -- so it charges the shared speculation budget and stops on `exceeded_max_recursive_depth`, as the BEVE, JSON, and TOML readers already do. It runs a strict conversion pass before a lenient one, so `variant<double, int64_t>` keeps an integer in `int64_t` rather than letting the float reader claim it.

## Also fixed

- `glz::read_msgpack` into a `glz::generic` did not compile. `no_header` leaked into nested parses, making the header-reading `parse<MSGPACK>::op` overload unviable one level down, so any `glz::meta::value = &T::some_container` hit it.
- `from<MSGPACK, int64_t>` read `uint64` `UINT64_MAX` as `-1`. The range check ran after narrowing to `int64_t`, so it could never fail for that target.
- `from<CBOR, double>` rejected the integer major types, leaving a CBOR integer too large for any integer target with nowhere exact to fall back to.
- JSONB integer alternatives fall through on sign or magnitude mismatch, so a negative number reads back in `u64` mode.
- CBOR rejected indefinite-length maps for a tagged variant, though its object reader accepts them and that is what a streaming encoder emits.
- Adjacent tagging accepted a map that resolved a discriminator but carried no `content` key, returning success with a default-constructed alternative.
- `glz::meta::ids` declared as integrals rather than strings now writes as well as reads.

## Compatibility

**This is a wire format break**, in three places:

- A MessagePack or CBOR variant written by Glaze 8.3.0 or earlier will not read back. In MessagePack that data was already compiler-specific, which is what makes it worth replacing.
- A MessagePack struct written by 8.3.0 or earlier reads back only under `structs_as_arrays`.
- Reflection needs member names, so a reflected struct must be a type with linkage. A function-local aggregate no longer serializes to MessagePack; it never did to JSON, CBOR, or BEVE.

Alternatives of an undeclared variant that share a wire shape can no longer be told apart on read. Declare a `tag` when that matters.

Under `structs_as_arrays`, adjacent tagging projects to the two element array `[id, value]` and internal tagging is a compile error directing you to declare `content` -- the same rule BEVE applies to positional writes.

## Testing

121/121 of the suite passes with ASan/UBSan enabled. New coverage: each tagging shape in both formats with byte-exact expectations, a discriminator sitting after the members, an id naming no alternative, the short-`ids` fallback to the first unlabeled alternative, bare output being byte-identical to the alternative written alone, items with no JSON counterpart being rejected, and every `num_mode` round-tripping, an exact alternative winning over a converting one, an ambiguous nest staying bounded, an input-level failure reporting itself rather than `no_matching_variant_type`, integral ids, indefinite-length CBOR maps, a missing `content` key, and the positional projection.


wilkolbrzym-coder (commented):
Review – PR #2853

Worth adding:
- Explicit wire-format break note for Glaze <=8.3.0 in changelog.
- Direct test for reserved MessagePack type 0xC1.

LGTM once CI is green.

cc @stephenberry

PR #2855, opened 2026-09-09
quote non-bare map keys in the toml writer

uwezkhan:
The TOML writer emits runtime map keys verbatim as bare keys, at both key sites (write_inline_map and the top-level map writer in toml/write.hpp). TOML bare keys only admit [A-Za-z0-9_-], so a std::map key holding a dot, whitespace, `=`, a quote, or a line break splices into the document instead of naming one entry. write_toml of `{"a.b": 1}` emits `a.b = 1`, which is the nested table a -> b (and the reader then rejects it), and `{"x = 1\ninjected": 2}` emits two lines that read back as `{x: 1, injected: 2}`. write_toml returns success in both cases.

Route both key sites through a write_toml_key that emits a quoted basic string with the existing char_escape_table escaping when the key is not a valid bare key, and leaves bare keys byte-identical. Before, a non-identifier key produced a different document or forged entries; after, the key round-trips. The YAML writer already quotes runtime keys through write_yaml_string and the CSV writer through dump_csv_string, so keeping the decision inside the map writer matches the siblings and callers keep passing a plain map. Tradeoff: a quoted key adds the two quotes plus a byte per escaped character, and (like the string value writer) a control byte with no short escape still passes through, but the structural characters that break framing are all escaped.

### DIFF
--- a/include/glaze/toml/write.hpp
+++ b/include/glaze/toml/write.hpp
@@ -35,6 +35,58 @@ namespace glz
    constexpr bool skip_toml_field =
       always_skipped<T> || (!check_write_function_pointers(Options) && is_member_function_pointer<T>);
 
+   // TOML bare keys admit only [A-Za-z0-9_-]. A runtime map key holding any other
+   // byte (a dot, whitespace, '=', a quote, a line break, an empty key) is not a
+   // bare key: written verbatim it splices into the document, turning one entry
+   // into a different structure or several forged entries. Detect the bare case so
+   // ordinary keys stay byte-identical.
+   GLZ_ALWAYS_INLINE constexpr bool toml_key_is_bare(const sv key) noexcept
+   {
+      if (key.empty()) return false;
+      for (const char c : key) {
+         const auto u = uint8_t(c);
+         const bool ok =
+            (u >= 'A' && u <= 'Z') || (u >= 'a' && u <= 'z') || (u >= '0' && u <= '9') || u == '_' || u == '-';
+         if (!ok) return false;
+      }
+      return true;
+   }
+
+   // Write a TOML map key. A bare key is written unquoted (output unchanged);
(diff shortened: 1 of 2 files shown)


wilkolbrzym-coder (commented):
Review – PR #2855

The fix is correct. `toml_key_is_bare` follows the TOML bare-key rule, `write_toml_key` reuses the existing escape table, and both key sites go through the same helper. Bare keys stay byte-identical.

Two things before approving:

1. Redundant `ensure_space` right before `write_toml_key` in the top-level map writer. Not a bug, but it reserves `key.size() + 4` and then the helper reserves again. Clean it up.

2. Control bytes with no short escape still pass through unescaped, so a key containing e.g. `0x01` produces invalid TOML. The string value writer has the same limitation, but here it defeats part of the fix's purpose. Either escape them as `\u00XX` or state this explicitly as out of scope.

Tests are good otherwise.

Not approving until those are addressed or explicitly deferred.

cc @stephenberry @uwezkhan

uwezkhan:
Pushed the control-byte handling. Before, a key holding e.g. 0x01 wrote the raw byte inside the quotes and reparsed as invalid TOML; after, it goes out as a backslash-u escape and round-trips. The structural bytes (newline, quote, backslash) already had short escapes in `char_escape_table`, so this only closes the validity gap, not a new injection path. Worst-case reservation grows from `2*size+2` to `6*size+2` to cover the six-byte sequence, and there's a control-byte test now.

On the redundant `ensure_space`: that pre-reservation is already gone in this revision. Both key sites dropped the old `ix + key.size() + N` reserve when they moved to `write_toml_key` (the helper owns the key bytes now); each site only reserves 3 more afterward for `" = "`. So there was no double reserve left to remove.

stephenberry:
Thanks for this!

PR #2856, opened 2026-09-11
Fix compilation of tests under new libstdc++, fix a couple of tests (UBSan + reflection)

nextsilicon-itay-bookstein:
* Fix UBSan error of zero-sized memcpy
* Fix build error regarding instantiation with incomplete type under libstdc++ from gcc-toolset-14
* Fix reflection test assertion under clang/libstdc++


stephenberry:
Thanks, all three fixes look right.

I pushed one commit to the branch: the indefinite-length byte string has the same null `memcpy` a few lines up. `value.resize(old_size + chunk_len)` on an empty vector leaves `data()` null, so `0x5F 0x40 0xFF` (indefinite bstr with one empty chunk) hits it the same way `0x40` does. Added the matching guard plus a test covering both, since the asan job builds with `-fsanitize=undefined` and will catch any regression there.

One note on the reflection test: every clang job in CI uses `-stdlib=libc++`, so clang + libstdc++ isn't a configuration we actually build. Your branch is almost certainly correct, but CI going green won't be what verifies it. That test is up to five branches of exact type-name matching now, so I may just loosen it to a substring check later.

PR #2859, opened 2026-09-11
Bound ostream_buffer memory to its unflushed window

stephenberry:
`basic_ostream_buffer` satisfies `vector_like`, so the write paths grew it with `resize(2 * k)`, where `k` is a **logical** end index. The buffer maps that onto physical storage as `k - flush_offset_`, so doubling a logical index sized the window to the document rather than to the unflushed span: a `glz::ostream_buffer<512>` held 1.19 MB while writing a 2.13 MB document, which defeats the bounded memory usage streaming exists to provide. Output was always correct; only memory was affected.

Growth becomes a `buffer_traits` customization point (`buffer_traits::grow`, and a free `grow_buffer`). Resizable buffers keep 2x growth. `basic_ostream_buffer` reserves the window the write needs plus half its configured capacity, so memory stays tied to the capacity the user chose. Every logical-index growth site in the write paths routes through it: `dump`, `write_chars`, `float_format`, the beve/cbor writers, and the lazy view writers. `grow_buffer` falls back to the old doubling when a `buffer_traits` specialization does not define `grow`, so user specializations written against the documented traits API keep compiling.

NDJSON had no flush point at all, so its window tracked the document even once growth was bounded. Records are independent by construction, which makes the record boundary a flush point; all three `to<NDJSON, T>` writers now take it.

Physical bytes held, same documents before and after:

| shape | capacity | document | before | after |
| --- | --- | --- | --- | --- |
| reflected structs | 512 | 2.13 MB | 1.19 MB | 1031 B |
| generic, prettified | 512 | 274 KB | 137 KB | 790 B |
| reflected structs | 64 KB | 2.31 MB | 2.10 MB | 66 KB |
| lazy json views | 512 | 560 KB | 524 KB | 512 B |
| ndjson records | 512 | 1.30 MB | 2.13 MB | 798 B |

No throughput regression: writing a 21.15 MB document to a streaming buffer takes 18.87 ms before and 18.15 ms after (best of 12).

`min_ostream_buffer_size` now derives from `write_padding_bytes`, which its comment always claimed it did. That derivation is what guarantees the growth slack covers the padding a write dumps past its request, and a `static_assert` pins the invariant so a change to `write_padding_bytes` fails loudly rather than silently eating the margin. `buffer_traits::grow` is constrained on `is_resizable`, so it is absent rather than ill-formed for a fixed buffer and `grow_buffer`'s detection of it stays honest.

Two cases remain proportional to the input and are left for separate work: a single value larger than the buffer cannot be flushed mid-value, so the window must still cover it (at roughly 2x the value for strings, which reserve for escapes); and the numeric-array fast path reserves worst-case space for the whole array before writing elements unchecked. `eetf/ei.hpp` is also left alone, since that encoder needs the whole term contiguous. `docs/streaming.md` now documents both of these on the write side, mirroring the section that already covered them for reads, and `docs/writing.md` documents `grow` as a customization point.

Adds bounded-growth cases to `ostream_buffer_test` (JSON, prettified generic, beve/cbor, ndjson arrays and tuples, and the default capacity) asserting the window stays near capacity with byte-identical output, plus a prettified reflected-struct case covering the flush offset on the `&b[ix]` write path. Full suite passes (121/121, ASan + UBSan).


wilkolbrzym-coder (commented):
Review – PR #2859

Approach is correct. `grow_buffer` as a `buffer_traits::grow` customization point is the right shape, and the `requires { traits::grow(...) }` fallback preserves user specializations written against the old trait API.

Behavior on regular buffers (std::string, vector, std::span with capacity) is unchanged: `grow_buffer(b, k)` reduces to `b.resize(2 * k)`.

The ostream_buffer path is where the fix matters, and the tests confirm the window stays bounded on >1 MB documents.

Two non-blocking observations:

1. Fresh ostream_buffer over-allocates relative to before

The old path did `b.resize(b.size() == 0 ? 128 : b.size() * 2)`. The new path does `grow_buffer(b, b.size() == 0 ? 64 : ix + 1)`.

For a regular buffer, `64 → resize(128)` matches old behavior. For an ostream_buffer, `grow(64)` becomes `resize(64 + N/2)`. With `N = 512` that is `resize(320)` for a fresh buffer, where the old code allocated 128. Bounded and one-shot, but a small regression in the very first allocation. Using `b.size() == 0 ? 1 : ix + 1` would keep the fresh path minimal; the slack already comes from `grow`.

2. `grow_buffer` in the lazy writers changed the growth shape

`json/lazy.hpp` and `beve/lazy.hpp` moved from:

    b.resize((std::max)(b.size() * 2, ix + n));

to:

    grow_buffer(b, ix + n);

For a regular buffer this is now `b.resize(2 * (ix + n))` rather than `max(2 * size, ix + n)`. When `ix + n > size` but `ix + n < 2 * size`, the new path allocates roughly twice what the old path did. Same correctness, slightly more memory churn in the lazy write paths. Worth checking whether the growth policy for these should mirror `dump` (double the required index) or keep the `max` form.

Everything else — the wiring in `dump.hpp`, `write_chars.hpp`, `float_format.hpp`, `beve/write.hpp`, `cbor/write.hpp`, the new tests, and the `ensure_capacity` change on ostream_buffer — is correct.

LGTM once the two above are either fixed or explicitly deferred.

cc @stephenberry

stephenberry:
Thanks for the careful read, and for checking the fallback path for existing `buffer_traits` specializations.

**1. Fresh-buffer allocation.** The arithmetic is right, but the branch is unreachable for `ostream_buffer`: `size()` returns `logical_size_`, which the constructor sets to `initial_capacity`, so a constructed buffer never reports `size() == 0`. It takes an explicit `initial_capacity = 0` to get there, and even then `resize(320)` stays inside the allocation a 512-byte buffer makes anyway.

I'd rather not take `b.size() == 0 ? 1 : ix + 1` — for a regular buffer that becomes `resize(2)` on the first write and doubles from there, spending several reallocations on the common path to save 192 bytes on one that can't occur.

**2. Lazy writer growth.** This one is deliberate. `b.size()` on an `ostream_buffer` is the *logical* size, which climbs with the document, so `max(b.size() * 2, ix + n)` sizes storage to twice the logical position — the exact pattern this PR fixes. Measured on a 490 KB document (2000 records) through `to<JSON, lazy_json_view>` into `ostream_buffer<512>`:

- `grow_buffer(b, ix + n)`: **512 B** window
- `max(b.size() * 2, ix + n)`: **262,303 B** window

Byte-identical output either way. Keeping the `max` form would leave the lazy writers unbounded.

The regular-buffer cost is real but bounded: across the `size < ix + n < 2 * size` interval the ratio to the old form runs between 1x and 2x, and `grow_buffer(b, k)` → `resize(2 * k)` is what `dump` already does, so this makes the lazy paths consistent with the rest of the writers.

Since your review the branch also picked up ndjson record-boundary flushing (no flush point existed, so its window tracked the document), `min_ostream_buffer_size` derived from `write_padding_bytes` with a `static_assert` pinning the slack invariant, and `buffer_traits::grow` constrained on `is_resizable` so detection no longer gives a false positive for fixed buffers.

PR #2860, opened 2026-09-12
handle yaml control characters

uwezkhan:
YAML's `c-printable` production excludes the C0 control range apart from `\t`, `\n` and `\r`, and also excludes DEL (`0x7F`). Glaze neither rejected these on read nor had a consistent story for them on write.

**Reading rejects them.** A raw control byte now errors with `invalid_control_character` in every position that can carry one: plain, single-quoted, double-quoted and block scalars, mapping keys, and comments. The `\xXX` escape form remains the supported way to carry such a character.

The check is free. The plain scalar, multiline and mapping-key loops bulk-copy each run of ordinary content and dispatch only on bytes a table marks, with the forbidden bytes included in that table, so rejection rides a lookup the scan already performs. This also replaces a per-character `push_back` and compare chain on the common path.

**Writing escapes them only on request.** Gated behind the existing `escape_control_characters` option (default off), so the default write path is byte-identical to before. `\n`, `\r` and `\t` still force quoting unconditionally, since a raw line break ends a plain scalar structurally rather than being an escaping concern.

This gives the same write/read asymmetry `glz::write_json` and `glz::read_json` already have: by default the writer emits a raw control byte the reader refuses, and the option is what makes such a value round trip.

Breaking change: YAML containing raw control bytes now fails to parse.

Docs updated in `docs/yaml.md` and `docs/options.md`.

### DIFF
--- a/include/glaze/yaml/common.hpp
+++ b/include/glaze/yaml/common.hpp
@@ -1400,9 +1400,14 @@ namespace glz::yaml
          return true;
       }
 
-      // Check for characters that require quoting
+      // Check for characters that require quoting. A C0 control byte (< 0x20)
+      // cannot appear literally in a plain scalar, so it forces a quoted style;
+      // the double-quoted writer then escapes it as \xXX. This mirrors the
+      // control-character check the multiline block and double-quoted paths
+      // already apply, which the single-line plain path was missing. The listed
+      // \n, \r and \t are all < 0x20, so the earlier explicit form is subsumed.
       for (char c : s) {
-         if (c == ':' || c == '#' || c == ',' || c == '\n' || c == '\r' || c == '\t') {
+         if (c == ':' || c == '#' || c == ',' || static_cast<unsigned char>(c) < 0x20) {
             return true;
          }
       }
--- a/tests/yaml_test/yaml_test.cpp
+++ b/tests/yaml_test/yaml_test.cpp
@@ -906,6 +906,48 @@ suite yaml_writer_edge_case_tests = [] {
       expect(parsed == original);
    };
 
+   "write_scalar_with_only_control_char_is_escaped"_test = [] {
+      // A control byte with no other quote-forcing character must still route to the
+      // double-quoted style. Emitting it in a plain scalar produces output no longer
+      // valid per the YAML character set (only \t, \n and \r are permitted C0 controls).
+      const std::string original = std::string("abc") + char(0x01) + "def";
+      std::string yaml;
+      auto wec = glz::write_yaml(original, yaml);
+      expect(!wec);
+      expect(yaml == "\"abc\\x01def\"");
+
+      std::string parsed{};
+      auto rec = glz::read_yaml(parsed, yaml);
+      expect(!rec) << glz::format_error(rec, yaml);
+      expect(parsed == original);
+   };
+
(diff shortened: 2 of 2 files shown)


wilkolbrzym-coder (commented):
Review – PR #2860

Not yet approving. One gap in the same class as the bug this PR fixes.

The change is correct as far as it goes:

- static_cast<unsigned char>(c) < 0x20 is a strict superset of the old \n, \r, \t checks.
- Casting to unsigned char before comparison is right.
- Both value and map-key paths are covered.
- Existing quoted outputs are unchanged.
- Tests are solid: single control byte, embedded NUL, map key with ESC, round-trips.

Gap

0x7F (DEL) is not a valid YAML character. YAML 1.2 c-printable is:

    #x9 | #xA | #xD | [#x20-#x7E] | #x85 | [#xA0-#xD7FF] | ...

0x7F sits outside [#x20-#x7E], so a string containing DEL written as a plain scalar still produces output a strict parser rejects. Same failure mode as the C0 case this PR fixes.

Same reasoning applies to C1 controls (0x80–0x9F) appearing as standalone bytes, but that only matters for invalid UTF-8 input, so lower priority.

Suggested fix:

    if (c == ':' || c == '#' || c == ',' ||
        static_cast<unsigned char>(c) < 0x20 ||
        static_cast<unsigned char>(c) == 0x7F) {
       return true;
    }

Or state explicitly in the comment that DEL and C1 are out of scope.

Everything else LGTM.

cc @stephenberry @uwezkhan

stephenberry:
Thanks @uwezkhan, the underlying catch was real: the single-line plain path was the only one of the three string paths not accounting for control characters.

I've reworked the approach on this branch so it matches how Glaze already handles this for JSON.

**Reading is the conformance gate.** The reader now rejects raw control bytes (the C0 range apart from `\t`, `\n`, `\r`, plus DEL) with `invalid_control_character` across every path that can carry one: plain, single-quoted, double-quoted, literal block, folded block, mapping keys, and comments. The `\xXX` escape form is still accepted, since that is how such a byte is legitimately carried.

**Writing does not pay for it.** Escaping is gated behind the existing `escape_control_characters` option (default off) using `if constexpr`, so the default write path is byte-identical to before. Only `\n`, `\r` and `\t` still force quoting unconditionally, because a raw line break ends a plain scalar structurally rather than being an escaping concern.

**The read-side check is free.** The plain scalar, multiline and mapping-key loops now bulk-copy each run of ordinary content and dispatch only on bytes a table marks, with the forbidden control bytes included in that table. The rejection rides a lookup the scan already performs, and it replaces a per-character `push_back` and compare chain on the common path. Comments are handled the same way: the scan stops on an invalid byte rather than swallowing it to end of line, so no error plumbing was needed through `skip_comment`'s call sites.

That leaves a deliberate asymmetry: by default the writer emits a raw control byte the reader then refuses. `glz::write_json` / `glz::read_json` already behave exactly this way today, so YAML now matches JSON instead of introducing a third behavior.

Worth flagging: this is a breaking change for reads, since YAML containing raw control bytes now errors. One rough edge, a control byte in a trailing comment surfaces as `syntax_error` rather than `invalid_control_character`; it is rejected either way.

121/121 test suites pass locally.

uwezkhan:
Thanks for taking this one and moving the gate to the reader, that's a cleaner shape than quoting on write.

PR #2869, opened 2026-09-15
Fix YAML quote lookahead crossing line boundaries

### LINKED ISSUE #2826: yaml: plain scalar containing an apostrophe breaks a later single-quoted scalar when reading into glz::generic
Reading into `glz::generic`, a plain scalar containing `'` makes a later single-quoted scalar containing `": "` fail to parse. 

Either construct alone is fine, and the order matters. 
The typed reader accepts the same document.

Version `2f6f5dcfe1a689bf79b443d83434fab264ce0a21`, MSVC 19.51, `/std:c++latest`.

### Repro

`cl /std:c++latest /EHsc /I<glaze>/include repro.cpp`

```cpp
#include <glaze/glaze.hpp>
#include <glaze/yaml.hpp>

#include <iostream>
#include <string>

struct Pair
{
    std::string one{};
    std::string two{};
};

int main()
{
    const std::string both{ "one: a'b\ntwo: 'c: d'\n" };
    const std::string apostrophe{ "one: a'b\n" };
    const std::string quotedColon{ "two: 'c: d'\n" };

    glz::generic_u64 document{};
    if (const auto ec = glz::read_yaml(document, both))
    {
        std::cout << "generic, both lines: " << glz::format_error(ec, both) << '\n';
    }

    glz::generic_u64 a{};
    std::cout << "generic, apostrophe alone:  " << (glz::read_yaml(a, apostrophe) ? "failed" : "ok") << '\n';

    glz::generic_u64 b{};
    std::cout << "generic, quoted colon alone: " << (glz::read_yaml(b, quotedColon) ? "failed" : "ok") << '\n';

    Pair pair{};
    std::cout << "typed, both lines:           " << (glz::read_yaml(pair, both) ? "failed" : "ok") << '\n';
}
```

### Actual

```
generic, both lines: 1:6: syntax_error
   one: a'b
        ^
generic, apostrophe alone:  ok
generic, quoted colon alone: ok
typed, both lines:           ok
```

### Expected

`generic, both lines: ok`.

The document is valid YAML. `'` is an indicator only in first position, so `a'b` is a well formed
plain scalar, and `'c: d'` is a well formed single-quoted scalar. PyYAML and ruamel.yaml both load
the combined document as `{'one': "a'b", 'two': 'c: d'}`.

### Narrowing

Each of these reads into `glz::generic`:

```
FAIL  one: a'b      + two: 'c: d'      both parts present
ok    one: ab       + two: 'c: d'      apostrophe removed
ok    one: a'b      + two: 'cd'        colon removed
ok    one: a'b      + two: "c: d"      second scalar double quoted
ok    one: 'a''b'   + two: 'c: d'      first scalar quoted
ok    two: 'c: d'   + one: a'b         order reversed
```

So it needs a plain scalar holding an apostrophe, followed later by a single-quoted scalar holding `": "`. 
Quoting the first, or double-quoting the second, or swapping them, all avoid it.

The caret points at line 1, which parses on its own; the document only fails once line 2 is present.
The order dependence suggests the scan for the `": "` mapping indicator treats the apostrophe in the
plain scalar as opening a quoted region, so the colon on line 2 is seen as quoted and the key/value
split is mislocated.

Skold177:
Fixes #2826.

A quote inside a YAML plain scalar could cause a later quoted value containing `: ` to produce a `syntax_error`. The inline mapping lookahead treated the quote as an opening delimiter and scanned across the newline, allowing later content to affect the earlier value.

## What changed

Keep `inline_value_has_plain_mapping_indicator` on the current physical line, including while skipping quoted text and after backslash escapes.

Add regression coverage for single and double quotes, LF/CRLF/CR line endings, generic numeric modes, typed maps, nested collections, and multiline quoted strings.

The issue’s literal example also has an extra space before `two:`. The valid regression uses aligned sibling keys, with a separate test preserving rejection of inconsistent indentation.

## Tests Run

Windows x64, MSVC 19.51, C++23, Debug:

- Full CMake build succeeded.
- Full CTest suite: **123/123 passed**, including SSL, SQLite, and the external YAML conformance corpus.
- `key_transformers_test`, run manually because it is not registered with CTest: **15/15 passed**.

The new regression cases reproduced the parsing failure before the fix.


wilkolbrzym-coder (commented):
Review – PR #2869

Fix is correct in intent. Bailing at the first newline inside the quoted-skip loop is the right guard for #2826.

One gap:

A YAML key may be a multiline quoted scalar with `: value` on the closing line:

    key: "line1
      line2": value

The fix returns false at the newline after `line1`, so it never sees the trailing `: value`. Not covered by the new tests. Confirm this is handled elsewhere or add a test.

Also verify the extra `++pos` is only on the non-escape path. If it runs on both, the loop advances twice per iteration. MSVC passes, but check it.

cc @stephenberry @Skold177

Skold177:
> Review – PR #2869
> 
> Fix is correct in intent. Bailing at the first newline inside the quoted-skip loop is the right guard for #2826.
> 
> One gap:
> 
> A YAML key may be a multiline quoted scalar with `: value` on the closing line:
> 
> ```
> key: "line1
>   line2": value
> ```
> 
> The fix returns false at the newline after `line1`, so it never sees the trailing `: value`. Not covered by the new tests. Confirm this is handled elsewhere or add a test.
> 
> Also verify the extra `++pos` is only on the non-escape path. If it runs on both, the loop advances twice per iteration. MSVC passes, but check it.
> 
> cc @stephenberry @Skold177

Thanks for the timely review. I added tests covering:
- Your exact block example, confirming it is rejected because implicit block keys cannot span multiple lines.
- Valid multiline quoted keys in explicit block mappings (?) and flow mappings ({...}).
- Escaped quotes and backslashes, including escapes immediately before closing quotes and quoted values containing : .

I also verified the ++pos behavior: the escape path intentionally advances twice - once past the backslash and once past the escaped character. Ordinary characters advance once.

Re-ran all tests locally and everything passes.

PR #2871, opened 2026-09-15
Handle // comments and report scanner errors in the JSONC formatters and read path (#2864)

### LINKED ISSUE #2864: minify_jsonc/prettify_jsonc silently truncate on // comments
**Version:** v8.3.0 (`b91062d`)

`read_jsonc_comment` (`include/glaze/json/json_format.hpp:156`) only handles `/* */`. Given a `//` comment it skips two characters, scans for a closing `*/`, never finds one, consumes to `end` and returns an empty view — so everything after the `//` is dropped and the caller is told nothing.

The header comment in `minify.hpp`/`prettify.hpp` already says only `/* */` is supported, so the limitation itself is intentional. The problem is that unsupported input silently produces truncated output rather than an error.

**Repro**

```cpp
#include <glaze/json.hpp>
#include <iostream>

int main()
{
   std::string in = "{\"i\":1, // note\n \"s\":\"x\"}";

   std::cout << "[" << glz::minify_jsonc(in) << "]\n";

   std::string out{};
   glz::prettify_jsonc(in, out);
   std::cout << "[" << out << "]\n";
}
```

```
[{"i":1,]
[{
   "i": 1,
   ]
```

`glz::read_jsonc` parses the same input correctly, so this is specific to the minify/prettify path.

**Two separate gaps**

- `detail::minify_json` does set `error_code::syntax_error`, but the public `glz::minify_jsonc` wrappers return `void`/`std::string` and drop it.
- `detail::prettify_json` sets no error at all (`error_code::none`), so even the detail API reports success on truncated output.

**Suggestion**

Either handle `//` comments in `read_jsonc_comment` (when minifying they would have to be dropped rather than preserved, since the terminating newline goes away), or set `syntax_error` on an unterminated comment in both paths so the detail API is trustworthy and the public wrappers have something to surface.

wilkolbrzym-coder:
Fixes #2864.

`glz::minify_jsonc` and `glz::prettify_jsonc` silently dropped everything after a `//` line comment. `detail::read_jsonc_comment` only understood block comments: on `//` it skipped two bytes, scanned for a closing delimiter that never came, consumed the rest of the buffer and returned an empty view without setting an error. The formatters treated that empty view as nothing to write, so the caller was told nothing. A hand written `.jsonc` file that opens with a comment, which is how most of them start, minified to an empty string.

While reproducing it I found the same shape in a few neighbouring places, so the change covers them as well:

- `read_json_string` (both variants) had the identical "empty view, no error" flaw, so an unterminated string truncated output the same way, and `prettify_json` handed that empty view to `dump_not_empty`, which `memcpy`'d from a null pointer. That one is reachable from the public API, `prettify_jsonc(R"({"a":1} /* oops)")`, and UBSan reports it on master.
- `detail::minify_json` reported `syntax_error` on **every** non-empty input, valid or not, because its null terminated loop ended by classifying the padding byte. That made the error channel meaningless, so it now returns cleanly on the sentinel and reports only what it really parsed.
- `read_jsonc_comment` accepted the `*` of the opening delimiter as a closing one, so `/*/` looked complete and swallowed what followed.
- `skip_comment` in the read path did not report an unterminated block comment, so `validate_jsonc` accepted input such as `{"a":1} /* oops`.
- `skip_comment` also stepped two bytes over every star it did not close on, so a run of an even number of stars before the closing delimiter hid it: `{/***/ "s":"x"}` was rejected while `{/****/ "s":"x"}` was accepted. The formatters scan these correctly, so `minify_jsonc` emitted block comments that `read_jsonc` could not read back. That one came out of fuzzing the round trip rather than from reading the code.

Behaviour, so it is not a surprise:

- `minify_jsonc` drops line comments, because minifying is what removes the newline that terminates them, and preserves block comments.
- `prettify_jsonc` keeps both comment styles and puts the content that follows a line comment on a new line, so the comment cannot swallow it.
- An unterminated block comment now reports `error_code::expected_end_comment` instead of arriving at the value parser as absent input, which came back as the coarser `no_read_input`. One existing test pinned that answer (`"comment only input holds no value"`) and its expectation is updated.
- The reader no longer swallows a stray `*` `/` after a closed comment, and a carriage return ends a line comment on its own, so CR line endings no longer swallow the rest of a document.
- `minify_json`, `minify_jsonc`, `prettify_json` and `prettify_jsonc` gained overloads taking a `glaze::context&` that return `error_ctx`, for input that may not be well formed. The existing overloads keep dropping the error as before. The output of `prettify_json` is otherwise unchanged: the break before a closing bracket stays unconditional, so a whitespace only container keeps the blank indented line it always had.

The change is four commits, separable if you prefer them apart:

1. the read path (`skip_comment`: unterminated comments, star runs, CR, streaming gate) with its tests;
2. the formatters (`json_format.hpp` scanners, `minify.hpp`, `prettify.hpp`, docs) with their tests;
3. fuzz coverage for the comment enabled formatter paths, which the minify and prettify fuzz targets did not touch before, and for the null terminated branch that only a string selects.
4. the follow-ups from going back over it: the chunked scans no longer form their limit before the start of a short buffer, and minifying into a fixed-size buffer reports `buffer_overflow` instead of writing past it.

Testing: `json_test` and `json_test_non_null` are 745/745 with 61781 asserts each (GCC 15.2, and on master the same suites are 741/742 before the test updates). The new tests were also compiled with the project's own `-Wall -Wextra -pedantic -Werror` and run under ASan+UBSan. A systematic sweep of comment shapes (stars, slashes, spaces, CR/LF, nested delimiters) plus 50000 mutated documents reports no case where an input that validates produces a formatter error or a round trip that changes the value.

One note on CI, so a red check is not a surprise: the `msys2-mingw` job on my fork fails, and it is not this change. It has failed on all of its runs there, including the base commit `2518e757`, with `no matching function for call to std::from_chars(const char*&, const char*&, _Float128&)` from the existing float128 test suite; this diff contains no float128 or charconv code at all. That workflow does not run on `pull_request`, so it should not appear here.


wilkolbrzym-coder:
I went back over this change with a fresh eye and with ASan, and found two things worth fixing. Both are in `a14517cb`.

### `end - 7` in the chunked scans forms a pointer before the buffer

`read_jsonc_comment` and the unpadded `read_json_string` took their chunked scan limit as `end - 7` before checking that eight bytes were left, so for input shorter than that the pointer was formed before the start of the buffer. That is undefined even though the loop body never runs, and `/*/` reaches it. Both now take the limit only when a chunk is left, which is the condition the loop was already relying on:

```cpp
const auto chunk_end = (size_t(end - it) > 7) ? end - 7 : it;
for (; it < chunk_end;) {
```

so the scan visits exactly the same bytes as before and the loop bodies are unchanged.

### Minifying into a fixed-size buffer wrote past the end of it

The dumping in the minify token loop is unchecked, which is only sound while the output is no larger than the input, and nothing checked that the output could hold the input. `glz::minify_jsonc(ctx, std::string, std::array<char, 8>&)` on a 32 byte document is a stack buffer overflow; ASan reports a 12 byte write past the array through `dump`. The entry point now reserves through `ensure_space`, the mechanism the binary-to-JSON converters use for the same reason, so a bounded output that cannot hold the input reports `error_code::buffer_overflow`. Output into a `std::string` or a large enough `std::array` is byte-identical to before, and the token loop is untouched.

`prettify_json` writes into a bounded output unchecked as well, and I have **not** fixed that here: its output is not bounded by the size of its input (indentation and line breaks), so it needs a capacity check per token rather than one before the run. That is a change to the token loop, and it belongs in its own change rather than in this one. The same `end - 7` pattern also sits in `skip_until_closed` (`include/glaze/util/parse.hpp:1267` and `:1362`), outside this diff, and looks like it wants the same guard.

### Checked and deliberately left alone

`case Comment` in the minify loop falls through to `default:` only when `Opts.comments` is off; the `else` branch carries an explicit `[[fallthrough]]`, and with comments off a `/` is not a JSON token, so reporting `syntax_error` there is the intent.

### Small clarity edits

The test now says why `string_copy` exists (minifying pads its input in place, prettifying does not), and the comment on `line_comment_pending` says that a flag still pending when the input runs out is intended, so that nobody flushes it later and gains a trailing newline the tests do not expect.

wilkolbrzym-coder:
Follow-ups are pushed, five commits in total now. The last two:

- both chunked scans no longer form their `end - 7` limit before the start of a short buffer (`read_jsonc_comment` and the unpadded `read_json_string`);
- minifying into a fixed-size buffer reports `buffer_overflow` instead of writing past the end of it, which `minify_jsonc` into a `std::array<char, 8>` did;
- a line comment sitting immediately before a closing bracket is pinned by a test with its exact output.

@stephenberry this is ready for a look whenever you have a moment.

wilkolbrzym-coder:
Closing this.

#2881 covers the same ground, and its description credits the analysis — the star-run bug named, and
how it was found. What that description does not contain is a single word to the author of #2871, and
neither does this thread: two comments, both mine, and no reply in two days. Rewriting a report and
shipping it is a maintainer's call, and I am not arguing with the decision. Doing it silently, with
this PR still open, is the part I am not going to pretend did not happen.

That is not the reason it closes, though. It closes because it cannot merge: the branch is cut from
`2518e757` and has been `dirty` since #2872 rewrote the scanners it patches, and #2881 and #2882 are
rewriting those same lines today. Rebased, it would be a duplicate.

The branch stays as the reference. What is still open from the same audit is tracked on its own:
#2877, #2878, #2879.

stephenberry:
Thanks for this!

PR #2873, opened 2026-09-16
http_server: negotiate TLS 1.3 and expose the SSL context

michaelaeriksen:
## The problem

`http_server` builds its SSL context as `asio::ssl::context::tlsv12` ([http_server.hpp#L1500](https://github.com/stephenberry/glaze/blob/main/include/glaze/net/http_server.hpp#L1500)). asio implements that method as a generic `TLS_method()` context with **both** bounds pinned:

```cpp
handle_ = ::SSL_CTX_new(::TLS_method());
SSL_CTX_set_min_proto_version(handle_, TLS1_2_VERSION);
SSL_CTX_set_max_proto_version(handle_, TLS1_2_VERSION);
```

So a `glz::https_server` can only ever speak TLS 1.2. Measured against a running server before this change:

- `openssl s_client` reports `Protocol : TLSv1.2`
- a `-tls1_3` client is refused with `tlsv1 alert protocol version` (alert 70)

and there is no way to raise the ceiling from outside: `http_client` carries `configure_ssl_context` and `ssl_context_unsafe`, `http_server` carries neither, and `ssl_context` is private.

The negotiated suite was ECDHE-ECDSA-CHACHA20-POLY1305, so this is a missing floor rather than a weak connection.

## The change

**1. `tls_server` instead of `tlsv12`,** so the handshake settles on the highest mutually supported version. This mirrors `http_connection_pool`, which already constructs its context as `tls_client` for exactly this reason.

**2. The protocols RFC 8996 deprecates are disabled explicitly.** Worth flagging separately: the comment in `http_connection_pool` saying `tls_client` "automatically disables insecure protocols (SSLv3, TLS 1.0, TLS 1.1)" does not hold — asio sets the minimum to `TLS1_VERSION` for `tls_client` and `tls_server` alike, so whether TLS 1.0 is reachable depends on the OpenSSL build's own defaults rather than on the method. This PR does not rely on it for the server. The client may deserve the same treatment; I left it alone to keep this focused.

**3. Adds `configure_ssl_context` and `ssl_context_unsafe`,** mirroring the two methods `http_client` already has, for the settings the wrapper does not cover: ALPN, cipher suites, version bounds, and `SSL_CTX_set_cert_cb` for per-handshake certificate selection. Both are documented as "call before `start()`", since the server holds no lock around the context (unlike the pool, which has `ssl_mtx`).

## Tests

Four cases in `https_test`: a TLS 1.3 client completes the handshake, a TLS 1.2 client still does, TLS 1.1 does not negotiate, and `configure_ssl_context` reaches the context and chains. The probe lowers its own security level so that a refusal is attributable to the server rather than to the client declining to offer a protocol.

## Verification

macOS / clang / OpenSSL 3.6.4:

- With the change, `https_test` is **19/19**.
- Reverting **only** the context line back to `tlsv12` gives **18 passed, 1 failed** — exactly `server_negotiates_tls13`, with `unsupported protocol (SSL routines)`. So the new case actually pins the behaviour rather than passing either way.
- 13 TLS- and server-adjacent suites pass, including `wss_test`, `http_client_ssl_test`, `http_smuggling_test`, `http_header_strictness_test` and `http_server_api_tests`.
- `clang-format` with the repo's own `.clang-format` reports zero delta on the changed files.

Happy to split the accessors out from the version change if you would rather take them separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


stephenberry:
Thanks for this — the diagnosis is exactly right, and I appreciate that you separated what you verified from what you were inferring.

I reviewed it locally and reproduced your results, plus checked the one claim the tests don't cover.

**Confirmed.** `asio/ssl/impl/context.ipp` implements `tlsv12` as a `TLS_method()` context with both bounds pinned to `TLS1_2_VERSION`, so the server really could only ever speak TLS 1.2. Reverting only the context line back to `tlsv12` on your branch fails `server_negotiates_tls13` with `unsupported protocol` — the new case pins the fix rather than passing either way.

**Your correction to the `http_connection_pool` comment is right, and the `set_options` line is load-bearing.** asio sets the floor for both `tls_client` and `tls_server` to `TLS1_VERSION`, so the method does not disable TLS 1.0/1.1. What blocks them on a stock build is OpenSSL's security level, not the method. Dropping the server's security level to 0 to isolate it (macOS, OpenSSL 3.6.2):

| server config | TLS 1.1 |
|---|---|
| `tls_server` + your `set_options` | refused |
| `tls_server`, no `set_options` | refused |
| `tls_server`, seclevel 0, no `set_options` | **negotiates** |
| `tls_server`, seclevel 0, + your `set_options` | refused |

So the explicit options are what holds when a user lowers the security level — something `configure_ssl_context` now makes easy to do. Good call not assuming the method covered it.

**One gap worth knowing about:** `server_refuses_tls11` passes with *or* without the `set_options` line on a default OpenSSL 3.x build, since the security level already blocks TLS 1.1 there. It's still a fine regression guard for permissive platforms, but it isn't covering the RFC 8996 half of this change on the platforms CI actually runs. No action needed; just don't read a green there as proof that line works.

**I pushed one commit to your branch** (`ec9b1c96`), hope that's alright:

- `static_assert(EnableTLS, ...)` in `ssl_context_unsafe()`. It's a new public method, and on a plain `http_server<false>` it produced `indirection requires pointer operand ('std::monostate' invalid)` rather than a diagnosis. Cheaper to add now than after the API ships. (Worth noting the assert only fires when the method is called — member bodies of a class template instantiate lazily — so `http_server<false>` still compiles clean at `-Wall -Wextra` when it isn't used.)
- A docs section in `docs/networking/tls-support.md` for both new methods plus protocol-version behavior, since the new public API had no documentation. **That prose is mine, so please rewrite anything that doesn't read the way you'd put it.**

Verified on the merged result: `https_test` 19/19, plus `wss_test`, `http_client_ssl_test`, `http_server_api_tests`, `http_smuggling_test` and `http_header_strictness_test` all pass; `clang-format` clean.

On splitting the accessors out — I'd keep them together. The version fix alone leaves no way to set ALPN, cipher suites, or custom version bounds, and the accessors are what makes that reachable.

I'll take the `http_client` side as a follow-up, including the comment you flagged. Merging once CI is green.

wilkolbrzym-coder (commented):
Review – PR #2873

Diagnosis is precise, the change is well scoped, and the two new accessors mirror the client API cleanly. No blocking defect, but two items should be closed before merge.

Required before merge

1. `configure_ssl_context` is a silent no-op when `EnableTLS` is true while `GLZ_ENABLE_SSL` is undefined. The callable is never invoked, yet the method still returns `*this` and reports success. The constructor carries a `static_assert`, but the accessor itself should carry one as well, matching the guard already present on `ssl_context_unsafe`.

2. `server_refuses_tls11` passes independently of `no_tlsv1_1` on a default OpenSSL 3.x build, because the security level already blocks TLS 1.1. The RFC 8996 half of this change is therefore not exercised on the platforms CI actually runs on. A dedicated case that lowers the server's security level and confirms TLS 1.1 still does not negotiate would close the gap.

Recommended, non-blocking

- Nothing enforces "call before `start()`". An atomic started flag with an assert in debug builds would catch misuse at negligible cost.
- `ssl_context_unsafe` is declared inside `#ifdef GLZ_ENABLE_SSL`, so misuse on `http_server<false>` yields "no member named" rather than the intended `static_assert`. Moving the declaration outside the guard and letting the assert be the sole gate improves the diagnostic.
- The negotiation ceiling of `tls_server` depends on asio leaving `max_proto_version` unset. Document the minimum asio version this was verified against.
- The same comment correction about `tls_client` applies to `http_connection_pool`. Already agreed as a follow-up.

CI

Five of sixty-two checks have completed at this point. The full matrix must be green before merge.

LGTM once item 1 carries a `static_assert` and item 2 is either covered by a dedicated test or explicitly acknowledged as uncovered on default OpenSSL.

cc @stephenberry

stephenberry:
@wilkolbrzym-coder Thanks for looking. I've taken item 2. The other required item and one of the recommendations don't hold, and I'd rather explain why than leave them open.

**Item 2 — taken.** Pushed in `66c089a5`. It lowers the server's security level to 0 through the new `configure_ssl_context`, then confirms TLS 1.1 still fails to negotiate, so the only thing standing is the explicit `no_tlsv1`/`no_tlsv1_1` options. Mutation-checked both ways: with the `set_options` line it's 20/20, and removing only that line fails exactly the new case. It now pins the RFC 8996 half of the change on a default OpenSSL build, which was the real gap. Nice catch on wanting it enforced rather than acknowledged.

**Item 1 — the state described isn't reachable.** `EnableTLS == true` with `GLZ_ENABLE_SSL` undefined already fails in the constructor, at `http_server.hpp:1511`:

```
error: static assertion failed due to requirement '!true':
TLS support requires GLZ_ENABLE_SSL to be defined and OpenSSL to be available
```

There's no way to obtain a constructed server in that configuration, so `configure_ssl_context` can't be called in it and no silent no-op occurs. A second `static_assert` would be unreachable code.

For what it's worth, `configure_ssl_context` *does* silently no-op in a different case you didn't mention — `EnableTLS == false` on a plain `http_server`. That one is reachable, and I'm leaving it, because `load_certificate` and `set_ssl_verify_mode` have behaved that way since they were added and diverging for one method would be worse than the inconsistency.

**Recommendation on `ssl_context_unsafe` — this one would break the build.** The premise is that the `#ifdef GLZ_ENABLE_SSL` causes "no member named" on `http_server<false>`. It doesn't: that's a per-TU preprocessor gate, not a per-instantiation one. With `GLZ_ENABLE_SSL` defined — the only build where TLS exists at all — calling it on `http_server<false>` gives the intended assert:

```
error: static assertion failed: ssl_context_unsafe() requires a TLS-enabled server
(glz::https_server, or glz::http_server<true>); http_server<false> holds no SSL context
```

"No member named" appears only when `GLZ_ENABLE_SSL` is undefined, i.e. a build with no SSL compiled in, where that is the correct diagnostic.

Moving the declaration out of the guard is actively harmful. `<asio/ssl.hpp>` is included only under `GLZ_ENABLE_SSL` (`glaze_asio.hpp:36-38`), so hoisting it puts `asio::ssl::context&` in the class body where the type is undeclared. I applied the suggestion and compiled an ordinary non-SSL server:

```
error: no member named 'ssl' in namespace 'asio'
error: use of undeclared identifier 'ssl_context'
```

That breaks every non-SSL build of glaze, which is the default configuration.

**On the rest.** The started-flag assert is a good idea and `std::atomic<bool> running` already exists at line 2389, so it's cheap — I'll do it separately rather than widen this PR. Documenting a minimum asio version I'll skip; `FindAsio.cmake` sets no minimum today and the `tls_server` mapping has been stable for years, so a version number in the docs would go stale faster than it would help. The `http_client` correction is already queued as a follow-up.

One request: the check count in your review ("five of sixty-two have completed") wasn't accurate when you posted — five were queued and none had finished. Small thing, but I'd rather the review record match what the run actually showed.

Full matrix has to be green before this merges regardless.

wilkolbrzym-coder:
@stephenberry Hold off on merging for a moment.

Something in your reply doesn't add up to me. Let me check it before this goes in.

wilkolbrzym-coder (approved):
**Approved.** I found no defect in this change: every claim in the description holds when
compiled, both new guards are load-bearing under mutation, and I could not break the accessors or
the non-SSL build with them.

**Item 2 — verified, and the guard is real.** Built the branch here (Linux, GCC 15.2,
OpenSSL 3.5.5, standalone Asio 1.36.0, direct `g++`, no CMake): 20/20. Removing only
`set_options(...)` (`http_server.hpp:1506`) fails exactly
`server_refuses_tls11_without_security_level_backstop` with `got ' TLSv1.1 '`; reverting only the
context line to `tlsv12` fails exactly `server_negotiates_tls13`. A raw OpenSSL client offering
only TLS 1.3 — no Asio on the client side — completes the handshake as `TLSv1.3`, so the fix is
not an Asio artefact.

**Item 1 — withdrawn.** Without `GLZ_ENABLE_SSL`, `glz::https_server server;` fails at
`http_server.hpp:1511` with the message you quoted, so there is no constructed object and no
reachable silent no-op. Treating `http_server<false>` the way `load_certificate` and
`set_ssl_verify_mode` already behave is the right call.

**The `ssl_context_unsafe` recommendation — withdrawn.** Applied, it breaks a non-SSL build at the
declaration itself (`'ssl' in namespace 'asio' does not name a type`, `<asio/ssl.hpp>` being
guarded), i.e. it would break the default configuration to improve a diagnostic. The guard stays,
and the intended assert does reach the user wherever TLS exists.

**Check count — corrected, in both directions.** My review was posted at 20:32:47Z against
`ec9b1c96`. Four checks had finished by then, all green: `rpm-build:fedora-rawhide-x86_64`
20:03:30Z, `-i386` 20:03:31Z, `-ppc64le` 20:06:05Z, `-aarch64` 20:06:09Z. Fifty-eight had not
started, and the fifth (`-s390x`) finished at 20:42:06Z. So "five" was mine by one, and "five were
queued and none had finished" does not match the run either. Re-runs only rewrite `completed_at`
later, never earlier, so the count at that moment is at least four.

**Outside this change — pre-existing, verified on `main` at `2518e757`, none blocking. Listed in
order of severity, line numbers on this branch.**

1. **TLS connections close without `close_notify`** — `http_server.hpp:2512`, `:3181`, and
   `websocket_connection.hpp:1261` (FIN only; no SSL-level shutdown anywhere in the file).
   Observed from a plain OpenSSL client: `SSL_read` returns 0 with `unexpected eof while reading`
   and `SSL_RECEIVED_SHUTDOWN` unset, and the client's decode-error alert comes back through
   `error_handler` as `tlsv1 alert decode error`. Truncation is indistinguishable at the TLS
   layer, and a normal disconnect is logged as an error. Version-independent; this change only
   makes it the default path for TLS 1.3 clients.
2. **`detail::close_socket` does wait for the peer** — the comment at `http_client.hpp:408` says
   it does not; Asio's `engine::do_shutdown` calls `SSL_shutdown` twice
   (`asio/ssl/detail/impl/engine.ipp:357`). Measured against a peer holding the socket open:
   `clear_connection_pool()` 5.00 s over TLS 1.2 and 4.96 s over TLS 1.3, returning exactly when
   the peer closed; 0.00 s with `set_graceful_ssl_shutdown(false)`. A peer that neither closes nor
   sends `close_notify` blocks the caller for as long as it chooses.
3. **`http_client.hpp:885` (same at `websocket_client.hpp:89`)** — the comment stating
   `tls_client` "automatically disables insecure protocols (SSLv3, TLS 1.0, TLS 1.1)" is false:
   the floor is `TLS1_VERSION` and the ceiling is unset (measured 769/0, against 771/771 for
   `tlsv12`). On stock OpenSSL 3 the security level masks it, so this is a configuration-dependent
   gap rather than an immediate one. Your queued follow-up, measured so it does not get lost.
4. **`http_server.hpp:1476`** — the constructor example in the doc comment does not compile: with
   an empty braced-init-list both constructors are viable with identity conversion, so the call is
   ambiguous. `asio::any_io_executor{}` in place of `{}` works.

**One note, not a defect.** `negotiated_protocol()` returns an empty string on refusal *and* when
the version cannot be pinned or an exception is thrown, and the refusal cases assert only
`empty()`, so they would pass vacuously on a platform built without TLS 1.1. The positive cases
keep the pair honest, as the mutations show.

PR #2874, opened 2026-09-16
net: raise the TLS floor on the client side, assert SSL context is configured before start()

stephenberry:
Follow-ups to #2873, which fixed the TLS floor and ceiling on `http_server` but left the client side and the "call before `start()`" contract alone.

## The client side had the same missing floor

`http_connection_pool` and `websocket_client` both build their context as `tls_client`, and its comment claimed that "automatically disables insecure protocols (SSLv3, TLS 1.0, TLS 1.1)". It does not: asio sets the floor for `tls_client` to `TLS1_VERSION`, exactly as it does for `tls_server`. What rejects the deprecated protocols on a default build is OpenSSL's security level, not the method — and that backstop disappears the moment a caller lowers the level through `configure_ssl_context`.

Both contexts now disable the RFC 8996 protocols explicitly, matching what `http_server` does, and the comment is corrected.

## `configure_ssl_context` / `ssl_context_unsafe` now assert they run before `start()`

Both are documented as "call before `start()`" because the server holds no lock around the context, so touching it once the acceptor is running races with the handshakes in flight. Nothing enforced that. A debug assert on the existing `running` flag catches the misuse at zero release cost.

This is the first runtime `assert` under `include/glaze/net/`, so it is a small precedent; `<cassert>` is already used elsewhere in the library.

## Tests

`client_refuses_tls11`, with `client_negotiates_tls12_against_permissive_server` as a control so a refusal is attributable to the version rather than to the probe server failing to start. The probe spins a deliberately permissive TLS server that offers exactly one protocol version, and lowers the *client's* security level to 0 so that the explicit context options are the only thing under test.

That last part matters: the first version of this test passed with or without the fix, because the client's own default security level already rejects TLS 1.1. It only pins the change once that backstop is removed — the same trap the `server_refuses_tls11` case in #2873 fell into.

## Verification

macOS / clang / OpenSSL 3.6.2, clean Debug build so the new asserts are live:

- Full suite **123/123**.
- Mutation-checked: removing only the `set_options` line from `http_connection_pool` fails exactly `client_refuses_tls11`, and nothing else.
- `-DNDEBUG -Wall -Wextra` clean, and a non-SSL `http_server<false>` build is clean.
- `clang-format` reports no delta on the changed files.


wilkolbrzym-coder (approved):
lgtm

PR #2875, opened 2026-09-16
docs: Add documentation on blocking stream requests

crueter:
It may be useful to have a stream request block the thread until
completion, so this adds docs thereof using `std::promise`.

Signed-off-by: crueter <crueter@eden-emu.dev>


wilkolbrzym-coder:
A page that tells users to block on a stream, in a library whose streaming callbacks do not fire exactly once. I ran both halves of the pattern below on `0b9376db`, before any of this PR's text, and it hangs on one path and throws on the other.

**It blocks forever on the immediate-failure paths.** `stream_request_v2` returns `nullptr` when the URL does not parse (`http_client.hpp:1551`) and when `https` is requested without `GLZ_ENABLE_SSL` (`:1747`). On both of those only `on_error` is posted — `on_disconnect` never runs — and the example sets the promise only in `on_disconnect` while never inspecting the returned handle. Running the documented pattern verbatim against a bad URL:

```
handle is null: 1
[callback] on_error: Invalid argument
now blocking on promise.get_future().get(), exactly as the docs do...
exit=124        (killed after 5 s; nothing ever sets the promise)
```

A null-handle check, or setting the promise from `on_error` as well, closes that hole. As written, the first thing a user meets after a typo in a URL is a deadlock.

**Even where the callbacks do fire, they fire twice.** Client timeout of 1 s against a peer that accepts and never answers, three consecutive runs, identical counts every time:

```
[callback] on_error: Operation aborted.        <- a cancellation reported as an error
[callback] on_error: Connection timed out
[callback] on_disconnect call #1
[callback] on_disconnect call #2
[callback] promise.set_value() threw: std::future_error: Promise already satisfied
```

Two `on_error` and two `on_disconnect` per request; only the order of the two errors varies between runs, because the client's workers run on more than one thread and neither callback has an at-most-once guard. Uncaught, that `future_error` leaves the io thread and terminates the process. So the documented wrapper has to defend against the library's own callback contract — the opposite of what a documentation example should be teaching.

**The defects are in the library, not on this page.** `on_disconnect` firing more than once, and `operation_aborted` reaching `on_error` on the connect/header path (`stream_read_failure_is_error` filters it; that path never consults it). Fix those two and a blocking recipe becomes safe to document. Until then this page hands users a deadlock and an aborted io thread, and they will arrive here with both.

**My position: close this one.** For callers who do not need incremental delivery, a future already exists — `get_async()`/`post_async()` return `std::future<std::expected<response, std::error_code>>` and block on `.get()`. For callers who do need streaming, a `std::promise` is a handful of lines they can write with their own null-then-error discipline. What this page adds is a sixty-line recipe that does not work, and what would make it work is a library change that belongs in its own issue. If anything is kept, it should be a short note — check the handle, set your promise from both `on_error` and `on_disconnect` — not this.

One thing for whoever reviews the pair: this example's `on_data` returns `void`, and #2876 changes that same handler to `bool`. Both PRs edit this file from the same base, so whichever lands second needs a rebase and a matching signature.

@stephenberry

crueter:
Turns out dead internet theory is real.

PR #2876, opened 2026-09-16
http_client: Allow `on_data` to cancel stream requests

crueter:
Like `on_progress` it now returns a bool. If that bool is false then the request is cancelled early. I also converted it to a wrapper which can accept any return type, so existing consumers don't get their code randomly broken.

I didn't add unit tests because that is a lot of effort that I'm simply
not familiar with. Feel free to add as needed.

Signed-off-by: crueter <crueter@eden-emu.dev>


wilkolbrzym-coder:
Seven lines of production change, pushed without compiling the test suite once. That is the whole story of this PR.

**It does not compile the repository it is proposed to.** `http_data_handler` is a public alias, so changing it from `void(std::string_view)` to `bool(std::string_view)` breaks every existing callback — not only third-party ones. `std::function` returning `bool` cannot be constructed from a lambda returning `void`:

```
tests/networking_tests/http_client_no_ssl_executor_test/http_client_no_ssl_executor_test.cpp:63:46:
  error: no match for 'operator=' (operand types are 'glz::http_data_handler'
  {aka 'std::function<bool(std::basic_string_view<char>)>'} and '<lambda(std::string_view)>')
  /usr/include/c++/15/type_traits:144:11: error: no type named 'type' in
  'struct std::enable_if<false, std::function<bool(std::basic_string_view<char>)>&>'
```

Every networking test that uses `on_data`, compiled on `0b9376db` and on this branch (GCC 15.2, C++23):

```
ok   [base] http_client_no_ssl_executor_test        FAIL [pr]  2 errors
ok   [base] http_client_redirect_test               FAIL [pr]  1 error
ok   [base] http_client_progress_test               FAIL [pr]  2 errors
ok   [base] http_client_test                        FAIL [pr]  5 errors
ok   [base] http_chunked_test                       FAIL [pr]  5 errors
```

`standalone-asio` and `boost-asio` both build `http_client_test` as an explicit target, so this goes red the moment the workflows are approved. The five green checks currently on this PR are the `rpm-build` packaging jobs — they never compile the tests, so a green check here says nothing about whether the change builds.

**What is right, on the record.** The mechanism itself is implemented properly: all three delivery sites go through `deliver_stream_body`, `false` tears the connection down exactly the way `on_progress` already does — socket closed rather than returned to the pool, `on_disconnect` fires, the read loop unwinds — and the comment above `http_progress_handler` already documents precisely that behaviour. Cancellation is not half-wired; the problem is the compatibility cost, not the design.

**The capability already exists without an API change.** `on_progress` runs after every span handed to `on_data`, and its documented `false` path is the same teardown, so a flag set in `on_data` plus `return !flag;` in a three-line `on_progress` cancels on the same chunk. The break therefore buys ergonomics, not capability — which matters when the price is a source-breaking change to a public alias that every existing user of this library pays.

**If the bool is wanted**, either make the alias accept both shapes with a thin wrapper (a `bool`-returning callable as-is, a `void`-returning one wrapped to return `true`) so existing code keeps compiling, or take the break deliberately — which then means updating the repository, adding a test that actually exercises cancellation, and finishing the conversion in the docs, because `docs/networking/http-client.md:358` in this very PR still shows a `void`-returning `on_data`.

**My position: request changes, not close — yet.** The want is legitimate and the implementation is sound, so this deserves one round. It cannot land in this state, and it must not be waved through if it goes quiet: the same behaviour is reachable today through `on_progress`, so there is no pressure to accept a breaking change to get it. If it is not rebuilt with a green networking build within a reasonable window, close it — that outcome costs the library nothing.

@stephenberry

crueter:
Yeah I'm fine without some snarky HR lady-esque LLM giving me a 5 paragraph explanation of things I already knew.

wilkolbrzym-coder:
Man; if you knew about this, you wouldn’t have even opened this PR. Close the PR.

wilkolbrzym-coder:
> Man; if you knew about this, you wouldn’t have even opened this PR. Close the PR.

Either you close it or you fix it and put it in order.

stephenberry:
@crueter, this is a good request. It would be better as an issue than a PR, but it is helpful to have some basic code to look at, so thanks. I'll leave this PR here for reference, in the future you can open issues.

@wilkolbrzym-coder, no need to ask to kill PRs. And, try to focus on the core of the request. Iteration on ideas is a good thing. The LLM just saw something incomplete and considered it useless, but that's not the case.

PR #2891, opened 2026-09-17
Name the member a writer or reader cannot handle, in every format

wilkolbrzym-coder:
A struct with one member a format cannot write fails from inside the member loop, where the only clue is

```
include/glaze/json/write.hpp:2729:65: error: incomplete type 'glz::to<10, Opaque>' used in nested name specifier
```

an opaque format id, no member, and a couple of hundred lines of instantiation frames around it. CSV is worse: the same struct produces 530 lines and 15 errors, none of which names the member.

I am replacing #2610 with this, which does the same thing correctly and for every format.

### What the error is now

| format | before | after |
|---|---|---|
| JSON | 159 lines, 1 misleading error | 30 lines, 1 error |
| YAML | 195 lines, 1 misleading error | 29 lines, 1 error |
| CSV | 530 lines, 15 errors | 29 lines, 1 error |

The member is named in the instantiation trace, which is the only place a compile-time diagnostic can carry a name:

```
In instantiation of 'struct glz::detail::unsupported_writer<Outer, 10, 1, glz::string_literal<2>{"o"}, glz::string_literal<7>{"Opaque"}>':
```

object `Outer`, format 10 (JSON), member 1 of `glz::reflect<T>`, key `o`, type `Opaque`. The key is the one the output uses, so a member renamed through `glz::meta` is reported under the name it has in the document.

### Where the check lives, and why

Inside the body of `op()`, not at class scope on `to<Format, T>`. `write_supported` is `requires { to<Format, T>{}; }`, so an assert in the class body fires while that probe is being evaluated, turning a query generic code depends on into a hard error. The convention is already recorded in the library, on the variant writers in jsonb and yaml: *"Inside op() rather than at class scope: write_supported is `requires { to<Format, T>{}; }`, so a class-scope assert turns that feature probe into a hard error instead of `false`."* `op()` is a member template, so an assert in its body is invisible to the probe and is only instantiated when the writer actually runs.

#2610 had it at class scope. On that branch, for a struct with one unsupported member,

```cpp
static_assert(!glz::write_supported<Outer, glz::JSON>);
```

reports *"One of the object's members is not serializable"* instead of answering.

### Which formats

JSON, JSONB, BEVE, CBOR, MSGPACK, BSON, YAML, TOML, CSV and EETF. BEVE checks both of its layouts, MSGPACK both its `glaze_object_t` path and its `reflectable` path (which forwards through `to_tie`). NDJSON needs nothing of its own: it writes records through the JSON writer, and a record with an unsupported member is reported through it.

Two things the check has to know about, or it reports members that a format writes:

- **BSON** emits nullable, null and variant members itself, inside its member loop, because a BSON element carries its type next to its key. `write_supported<std::optional<int>, BSON>` is false while BSON writes the member, so the format declares those members (`writer_emits_member_inline`). The counterpart on the read side is not needed: the reader has real `from<BSON, ...>` specializations for all three.
- **Function pointers** are dropped by some formats and written by others (JSON and BSON drop both kinds, JSONB, BEVE and CBOR drop both, TOML drops member function pointers, and YAML, MSGPACK and CSV drop neither). The check exempts them everywhere rather than tracking that table, which is the lenient direction: it means such a member is never *reported* by this diagnostic, not that every format writes it.

The check also honours what an operation skips by design: `meta::skip` for that operation, so a write-only or read-only field is not reported for the direction it is skipped in; `glz::skip`, `hidden` and `includer`; and members the reflection does not carry.

### Behaviour changes

- `to<EETF, T>`, the least specialized specialization, is still there, and now says why a type has no EETF writer instead of leaving the dispatcher to report a missing `op`. `write_supported<T, EETF>` is unchanged -- it answers true for every type, before and after.
- Nothing else: a struct every format can write produces the same bytes.

### Commits

1. the mechanism in `core/reflect.hpp`
2. the check in the object writer and reader of every format
3. EETF's blanket specialization: the message it was missing
4. the test suite: the accepted cases, and twenty programs that must be rejected
5. the two defects the suite found, fixed (below)
6. the function-pointer exemption stated where it does the work
7. a comment where the key label cannot be produced

The last three are what came out of running the repository's own suite over the first four. Both of the defects in commit 5 were real and would have shipped:

- `bson_test` stopped compiling, on `optional_s`, `maybe_int_s`, `monostate_field_s`, `int_or_string_s`, `id_or_time_s` and `scalar_or_array_s` -- the false positive above.
- `msgpack_test` stopped compiling on the structs it declares inside `main()`. The check asked for the member count and member types through `reflect<T>`, which for a reflectable type declares `static constexpr auto keys = member_names<V>`, and deducing `auto` there evaluates the member names -- a hard error in GCC for a type with no linkage, because the pre-C++26 reflection passes the type through `external<T>`, which cannot be defined for a type declared inside a function. MSGPACK's reflectable path never touches the names, which is why it writes those types today. The count and the types now come from the tie.

### What was tested

- `json_test`: 737 tests, 61,712 asserts, in both configurations (`GLZ_NULL_TERMINATED` true and false). Also yaml 813/3124, toml 419/1515, cbor 316/1273, jsonb 163/679, csv 127/624, bson 124/290, msgpack 70/251, beve. Built and run directly with g++ 15.2, `-std=c++23 -O1`; the library's own flags (`-Wall -Wextra -pedantic`, `-Werror` in Debug) are clean.
- The new suite: `diagnostics_test` in both configurations, and twenty programs that must be rejected, one per format and per direction plus a nested case and a renamed-key case. Each carries the fragments its output has to contain; `check_diagnostic.cmake` compiles it and reads the message, so a format whose check stops firing fails the suite.
- Mutation testing, all six detected: predicate always true, the assert moved to class scope, the key label emptied, `meta::skip` ignored, function pointers diagnosed, BSON's exemption removed.
- EETF could not be run: it needs Erlang's `ei.h`, which is not installed here. Its test translation unit is syntax-checked, and its write path is the one place the member check cannot report, because of the blanket specialization.

### Note on the fork's CI

The `msys2-mingw / build` job on my fork is red for a reason unrelated to this change: MinGW's libstdc++ has no `std::from_chars`/`std::to_chars` overloads for `_Float128`, which the pre-existing `float128_test` needs. It fails the same way on the base commit, and `msys2.yml` has no `pull_request` trigger, so it does not run here at all.

### Two follow-ups from going back over the branch

- The question is about the member's own type, and that boundary is now written down in `core/reflect.hpp` where the predicate is: a member whose type *contains* a type with no writer -- `std::vector<Opaque>`, `std::optional<Opaque>` -- answers as supported, because `to<Format, ...>` for the container exists, and the failure then comes from inside the container's writer, as it did before this change. Reporting that case would mean instantiating the container's writer, which is the cost the check exists to avoid.
- `two_bad_writer` pins which member the report names when a type has more than one problem. Nothing covered it: every shape had a single unsupported member, so a fold keeping the last index instead of the first passed the whole suite (6/6 in both configurations) and all twenty cases. The new case requires the trace to name index 0 and the key `first_bad`, and with the same mutation its fragments are now missing.

The limits above are now also written down for users: `docs/compile-time-diagnostics.md` (linked from the mkdocs nav next to Exceptions) has the message and how to read the trace, the per-format coverage, what is deliberately not reported, and the reminder that a format has to ask for the check by hand.
- TOML had the same false positive BSON did, one format further along: its writer emits an always-null member as nothing, so `struct S { int a{1}; std::monostate m{}; };` compiled and wrote `a = 1` on `main` and was rejected on this branch. Fixed the same way and limited to the write side, because TOML's reader genuinely cannot read those members; `TomlInlineMembers` in the suite pins it, and removing the declaration fails three assertions.


stephenberry:
How does this affect compilation times?

wilkolbrzym-coder:
I measured it. GCC 15.2, one translation unit at a time on an otherwise idle 4-core box, the same TU and the same flags on both sides with only `-I` pointing at the other tree: base `96cc2a39` (this branch's merge base) against head `c8905c96`. The two compiles of a pair run back to back and the whole run is interleaved base/branch/base/branch, so drift lands on both sides. Command line: `g++ -std=c++23 -O2 -fno-exceptions -fno-rtti -Wall -Wextra -pedantic -I <tree>/include -c <tu> -o /dev/null`.

Ordinary TUs, median of 5 runs each -- the difference is smaller than the spread between two runs of the same compile (1-3 %):

| TU (one compile) | base | branch | delta |
| --- | --- | --- | --- |
| 1 struct, JSON | 5.02 s | 4.94 s | -0.08 s |
| 9 structs, JSON, 2 option sets | 13.45 s | 13.48 s | +0.03 s |
| a 64-member struct and a struct of 17 sub-structs | 10.61 s | 10.61 s | +0.01 s |
| 1 struct, 8 option sets | 9.84 s | 9.81 s | -0.04 s |
| recursive and mutually recursive types | 6.96 s | 6.97 s | +0.01 s |
| 3 structs across 9 formats (JSON, YAML, TOML, CBOR, BEVE, MSGPACK, BSON, JSONB, CSV) | 27.25 s | 27.35 s | +0.09 s |

The repository's own TUs, where the instantiation load is far above any user's (1-2 compiles per side, each takes minutes):

| suite | -O2 base | -O2 branch | -O0 base | -O0 branch |
| --- | --- | --- | --- | --- |
| csv_test | 23.92 s | 23.91 s | 13.10 s | 13.24 s |
| bson_test | 17.93 s | 17.97 s | 11.88 s | 11.89 s |
| msgpack_test | 32.00 s | 32.33 s | 11.93 s | 12.42 s |
| jsonb_test | 25.90 s | 25.90 s | 16.46 s | 15.82 s |
| cbor_test | 49.83 s | 49.98 s | 19.06 s | 19.09 s |
| toml_test | 57.58 s | 57.79 s | 25.30 s | 25.52 s |
| beve_test | 117.90 s | 118.23 s | 46.88 s | 46.77 s |
| yaml_test | 234.42 s | 235.09 s | 63.44 s | 63.93 s |
| json_test | 216.06 s | 215.97 s | 71.49 s | 71.97 s |
| json_test, `GLZ_NULL_TERMINATED=false` | 225.59 s | 225.99 s | -- | -- |

The largest delta anywhere is +0.7 s on a 234 s compile, +0.3 %. `json_test`, the largest TU here, comes out 0.1 s faster on the branch. The sign is not reliable at this size, which is what "no measurable cost" looks like.

Where the work lands, measured directly rather than inferred:

- `-ftime-report` on the 27 s multi-format TU: template instantiation 2.04 s -> 2.11 s; every other phase unchanged.
- `-fstats` on `json_test` at -O1, the type hash, which is where instantiations are counted: **268,263 -> 273,158 elements, +1.8 %**, with the instantiation phase 18.20 s -> 18.60 s of a 152 s compile.

That is the shape of the check: a single pass over a type's members per (operation, format, type), whatever the number of members, and evaluated once however many option sets instantiate the same writer. The eight-option-set probe above is that last claim measured -- it shows no delta at all.

Generated code is unaffected. On the multi-format TU at -O2 the object file has the same 526 defined symbols -- none added, none removed, none resized -- and differs by 4 bytes of section data out of 1.18 MB. The check is `consteval`: it runs at compile time and emits nothing.

And the new test target this PR adds: `diagnostics_test` compiles in 18.5 s at -O2 (8.5 s at -O0), and the twenty rejection cases are 3.1-4.1 s each -- the same as before the change, because they are rejected on `main` too. What changes is the output: 3030 lines across the twenty cases become 878, e.g. CSV 530 lines and 15 errors -> 28 lines and 1 error.

So: no, it does not lengthen compilation in anything I can measure, and on the path it exists for -- a build that should fail -- it fails in the same time with a quarter of the output. (Status of the head, for completeness: every job on my fork is green except `msys2-mingw`, which fails on `main` there too, 13 runs out of 13, on MinGW's missing `_Float128` charconv overloads.)


---

One more thing, on status: this is not ready yet. I rebased the branch onto current `main` a moment ago and I am still fixing a few things, so it is not worth a full review yet -- I will write here and mark it ready when it is.

wilkolbrzym-coder:
One thing about the fork's CI, separately from the numbers above: I am not fixing the `msys2-mingw` job in this pull request. It is red on `main` there as well, for two reasons that have nothing to do with this change -- MinGW's libstdc++ has no `_Float128` overloads in `<charconv>`, which the pre-existing float128 test needs, and once that test compiles, `json_test` still fails to link because `hostname_include.hpp`'s `#pragma comment(lib, "ws2_32.lib")` is MSVC only. I have both fixes on a separate branch and will send them as their own change.

PR #2895, opened 2026-09-18
Reflect inherited members in the C++26 path, base members first (#2852)

### LINKED ISSUE #2852: Reflection behavior of extended structs
When serializing extended structs with the new c++26 reflection it only exports the additional (or direct) members of that object instead of all members as inherited by its base class(es).

This could be intended, but it seems weird and counterintuitive as the struct does in fact contain all the members but is only exported to a subset.

```cpp
#include <meta>
#include <print>
#include <glaze/glaze.hpp>

consteval std::vector<std::meta::info> get_all_members(std::meta::info n)
{
    auto val = std::meta::nonstatic_data_members_of(n, std::meta::access_context::unchecked());
    for (auto x : std::meta::bases_of(n, std::meta::access_context::unchecked()))
    {
        val.append_range(get_all_members(std::meta::type_of(x)));
    }
    return val;
}

struct User
{
    std::string baseMember;
    int baseMember2;
};

struct AdminUser : User
{
    int extendedMember;
};

struct MegaUser : AdminUser
{
    int secondExtendedMember;
};

int main()
{
    std::println("{}", glz::write_json(MegaUser{}).value_or("error"));
    // {"secondExtendedMember":0}
    template for (constexpr std::meta::info member : std::define_static_array(get_all_members(^^MegaUser)))
    {
        std::println("{}", std::meta::identifier_of(member));
    }
    // Should print:
    // secondExtendedMember
    // extendedMember
    // baseMember
    // baseMember2
}
```

`get_all_members` recursively iterates the base classes to collect all members.
There is some funny business going on when using diamond inheritance and repreated inheritance.
Members duplicate ...,  but you could remove duplicates or disallow that (already is?).

Tested on gcc 16.2.1 with -std=c++26 -freflection

wilkolbrzym-coder:
A derived type lost every member it inherits: `struct MegaUser : AdminUser : User` was written as `{"secondExtendedMember":0}` and `glz::member_names<MegaUser>` returned one name, while `docs/p2996-reflection.md` promised the opposite. This fixes the C++26 reflection so the members of base classes are reflected too, base members first, and documents the shapes that still need a `glz::meta`.

**Root cause.** `count_members`, `member_info_at` and `get_member_name_p2996` each asked `std::meta::nonstatic_data_members_of`, which answers with the members a class declares itself and says nothing about its bases; nothing walked `std::meta::bases_of`. The failure was silent: the type compiled and serialized a subset of its state, and both the comments and the documentation claimed the members were included.

**The change.** `include/glaze/reflection/to_tuple.hpp` derives the member list from one helper, `detail::all_members_of`, which walks the bases depth-first before taking the type's own members. The three call sites above use it, so the member count, the tie indices and the key names stay in step:

- base members come first, in declaration order, then the members the type declares itself, which is the order the documentation already described;
- a base reached through more than one path contributes once, so a virtual diamond (a single shared subobject) is reflected once instead of once per path;
- private bases and private members arrive through the `access_context::unchecked()` the reflection already used;
- a type with no base classes takes the previous single query, so the common case does not pay for the walk.

**Behaviour changes.**

- A derived type without a `glz::meta` now writes and reads its base members too, which is the fix, in every format, including the array-shaped writes (`structs_as_arrays`, `glz::reflect_array`), whose element count comes from the same member list.
- Two shapes still need a `glz::meta`, now documented in `docs/p2996-reflection.md`: two members with one name, which is a member that hides a base member or two bases that repeat a name (the writer emits that key twice while the reader cannot be instantiated for the type at all), and a base inherited twice *non*-virtually, where the compiler rejects the ambiguous access with `'NA' is an ambiguous base of 'ND'`. A `glz::meta` written for a base class is also not consulted for a derived type that is reflected automatically, because the reflection reads the base's data members, which is why the same base serializes as `{"renamed":1}` on its own and as `{"raw":1,"extra":2}` inside such a type.
- The traditional reflection is untouched: it refuses a derived type whose base has data members, with the same error before and after this change.

**Commits**

1. `list the members of base classes too, base members first` - the helper and the three call sites
2. `pin the inherited members, a virtual diamond and a private base with tests` - `tests/p2996_test`
3. `document inherited members, and the shapes that still need a glz::meta` - the documentation and a stale comment in the C++26 test

**Testing** (GCC 16, `-std=c++26 -freflection -DGLZ_REFLECTION26=1`):

- `tests/p2996_test`: 26 tests / 169 asserts pass; eight of them are new and cover member names for a two-level hierarchy, JSON and BEVE round trips, a virtual diamond, a private base, a polymorphic base, an empty base, and the `glz::meta` that gives a hidden member its own key. Against the unpatched headers the file does not build at all, because the counts are pinned with `static_assert`, and a probe that avoids them writes `{"factor":0}` where this branch writes the four members.
- `tests/json_test/non_aggregate_reflection_test.cpp` (the C++26 suite): 19 tests / 86 asserts pass.
- `tests/reflection/reflection.cpp`: 60 tests / 282 asserts pass.
- `-fsyntax-only` over beve, msgpack, cbor, bson, toml, jsonb, yaml and json_test in the C++26 configuration: no errors.
- Compile time, four interleaved reps on GCC 16: the test file builds in 8.74 s against the base headers and 9.00 s here, while a smaller TU that builds on both sides shows no difference beyond run-to-run noise (about 7.2 s each).

**One thing I could not check on this machine:** the repo notes that P2996's `bases_of` iteration has limitations in the Bloomberg clang that `p2996-reflection.yml` runs, and this box has neither clang nor docker, only GCC 16. That job is where the clang side is settled. The helper uses no compiler-specific `std::meta` API and keeps the indexed form of the queries rather than range-for loops, because a range-for loop variable of a `std::meta` query is not usable in a constant expression, and deriving a base type from one that way yields an empty member list on GCC 16.

The fork's `msys2-mingw` job is red for a pre-existing reason unrelated to this change (MinGW's libstdc++ has no `_Float128` charconv overloads for `json_test`, and its `ws2_32` link is missing); it is red on the fork's `main` as well, and `msys2.yml` does not run on pull requests.

Fixes #2852.


wilkolbrzym-coder:
@stephenberry I sat down at the computer two hours ago and already found quite a few holes in this one. Let it stay open for now, because what is not visible today may be visible tomorrow. I think 1 to 2 days and it will be ready.

Theoretically this was supposed to be done already, but as always when I check my own code, something comes up. This PR is not ready, I will mark it explicitly when it is. I consider #2898 ready.

wilkolbrzym-coder:
@stephenberry the branch now has the full set: inherited members reflected base-first, the two shapes the reflection refuses, and the `modify` resolution the documented `glz::meta` depends on.

- `092de30b` refuses a keyed lookup whose keys cannot be told apart, as prose at the type rather than an incomplete-type error, and adds the two build-failure cases in `tests/p2996_test` that pin both refusals. Each ctest entry builds an `EXCLUDE_FROM_ALL` target and passes only when the guard's own message is in the compiler output, so a guard that stops firing fails the test.
- `bc611325` resolves a `glz::meta<T>::modify` entry by the member its pointer names, not by the first name that matches, so `glz::object("base_x", &Base::x, "x", &Derived::x)` writes `{"base_x":41,"own_x":42}` where it wrote `{"own_x":42,"x":42}` with no error reported.
- The earlier commits reflect the inherited members base-first and document the shapes that still need a `glz::meta`.

The description is updated with all eight commits and the current test totals.

PR #2898, opened 2026-09-19
Fix the BEVE nested-variant timing test under a coarse clock

### LINKED ISSUE #2896: 1 test fails
```
2/123 Testing: beve_test
2/123 Test: beve_test
Command: "/usr/ports/devel/glaze/work/.build/tests/beve_test/beve_test"
Directory: /usr/ports/devel/glaze/work/.build/tests/beve_test
"beve_test" start time: Sep 18 22:41 PDT
Output:
----------------------------------------------------------
FAILED "deeply nested variants read in linear time" ....0/tests/beve_test/beve_test.cpp:4951
 read time grew  inf x for 4x the depth
FAILED
tests: 461 (460 passed, 1 failed, 0 compile-time)
asserts: 208121 (208120 passed, 1 failed)

Performance regression test:
to_beve size: 504 bytes
to_beve: 0.131586 s, 365.275 MB/s
from_beve: 0.119572 s, 401.976 MB/s
d: ,
hello: ,
i: ,
map: /bear, /fish,
sub: /x, /y,
<end of output>
Test time =   0.54 sec
----------------------------------------------------------
Test Failed.
"beve_test" end time: Sep 18 22:41 PDT
"beve_test" time elapsed: 00:00:00
----------------------------------------------------------
```

Version: 8.4.0
FreeBSD 15.1

wilkolbrzym-coder:
This fixes the failing test reported in #2896.

`tests/beve_test/beve_test.cpp`, in "deeply nested variants read in linear time", timed a fixed 200 reads at each depth and asserted that the ratio between the two depths stayed below 8x. On FreeBSD it read `read time grew inf x for 4x the depth`, which looks like a reader regression but is not one. FreeBSD's `<time.h>` defines `CLOCKS_PER_SEC` as 128, so `std::clock()` there advances in steps of 7.8125 ms (glibc uses 1000000, MSVC 1000). Both timed loops finish inside a single tick, about 1 ms at depth 50 and 4 ms at depth 200, so the shallow measurement lands on the same tick it started on, reads exactly 0, and the ratio comes out inf, NaN or 0 depending on where the tick boundary falls. The reader itself is linear: it reads 4.1x for 4x the depth against a continuous clock, and the test's own correctness check of the deep read passes.

The measurement now counts reads per unit CPU time instead of timing a fixed number of reads: how many reads fit in a budget of `max(16 clock ticks, 20 ms of CPU time)`, with the clock sampled once per 8 reads, and the median ratio of five rounds. Counting reads cannot divide by zero, and the budget spans enough ticks that its own quantisation cannot dominate either count.

Behaviour changes, all inside this one test:

- The ratio is reads-per-budget instead of time-per-fixed-count. The bound is still 8x and the median of five rounds is unchanged.
- A clock that fails, with `clock()` returning `(clock_t)-1` as FreeBSD's getrusage-backed clock does when the syscall fails, now reports zero reads and fails an assertion. I traced this as a loop that would otherwise spin forever, and as a failure of the first sample alone stopping after one batch with equal counts for both depths, which passes the bound.
- Cost: the measurement spends about 1.5 s of CPU time on a 128-tick clock (about 30 ms before) and about 240 ms on a microsecond clock. That is what spanning 16 ticks costs. It is also the only ratio-based timing assertion in the suite; the two other timed checks assert absolute bounds on `steady_clock`, which quantisation cannot turn into garbage.

What I verified with GCC 15.2 at `-O2` (`-Wall -Wextra -pedantic`):

- `beve_test` on an unmodified tree: 461/461 tests, 10126 asserts, rc=0, 0.30 s.
- The same suite with the key-scan short-circuit reverted (`possible.popcount() == 1` to `false` in `include/glaze/beve/read.hpp`): 461 tests, 460 passed, 1 failed, only this test, `read time grew 13.8 x for 4x the depth`. The measurement still catches the regression it was written for.
- Both the old and the new measurement run against clocks quantised to 7812 us (FreeBSD) and 1000 us (MSVC): the old form reads inf on the quantised FreeBSD clock, the new one reads 3.98x to 3.99x, and the reverted reader 11.5x to 13.7x against the 8x bound.
- On the same machine the unpatched test also passes (461/461 in 0.07 s), which is why the report is FreeBSD-specific rather than a plain failure.

One note on the fork's CI: `msys2-mingw` is red on this branch, as it is on the fork's `main` at `a5fcd46f` and on all 14 of that workflow's runs on `main` there, because MinGW's libstdc++ has no `from_chars`/`to_chars` for `_Float128` while `json_test` compiles the float128 suite. That workflow has no `pull_request` trigger, so it does not run on this PR. Nothing here touches that path.

No library code is touched: this is a test-only change.

Fixes #2896.


wilkolbrzym-coder:
@stephenberry ready for review — `beve_test` is 461/461 in Debug (-O0), and with the reader's key-scan short-circuit reverted the timing check now fails at 11.1x there (13.7x at -O2).

PR #2899, opened 2026-09-19
fix int64 wrap of far-range dates in time_point readers

uwezkhan:
`chrono_detail::parse_iso8601` (JSON, YAML, CBOR tag 0) and its copy in the TOML datetime reader build the instant as `sys_days + h + min + s + nanoseconds{frac}`, so the sum is an int64 nanosecond count whatever the target is, and that count only spans 1677-09-21 to 2262-04-11. Before: `read_json` of `"9999-12-31T23:59:59Z"` into `sys_seconds` returns success holding 1816-03-30T05:56:08Z, `2300-01-01` reads as 1715, `1600-01-01` reads as 2184, and UBSan reports `253402300799 * 1000000000`. The MessagePack timestamp reader sums `seconds + nanoseconds` the same way (a timestamp 64 for year 2300 decodes as 1715), the BSON datetime reader scales a full int64 of milliseconds into the clock's finer duration (INT64_MAX decodes as 1 ms before the epoch), and `date_format` casts into a nanosecond member unchecked.

After: a shared `chrono_detail::make_sys_time` sums whole seconds and the sub-second part in the target's own precision and returns false when the target cannot hold the instant, so all five sites keep every `[0000, 9999]` date exact for a target that fits it and return `parse_error` otherwise, matching the CBOR tag 1 reader. The bound depends on the target's duration and not on the wire format, so it lives next to the cast instead of in each reader. Tradeoffs: a nanosecond target (the default `system_clock::time_point` on libstdc++) now rejects dates outside 1677-2262 that used to wrap, the MessagePack reader rejects a nanoseconds field above the spec limit of 999999999, and the numeric `epoch_time`/`epoch_count` wrappers are not touched; in-range results are bit-identical to the old `time_point_cast` expression, and the added chrono, toml, msgpack and bson tests fail without the patch.

Update after review: the bound at the boundary second now compares against the fraction `max()`/`min()` still hold in the target's precision rather than zero, so `sys_time<nanoseconds>{nanoseconds::max()}` (`2262-04-11T23:47:16.854775807Z`) round-trips as it did on main and one nanosecond past it is rejected. The MessagePack `time_point` writer also floors its second/nanosecond split instead of truncating toward zero: a pre-epoch fraction goes out as the previous second plus a positive nanoseconds field (timestamp 96) instead of a negative count wrapped into the uint32 field, so `-0.5 s` round-trips instead of reading back as `+0.574 s`, and both ends of `system_clock::time_point` survive the round trip.


wilkolbrzym-coder (commented):
Spent some time on this against `main`: the wrap is real (`9999-12-31T23:59:59Z` reads as 1816-03-30T05:56:08Z, `2300` as 1715, and UBSan flags `253402300799 * 1000000000`), in-range results are unchanged, and the new tests fail without the patch. Good catch.

One thing I'd tighten before merging: the guard compares against the truncated boundary second, so it rejects instants the target can hold. For `sys_time<nanoseconds>` that's the last 0.854775807 s of the range — `write_json(sys_time<nanoseconds>{nanoseconds::max()})` writes `"2262-04-11T23:47:16.854775807Z"` and the read now fails, where `main` round-tripped it exactly. Same in msgpack/toml/cbor/bson. Comparing against the sub-second remainder instead of `0` at the boundary keeps the genuine rejections and drops the false ones. It's the same conservatism the CBOR tag-1 path already has, so it's a consistent choice — just a lossy one, and it's the only place the round trip breaks.

Related, in the msgpack reader: the writer splits a pre-epoch epoch with `duration_cast<seconds>`, which truncates toward zero, so `-1ms` now fails to read back while `-500ms` still decodes silently wrong. A floor split there would make the two agree.

Last thing, pure curiosity: is going straight to a PR, without an issue first, just how you work? I always file the issue first, so I'd be curious to hear your reasoning.
@uwezkhan

uwezkhan:
Good catch on the boundary. The guard compared the last second against a zero fraction; it now compares against the remainder `max()`/`min()` carry in the target's precision, so `2262-04-11T23:47:16.854775807Z` reads into `sys_time<nanoseconds>` as `nanoseconds::max()` again and one nanosecond past it is still `parse_error`. The readers all share the helper, so JSON, YAML, TOML, CBOR tag 0, msgpack and BSON pick it up together. The CBOR tag 1 integer path keeps its own whole-second bound; this PR doesn't touch it, so I left it alone.

The msgpack writer now floors the split: `-1ms` goes out as `{-1, 999000000}` in a timestamp 96 and reads back exact, and `-500ms` no longer decodes as `+0.574s`. I did the floor on the split parts rather than with `floor<seconds>(epoch)`, which overflows at the clock's `min()`; with the adjusted parts both `time_point::max()` and `min()` round-trip.

Tests for both are in chrono, toml and msgpack. They fail on the previous revision and pass now, and the full suite is green under ASan/UBSan.

On the issue question: for a fix this contained, a PR with a failing test is the clearest statement of the problem. The test is the reproducer, the diff is the proposal, and the description carries what the issue would. I'd file an issue first when the right fix isn't obvious or it touches a design choice.

PR #2907, opened 2026-09-19
Fix the MinGW build: the float128 guard, Winsock for Asio, and a sanitizer probe

wilkolbrzym-coder:
The `msys2-mingw` workflow builds the test suite with MinGW GCC and runs it. It has never reached `ctest`, on any branch, `main` included. Three independent things stop it, one commit each. The branch is green end to end on a fork that runs the same job.

**The float128 suite asks the wrong question.** `__STDCPP_FLOAT128_T__` says the type exists. MinGW's libstdc++ defines it while its `to_chars` and `from_chars` have no `_Float128` overloads, and those two calls are the library's entire float128 path (`core/write_chars.hpp`, `json/read.hpp`), so `json_test` cannot compile. The suite now asks `<charconv>` itself. The type is a template parameter, which is what makes the calls dependent, so a missing overload answers `false` instead of failing the build, and the suite lambda is generic for the same reason, so the branch that cannot compile is never instantiated. Where the answer is no, the suite registers with nothing in it.

**Winsock is named for one target out of the many that need it.** `json_test` includes `glaze/file/hostname_include.hpp`, whose `#pragma comment(lib, "ws2_32.lib")` only MSVC honours, and it is the first target that needs the library. Naming it there moves the build one target further, to the next Asio target, where `WSAStartup` and `WSACleanup` are undefined. Asio names its libraries the same MSVC-only way (`asio/detail/socket_types.hpp`), and it names two. Both are needed, because `AcceptEx` and `GetAcceptExSockaddrs` live in `mswsock.dll` and not in `ws2_32.dll`. Every target in the test tree reaches Asio through `glaze::asio`, either directly or through `glz_test_common` and `glz_test_exceptions`, so the requirement is stated once on that target behind `if(MINGW)`, and it holds for all three Asio backends. Consumers of the installed package load the same module through `find_package(glaze)`.

**MinGW's GCC has no sanitizer runtime.** `fuzzing/CMakeLists.txt` applies `-fsanitize=address,undefined` to every fuzz target built with GCC or Clang. MinGW's GCC accepts the flags, instruments the objects, and then fails the link with `cannot find -lasan`, because libsanitizer is not built for MinGW targets and MSYS2 ships no `libasan` or `libubsan` in any package. The file already probes for libfuzzer a few lines above, so the sanitizer runtime is probed the same way, over the same flags the targets receive. Where a runtime exists nothing changes, and where it does not, the fuzzers build uninstrumented instead of not building at all, which is what that file is after: its first comment says the sources are compiled on every compiler to prevent code rot.

## Verification

The job on this branch, on the fork that runs the workflow on pushes: the build completes, and `ctest` reports 124 of 124 tests passing in 130 seconds. Before this branch the job never got past compiling `json_test` and never ran a single test. That run is https://github.com/wilkolbrzym-coder/glaze/actions/runs/35437549732 . The workflow only triggers on pushes, so it does not run here.

Cross compiling the branch for `x86_64-w64-mingw32` (GCC 13.2, Release, Ninja, standalone Asio 1.36.0, OpenSSL 3.6.4) checks the pieces on their own: the tree has four distinct executable link lines (plain `glaze::glaze`, that plus the Winsock pair, that plus Winsock and OpenSSL and crypt32, and `json_test`, which names `ws2_32` twice because its own link sits beside the one `glaze::asio` supplies), and a representative of each links. A native Linux configure of the same tree keeps the sanitizer flags on the fuzz targets, and the MinGW configure drops them and says so.

One test deserves a note, because it is unrelated to this change but it is what the job does next. `beve_test`'s "deeply nested variants read in linear time" failed once while I was checking this, reading `inf x for 4x the depth`, and passed on the next run. On Windows `CLOCKS_PER_SEC` is 1000, so a loop of 200 reads that finishes inside one tick measures zero and the ratio divides by it. That is #2896, with a fix open in #2898.


wilkolbrzym-coder:
Would you approve the workflow runs for this PR? As a first-time contributor they sit at `action_required`, so the matrix (gcc, clang, msvc, sanitizers) does not appear here at all. The MinGW job this PR is about only triggers on pushes, so its runs stay on the fork, where the current head is https://github.com/wilkolbrzym-coder/glaze/actions/runs/35773910015 @stephenberry

Issue #2901, opened 2026-09-19
shared_async_map element proxy is treated as a nullable, so reading it depends on the element's current value


wilkolbrzym-coder:
The element proxy of `glz::shared_async_map` exposes `operator V&()` (`include/glaze/thread/shared_async_map.hpp:305`), and `glz::nullable_t` (`include/glaze/core/common.hpp:456-463`) is:

```cpp
concept nullable_t = !meta_value_t<T> && !str_t<T> && !is_function_ptr_or_ref<T> && requires(T t) {
   bool(t);
   { *t };
};
```

For a value type that converts to `bool` (`int`, `double`, `bool`, ...) the proxy satisfies both `*proxy` and `bool(proxy)` (one user-defined conversion to `V&`, then the standard conversion to `bool`), so it is `nullable_like`.

The reader is then chosen by partial ordering of the specializations. `from<JSON, T> requires nullable_like<T>` at `include/glaze/json/read.hpp:4799` pins the format to JSON, so it is more specialized than `from<Format, is_value_proxy T>` in `include/glaze/thread/value_proxy.hpp:14`, and the nullable path wins. It does (`include/glaze/json/read.hpp:4846-4853`):

```cpp
if (!value) {
   if (!nullable_emplace<Opts>(value, ctx)) return;   // cannot emplace a proxy
}
parse<JSON>::op<Opts>(*value, ctx, it, end);
```

`nullable_emplace` has no way to set up a proxy, so it sets `invalid_nullable_read`.

**1. Deserialization of the whole map fails for scalar value types.**

```
$ g++ -std=c++23 -I include ... map_values.cpp && ./map_values
V = int                     read=1 size=1  FAILS (invalid_nullable_read)
V = double                  read=1 size=1  FAILS
V = bool                    read=1 size=1  FAILS
V = struct Point            read=0 size=1  ok
V = std::string             read=0 size=1  ok
V = std::atomic<int>        read=0 size=1  ok
```

`glz::write_json` of the same map produces `{"one":1,"two":2}`, and `std::map<std::string,int>` reads that document without any problem, so the writer and the plain map path agree.

**2. Reading a single element depends on the value it currently holds.**

```
$ ./proxy_value_dep
vector<int> element starts at 0: read_json("42") ec=1 value=0  FAIL
vector<int> element starts at 7: read_json("42") ec=0 value=42 ok
map<string,int> element starts at 0: read_json("42") ec=1 value=0  FAIL
[...]

PR #2915, opened 2026-09-19
Serialize value proxies as their element in any format

### LINKED ISSUE #2903: Reading a value_proxy always parses as JSON, so read_beve and other formats fail on their own output
`from<Format, is_value_proxy T>` always parses as JSON, ignoring the format parameter, while the matching `to` specialization respects it (`include/glaze/thread/value_proxy.hpp`):

```cpp
template <uint32_t Format, is_value_proxy T>
struct from<Format, T>
{
   static void op(auto&& value, is_context auto&& ctx, auto&& it, auto end)
   {
      parse<JSON>::op<Opts>(value.value(), ctx, it, end);      // line 19
   }
};

template <uint32_t Format, is_value_proxy T>
struct to<Format, T>
{
   static void op(auto&& value, is_context auto&& ctx, auto&&... args) noexcept
   {
      serialize<Format>::template op<Opts>(value.value(), ctx, args...);   // line 29
   }
};
```

Writing a proxy therefore produces bytes in the requested format, while reading it back expects JSON. Any format other than JSON fails on data the library itself produced.

```
$ g++ -std=c++23 -I include ... valueproxy2.cpp && ./valueproxy2
beve bytes=5 json=["new"]
read_beve(proxy): ec=1 value=[old]   FAILS
read_json(proxy): ec=0 value=[new]   ok (value changed)
```

The proxy types affected are `glz::shared_async_vector<V>::value_proxy` and `glz::shared_async_map<K, V>::value_proxy`, the only types declaring `glaze_value_proxy`.

A second visible consequence: `glz::read_beve` of a `glz::shared_async_map<std::string, std::string>` fails with `expected_quote` (code 17), because the JSON string reader is run over BEVE bytes. For other value types that read fails with a different error, from a separate cause.

Context: PR #1707, titled "Make value_proxy format agnostic", was merged on 2025-04-28 and changed this specialization's signature from `from<JSON, T>` to `from<Format, T>`. Its diff is two lines in the signature block and does not touch the call in the body, which still names JSON. The signature became format agnostic while the body did not.

Not covered by the tests because `tests/exceptions_test/exceptions_test.cpp` serializes whole containers rather than a single element proxy.

Expected: `from` uses the format it was asked for, symmetric with `to`.

stephenberry:
Fixes #2903 and #2901.

A `value_proxy` is a locked reference to one element of a `shared_async_map` / `shared_async_vector`. It now reads and writes exactly as its element does, in whatever format is asked for.

Three causes:

- `from<Format, is_value_proxy T>` called `parse<JSON>` in its body, so reading a proxy always parsed JSON while writing honored `Format`. Reading back the library's own non-JSON output failed. The reader also gains the tag-ahead-of-context overload that MSGPACK (and BEVE under `no_header`) dispatch with, which it never matched before.
- A proxy converts implicitly to its element, so for a bool convertible element it satisfied `nullable_t`. Format-specific nullable specializations out-rank the format-generic proxy specialization, so the proxy was read and written as a nullable: `read_json` of a `shared_async_map<std::string, int>` failed with `invalid_nullable_read`, and a proxy holding `0` or `false` wrote as `null`. `nullable_t` now excludes value proxies, and `const_value_proxy` carries the `glaze_value_proxy` marker it was missing.
- The proxies' `operator=` was an unconstrained template, so a proxy answered yes to every `requires { proxy = x; }` probe and then hard errored in the body. That is why the YAML reader, which probes for null assignment, could not compile against a proxy of a non-nullable element. It is now constrained to what the element accepts.

Behavior change worth noting: `read_json` of a `shared_async_map<std::string, int>` given `{"x":null}` used to return success and insert a default-constructed element. It now reports `parse_number_failure`, matching what `std::map<std::string, int>` does with the same input.

Tests: new `value_proxy_serialization_tests` suite in `threading_test`, covering proxy and whole-container round trips in JSON, BEVE, MSGPACK and YAML, falsy elements writing as values rather than `null`, and null parity with `std::map`. The existing suites missed this because they use `std::atomic<int>` elements, which need two user-defined conversions to reach `bool` and so never satisfied `nullable_t`.


wilkolbrzym-coder (commented):
Verified on my side with g++ 15.2, including ASan and UBSan: threading_test 88/88, exceptions_test 50/50, no sanitizer reports. The existing shared_async tests on std::atomic still pass, so the constrained operator= and the nullable_t change did not disturb them.

The comment in the code about dispatching to from rather than to parse matches what the dispatchers actually do, MSGPACK hands the tag straight to from, while parse's tag overload requires no_header. Good that it is written down, it is not obvious.

The one thing I would add is coverage. The new tag overload is only exercised through MSGPACK, while BSON always dispatches with a tag and has no check_no_header anywhere in bson/read.hpp, so it goes exactly this way. A proxy round trip in BSON would close that gap.

PR #2916, opened 2026-09-19
Remove shared_async_map and shared_async_vector

stephenberry:
Stacked on #2915; retarget to `main` once that merges.

Removes `glz::shared_async_map`, `glz::shared_async_vector` and the `value_proxy` machinery they alone used, ~2000 lines.

Why:

- No callers. Nothing in `include/` used them, and a GitHub code search finds no instantiation of either type in any public source file. Every hit is a vendored copy of the header, an OS package file manifest, a vcpkg install listing, or a snapshot of this repo.
- Undiscoverable. `docs/threading/shared-async-map.md` was never added to the `mkdocs.yml` nav, so the documentation was unreachable from the site.
- They cost the core. Their per-element proxy converted implicitly to the element and dereferenced to it, so for a bool convertible element it satisfied `nullable_t`. Working around that put an `is_value_proxy` clause into `glz::nullable_t` in `core/common.hpp`. That clause and the concept go away here; `core/common.hpp` and `threading_test.cpp` return byte for byte to their pre-#2915 state.
- The design is served better elsewhere. `glz::async<T>` wraps any container behind `read`/`write`, and `async_vector` / `async_string` cover the common cases by holding one lock for the whole document rather than handing out a proxy that impersonates an element.

Also repoints `async_vector.hpp` at `glaze/core/common.hpp`, which is what it was actually getting through `value_proxy.hpp`.

This is a breaking change for anyone using these types; pinning the previous release is the migration path.


wilkolbrzym-coder:
Removal verified: a grep across the whole repo, code, CI, CMake, docs and examples, leaves no reference to shared_async_map, shared_async_vector or value_proxy. threading_test 78/78, exceptions_test 32/32 on my side with g++ 15.2.

core/common.hpp and threading_test.cpp are identical to current main, so the return to the pre-2915 state holds against today's main as well, not just against the branch point. async_vector.hpp now includes common.hpp directly, which is correct, it was only getting it transitively before.

Two formal things. These types have shipped from v4.0.3 through v8.4.0, so the removal deserves a line in the release notes. And the base needs retargeting from fix-value-proxy-format to main, though there will be no conflict, since the only newer commit on main, 85ce28f6, touches json_test.cpp alone.

Stephen, honestly I did not expect you to look in here, I assumed you take weekends off, but apparently not. If anything, I will slow the pace down a bit, because I can see I have been pushing it too hard.
@stephenberry

PR #2917, opened 2026-09-20
send no body for HEAD, 204 and 304 replies in send_response_with_conn

uwezkhan:
`send_response_with_conn` writes `Content-Length` and the body for every response, including the ones RFC 9112 6.3 ends at the empty line after the headers whatever their fields say: a reply to HEAD and any 1xx, 204 or 304. A recipient reads no body from those, so the bytes stay on the keep-alive connection and are read as the start of the next response (response-side desync, CWE-444 family). Two triggers need no handler mistake: `HEAD` to a path with no HEAD route gets the server's own 404 text, so the wire shows `Not FoundHTTP/1.1 200 OK`, and `create_cors_middleware` writes its 403 text before checking the origin and then only replaces the status, so every allowed preflight, the default `enable_cors()` included, answers `204 No Content` with `Content-Length: 24` and `CORS: Origin not allowed`. curl frames both replies as bodyless, logs `Excess found writing body: excess = 24` and drops the connection.

Before: content follows the headers of a HEAD, 204 or 304 reply, and a `Content-Length` the handler set reaches the wire on a 1xx or 204. After: the writer sends the header section alone for those, sends no `Content-Length` on a 1xx or 204 at all, a handler-set one included (RFC 9110 8.6, the same treatment a handler-set `Transfer-Encoding` already gets), generates none for a 304 while keeping one the handler set, and keeps it for HEAD, where it is the length GET would return; a 304 gets its `Not Modified` reason phrase, which `get_status_message` lacked; the CORS middleware writes the 403 only when the origin is not allowed. The rule sits in the wire writer because it is the one layer that sees both the request method and the final status, and `http_client`'s streaming reader already frames these replies the same way (`carries_no_body`), so both ends now agree. Tradeoffs: a body a handler set on a 1xx, 204 or 304 is no longer sent, a 204 loses its generated `Content-Length: 0` and any handler-set one, and the connection's request method now resets on keep-alive so a malformed request line after a HEAD is not answered as one; responses that may carry a body are byte-identical. The added `http_smuggling_test` and `cors_test` cases fail without the patch, except "A malformed request after HEAD is not answered as a HEAD", which passes on the parent commit: the method reset it guards is introduced here, and the case pins it against dropping the reset.


wilkolbrzym-coder (commented):
Tested against 85ce28f6 (built both, pipelined raw requests, plus glz::http_client end to end). The rule is in the right layer and the two ends agree now: a 204 no longer poisons the keep-alive socket, HEAD keeps its Content-Length and drops the body, and 1xx/204/304 lose the generated length. The new tests hold up: on the parent commit 5 of 6 smuggling cases and both new cors cases fail, which I checked by dropping the new test files onto 85ce28f6 and building them.

One blocker. A Content-Length set by the handler still reaches the wire on 1xx/204/304, because only the generated one is gated (http_server.hpp:3140). Measured:

    204 + res.header("Content-Length", "99")  ->  HTTP/1.1 204 No Content ... Content-Length: 99 ... no body

RFC 9110 8.6 is an unconditional MUST NOT for 1xx and 204, this function already drops a handler-set Transfer-Encoding twelve lines above for the same reason, and a recipient that frames by Content-Length reads the next reply as this body. So the comment and the docs bullet are true only for the generated case. Fix: skip a user-supplied content-length when the status is 1xx or 204. Keep it for HEAD, it is legitimate there, and 304 can keep it too since the handler may know the entity length.

Then three smaller ones:

1. The 1xx clause has no coverage. Delete it and the suite stays green (11/11, 8/8), and nothing in the server emits 1xx on its own. Pin it with res.status(100) plus a body.
2. A 304 goes out as "HTTP/1.1 304 Unknown". get_status_message has no case 304 in either copy (:1243, :3305), while glz::http_status_reason_phrase in http.hpp already returns "Not Modified". One line, and this PR is what documents 304.
3. "A malformed request after HEAD is not answered as a HEAD" passes on the parent, since the method reset it guards is new here. Say that in the description instead of listing it under the cases that fail without the patch.

Numbers: 7 networking targets green on the PR (191 tests, 670 asserts), 6 of 7 mutations of the new rules caught by the suite, the uncaught one being the 1xx clause.

uwezkhan:
Good catch on the handler-set length, pushed in 38b63da.

Blocker: a handler-set Content-Length is now dropped on 1xx and 204 in the same header loop that drops Transfer-Encoding, twelve lines up from where the generated one was gated. HEAD and 304 keep theirs. Before: `204 + res.header("Content-Length", "99")` went out as `Content-Length: 99` with no body. After: no Content-Length at all. The docs bullet now splits 1xx/204 (none sent) from 304 (none generated, handler-set kept).

1. 1xx is pinned: `/continue` sets status 100 plus a body. Removing the 1xx clause fails it with `staleHTTP/1.1 200 OK` in the remainder. A second case pins the handler-set drop on a 204; removing the drop fails it with `Content-Length: 99` in the head.
2. 304 returns `Not Modified` from both `get_status_message` copies, and the 304 case checks the full status line.
3. Description now says the malformed-after-HEAD case passes on the parent and is there to pin the method reset.

Tradeoff: a handler that put a Content-Length on a 204 loses it silently, same as its Transfer-Encoding. 1xx reason phrases are left alone, so a 100 still reads `100 Unknown`; nothing in the server emits one on its own.

14 networking targets green under ASan+UBSan, clang-format clean.

